PR #15, opened 2015-11-18
Travis integration

cfieber:


cfieber:
noticed we try to artifactoryUpload from a PR within the project, trying again as #16 from a project fork to test difference in behaviour

PR #22, opened 2015-11-23
Upgrade to Swagger2 (as part of springfox)

ajordens:
- Supports generation of a static API site via an integration with swagger2markup (http://springfox.github.io/springfox/docs/snapshot/#configuring-springfox-staticdocs)


ajordens:
- No need to bundle the swagger-ui assets
- Swagger2 > Swagger
- io.springfox > com.mangofactory:swaggermvc

duftler:
Looks like this breaks front50/rosco/rush as they are using swagger annotations that no longer resolve.

ajordens:
Yes, I'll need to push through updates for anything that was using the old
annotations.

I guess I'm the new world any updates to dependencies take effect
immediately?

I'll get those other packages updated right away.

duftler:
Thanks!

PR #26, opened 2015-12-01
Updates kork-core with more complete configuration for Eureka

cfieber:
Adds archaius binding on to spring Environment
removes the mocked out Eureka LookupService and replaces it with full DiscoveryClient configuration (Conditional on eureka.enabled property).

Adds additional spectator configuration

Removes kork-elasticsearch since it is no longer used or necessary

@spinnaker/netflix-reviewers PTAL


cfieber:
Just a bit of commentary here since it's a lot of files changed:

Initialization is no longer conditional on the absence of some internal Netflix classes - this change set is a move to make it so we aren't bringing in our internal platform library in an external build.

Initializes Archaius (always) with a configuration source that reads from Spring's Environment. This allows us to configure libraries such as Eureka or Hystrix in our service yml config files.

Initializes Spectator (always) with JVM metrics, GC logging. Conditionally exports a DefaultRegistry (pretty much useless) if no other Registry is found. Internally we provide a different binding (ServoRegistry) so that we can publish to our internal metrics system. Someone else interested in monitoring Spinnaker could, for example, export a Dropwizard-metrics Registry and use that to export metrics to graphite.

Initializes Eureka (conditionally on property eureka.enabled). Due to the Archaius binding, Eureka configuration can be read from the service yml file.

ajordens:
LGTM

PR #29, opened 2016-01-12
prevent OkHttp from retrying HTTPS POSTs when they time out

robfletcher:
@cfieber PTAL


ajordens:
We weren't expecting the auto-retry behavior previously so I'm not sure what the harm would be ... that said if it were happening a lot it may point out areas where we'd need to handle our own retries (possibly like we already do on `NETWORK` errors in `orca`).

PR #33, opened 2016-03-10
Support explicitly disabling Cassandra

ajordens:


ajordens:
Any reason not to do this @cfieber?

Motivation is to support having a dependency on kork-cassandra (in Front50) but not actually initializing anything unless explicitly enabled.

cfieber:
yeah makes sense to me

PR #37, opened 2016-05-03
Makes User implement UserDetails to work with Spring Security

ttomsu:
@cfieber or @ajordens PTAL

@jtk54 FYI


ajordens:
LGTM

ttomsu:
@ajordens FYI I added a Jackson dependency to suppress printing the 'password' field - even though it would never be filled in anyway.

PR #38, opened 2016-05-20
Add some defensive checks around spectator tag generation for AWS backoff/retry instrumentation

cfieber:
@anotherchrisberry PTAL addresses exceptions we were seeing this morning


anotherchrisberry:
👍

PR #41, opened 2016-06-25
Adds support for configuring CRL or custom blacklisting on tomcat SSL

cfieber:
This is a bit convoluted due to what it takes to wire in the TrustManager to tomcat

This exposes a `server.ssl.crlFile` configuration property which is configured onto the tomcat connector.

That can either be the path to a legitimate CRL file, in which case the standard Tomcat CRL handling will apply.

Alternatively this can be the path to a custom blacklist file. To configure this way, prefix the path with `blacklist:`

E.g. in the relevant yaml file:

````yaml
server:
  ssl:
    crlFile: blacklist:/path/to/my/blacklist
````

The custom blacklist takes the form of one entry per line following
````
certificate issuer x500 name:::certificate serial number
````

e.g.
````
OU=Some OU, O=My Org, CN=My CA, L=Where Waldo is, ST=CA, C=US:::12345
````

A client presented certificate will be compared against the blacklist and rejected if it is found, otherwise SSL negotiation will fall back to Tomcat's standard behavior.

The two primary advantages of this custom implementation are:
1) Managing a CA and generating CRLs is hard
2) The file will refresh every 5 seconds, so it can be updated in a running system to revoke certificates without restarting the process

I'd like to say there are lots of unit tests around this but standing up the PKIX bits for a test seemed a bit painful.  I've manually tested this pretty extensively.

If https://github.com/spring-projects/spring-boot/issues/6171 gets implemented then the `@Value` annotation in here probably becomes redundant.

@spinnaker/netflix-reviewers PTAL

### DIFF
--- a/kork-web/kork-web.gradle
+++ b/kork-web/kork-web.gradle
@@ -7,6 +7,7 @@ dependencies {
   compile spinnaker.dependency('groovy')
   compile spinnaker.dependency('okHttp')
   compile spinnaker.dependency('retrofit')
+  compile spinnaker.dependency('guava')
 
 
   spinnaker.group('spockBase')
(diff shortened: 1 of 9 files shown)


cfieber:
added a couple unit tests

PR #45, opened 2016-09-27
Add kork-stackdriver module providing optional stackdriver publisher.

ewiseblatt:
This also pulls in spectator-web-spring which is in review but expect it
to be available soon.

@cfieber 
This is really two modules in one:
(1) A "spectator-ext-stackdriver", which conceptually belongs in spectator but logistically wont work out because I dont have access to that repo to maintain.
(2) A kork-spectator module that just adds a StackdriverConfig class to wire up an optional stackdriver writer and a metrics web endpoint provided by spectator-web-spring. I dont actually need the web endpoint from a production standpoint, but it is convenient for debugging, offers another option for grabbing metrics for all spinnaker deployments, and I do reference a couple classes I added into that package.

I'm only asking you to review (2) keeping in mind that I'm going to add this dependency to all the spinnaker microservices. You are welcome to review (1) as well, but I can get other people to review that to distribute the load.

The local reference to :spectator-web-spring is temporary while I am waiting for that module to get approved and published. I'll consume the published version before ultimately merging the PR.

@duftler 
If you are curious to know more about Stackdriver, you can review (1).
Otherwise I can find someone else.

### DIFF
--- a/kork-stackdriver/kork-stackdriver.gradle
+++ b/kork-stackdriver/kork-stackdriver.gradle
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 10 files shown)


ewiseblatt:
Breaking this up into individual PRs

PR #46, opened 2016-09-30
Added stackdriver writer.

ewiseblatt:
@duftler 
This PR is for a module intended for spectator but being placed here for the initial release to simplify initial integration with spinnaker as well as ongoing maintenance until the integration has a chance to stabilize and spinnaker catches up to spectator.

I am going to layer the KorkConfig on top of this in a followup PR. In the long term, this module would only be the KorkConfig and this PR would be refactored out into a spectator/spectator-ext-stackdriver. I think it will be easiest and simplest between now and then to just keep everything here. The java namespace I use for this PR pretends this is in spectator, not kork.

This StackdriverWriter dumps a SpectatorRegistry into Stackdriver Time Series data.
It is based on a runtime conversion, making up Stackdriver schemas on demand.
I think this is simpler from an experimental and maintainence perspective, however doing so requires the ability to read from stackdriver at runtime in order to see what schemas have already been created or not. That means that you need a read-write scope whereas the GCE default scope is just write. If this is a problem in practice, I can change it later so that someone else creates and maintains the schemas (which seems like a complicated process) or I could add some error handling for error cases and respond by creating. The approach here seems the simplest in addition to minimizing the HTTP errors that would be reported back.

The implementation here allows two different strategies for storing data because I havent committed to one yet. One uses reuses metric types across the different services, and adds an additional label to identify which service the metric is for. The other uses a different metric type for each service. There are different tradeoffs here. I'm likely to go with the second approach (which is the default) but left the other implementation here until we get more experience with it.

There are some other stackdriver bug/limitation workarounds here too. I support EC2 but havent tested it there yet. I've run this on GCE as well as on Prem. There's a hook for GKE but I'll probably populate it in a future PR.

### DIFF
--- a/kork-stackdriver/kork-stackdriver.gradle
+++ b/kork-stackdriver/kork-stackdriver.gradle
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 Netflix, Inc.
+ * Copyright 2016 Google, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


ewiseblatt:
I used 80 columns. I know Java is typically wider, but I live in 80 column windows (and use 2 space indents to get some width back).

I think my formatting was biased from all the python I've been writing, using 2 space indent with 4 spaces for statement continuations.

I could go wider if that's expected. I'm not sure of a style guide. I originally wrote this in the spectator project which has a "checkstyleMain" target (and findbugsMain and pmdMain) which passed all 3 of those rule bases. Kork doesnt have anything to my knowledge.

PR #48, opened 2016-10-05
Pull User.username rather than email when dealing with the X-SPINNAKE…

ttomsu:
…R-USER header. 

Fixes https://github.com/spinnaker/fiat/issues/93
Fixes https://github.com/spinnaker/spinnaker/issues/1143

@duftler @jtk54 @ajordens @cfieber PTAL
cc @gbougeard

### DIFF
--- a/kork-security/src/main/groovy/com/netflix/spinnaker/security/AuthenticatedRequest.groovy
+++ b/kork-security/src/main/groovy/com/netflix/spinnaker/security/AuthenticatedRequest.groovy
@@ -29,9 +29,10 @@ class AuthenticatedRequest {
   /**
    * Ensure an appropriate MDC context is available when {@code closure} is executed.
    */
-  public static final Closure propagate(Closure closure,
-                                        boolean restoreOriginalContext = true,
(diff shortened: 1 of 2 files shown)


kork-security/src/main/groovy/com/netflix/spinnaker/security/AuthenticatedRequest.groovy
```diff
@@ -54,17 +55,15 @@ class AuthenticatedRequest {
           MDC.put(SPINNAKER_ACCOUNTS, spinnakerAccounts)
         }
         closure()
-      } catch (Exception e) {
-        log.error("Error occurred propagating authentication context", e)
```
ajordens: Did you happen to verify that we get logged output still? 

Seems to me that a lot of times this happens over thread boundaries and the exception might not propagate out.

ttomsu: `AbstractHystrixCommand` in Gate still catches and prints an error. 

More to the point, printing this exception at this point (and with this error message) causes confusion.

PR #50, opened 2016-10-06
Adds Stackdriver config

ewiseblatt:
@cfieber 
This is the PR from earlier this week. I went to update it against head and did something that caused the PR to get dropped for some reason.

This adds a bean that can get picked up by all the microservices to add stackdriver support in (false by default). I'd like it to also add in the web metrics API I added into spectator 0.41 but commented it out because spinnaker is only on 0.36. I originally wrote it against the source so it had been working until I stopped building against source and started consuming spectator as spinnaker dependencies. I'm leaving it in.

There are a couple comments in the config with questions to you basically asking whether I can delete the commented out block or if in fact there is an issue I need to consider. Should be straightforward.

Thanks.

### DIFF
--- a/kork-stackdriver/src/main/java/com/netflix/spinnaker/config/StackdriverConfig.java
+++ b/kork-stackdriver/src/main/java/com/netflix/spinnaker/config/StackdriverConfig.java
@@ -1,0 +1,173 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.config;
+
+import com.netflix.spectator.controllers.filter.PrototypeMeasurementFilter;
+import com.netflix.spectator.controllers.MetricsController;
+import com.netflix.spectator.stackdriver.ConfigParams;
+import com.netflix.spectator.stackdriver.MetricDescriptorCache;
+import com.netflix.spectator.stackdriver.StackdriverWriter;
+
+import com.netflix.spectator.api.DefaultRegistry;
+import com.netflix.spectator.api.Measurement;
+import com.netflix.spectator.api.Registry;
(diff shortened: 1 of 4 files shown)


cfieber:
I think you should update to latest spectator, but just be aware that means that there are a couple of minor fixes required to the rest of the services to update them to that version. They are not significant, I think it is just changing spots that were previously using Spectator's `ValueFunction` to use java's `ToDoubleFunction`

kork-stackdriver/src/main/java/com/netflix/spinnaker/config/StackdriverConfig.java
```diff
@@ -0,0 +1,206 @@
...
+  @Bean
+  @ConditionalOnExpression("${spectator.stackdriver.defaultRegistry:false}")
+  Registry defaultMetricRegistry() {
+    Logger log = LoggerFactory.getLogger("StackdriverConfig");
+    log.info("Creating default registry.");
+    return new DefaultRegistry(com.netflix.spectator.api.Clock.SYSTEM);
+  }
+   * This causes a circular dependency for some reason, even with disabled.
+   *
+   * Reviewer, please confirm that I dont need this.
```
cfieber: I think because you have an `@Autowired` Registry, and also an `@Bean` Registry (it can't create this class because the bean it needs to autowire in is provided by the class)

kork-stackdriver/src/main/java/com/netflix/spinnaker/config/StackdriverConfig.java
```diff
@@ -0,0 +1,206 @@
...
+
+    return stackdriver;
+  }
+
+  /**
+   * We can optionally add a registry if one doesnt exist.
+   *
+   * While we need a registry, it could be provided elsewhere.
+  @Bean
+  @ConditionalOnExpression("${spectator.stackdriver.defaultRegistry:false}")
```
cfieber: there is already a Registry getting exported in kork-core (essentially this same export).

if you want to leave this in here, I'd suggest using `@ConditionalOnMissingBean(Registry.class)` on this (if no one else has supplied a registry, then create one)

You will also need to take off the class level `@Autowired` Registry, and make it a parameter of all the `@Bean` methods that are using it

kork-stackdriver/src/main/java/com/netflix/spinnaker/config/StackdriverConfig.java
ewiseblatt: Is it safe for me to assume the autowire exists and hook into it?

cfieber:
LGTM - my preference is update the spectator version in spinnaker-dependencies to whatever you need instead of working around the older version

ewiseblatt:
I took out the stuff that was commented out and assume the registry is elsewhere (autowired).
To fix the problems I had to update jacoco in kork itself.
Please confirm this version change is ok before I submit.

cfieber:
LGTM

PR #57, opened 2016-10-19
Stackdriver fixes.

ewiseblatt:
@lwander 

If running on gce, you have to store metrics in the project of the managed
resources so force the project. But still need to configure the project if
you are not on gce (and thus dont have a gce managed resource).

When transforming [timer] metrics, lookup the hints for the original metric
name, not the transformed one.

### DIFF
--- a/kork-stackdriver/src/main/java/com/netflix/spectator/stackdriver/MetricDescriptorCache.java
+++ b/kork-stackdriver/src/main/java/com/netflix/spectator/stackdriver/MetricDescriptorCache.java
@@ -623,13 +623,18 @@ public class MetricDescriptorCache {
     descriptor.setDisplayName(id.name());
     descriptor.setType(descriptorType);
     descriptor.setValueType("DOUBLE");
+
+    String untransformedDescriptorType;
     if (meterIsTimer(registry, meter)) {
-      if (id.name().endsWith("_totalTime")) {
+      if (id.name().endsWith("__totalTime")) {
         descriptor.setUnit("ns");
       }
       descriptor.setMetricKind("CUMULATIVE");
+      int suffixOffset = descriptorType.lastIndexOf("__");
+      untransformedDescriptorType = descriptorType.substring(0, suffixOffset);
     } else {
       descriptor.setMetricKind(meterToKind(registry, meter));
+      untransformedDescriptorType = descriptorType;
     }
 
     List<LabelDescriptor> labels = new ArrayList<LabelDescriptor>();
@@ -647,7 +652,7 @@ public class MetricDescriptorCache {
        labels.add(labelDescriptor);
     }
 
-    maybeAddLabelHints(descriptorType, labels);
+    maybeAddLabelHints(untransformedDescriptorType, labels);
     if (labels.size() > 10) {
         log.error("{} has too many labels for stackdriver to handle: {}",
                   id.name(), labels);
--- a/kork-stackdriver/src/test/java/com/netflix/spectator/stackdriver/MetricDescriptorCacheTest.java
+++ b/kork-stackdriver/src/test/java/com/netflix/spectator/stackdriver/MetricDescriptorCacheTest.java
@@ -111,7 +111,8 @@ public class MetricDescriptorCacheTest {
   DefaultRegistry registry = new DefaultRegistry(clock);
 
   TestableMetricDescriptorCache cache;
-  String projectName = "test-project";
+  String projectName = new MonitoredResourceBuilder().determineProjectName("test-project");
+  String projectResourceName = "projects/" + projectName;
   String applicationName = "test-application";
 
   Id idA = registry.createId("idA");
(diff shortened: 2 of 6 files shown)


ewiseblatt:
@ttomsu 
Could you review this, I need to get into the release queue here.
I keep spinning under version numbers changes and every time I go back, I need to update again.

kork-stackdriver/src/test/java/com/netflix/spectator/stackdriver/StackdriverWriterTest.java
```diff
@@ -135,20 +138,24 @@ public boolean test(Measurement measurement) {
...
       descriptor.setMetricKind("GAUGE");
 
       List<LabelDescriptor> labels = new ArrayList<LabelDescriptor>();
-      LabelDescriptor labelDescriptor = new LabelDescriptor();
-      labelDescriptor.setKey(MetricDescriptorCache.APPLICATION_LABEL);
-      labelDescriptor.setValueType("STRING");
-      labelDescriptor.setKey(MetricDescriptorCache.INSTANCE_LABEL);
-      labelDescriptor.setValueType("STRING");
-      labels.add(labelDescriptor);
+      for (String extra
```
lwander: The defacto line-length in Spinnaker seems to be 100 characters rather than 80

kork-stackdriver/src/test/java/com/netflix/spectator/stackdriver/StackdriverWriterTest.java
```diff
@@ -490,4 +498,76 @@ public void writeRegistryWithLargeRegistry() throws IOException {
...
+    when(descriptorsApi.create(eq(projectResourceName),
+                               eq(descriptorTime)))
+        .thenReturn(timeMockCreateMethod);
+    doAnswer(new Answer<MetricDescriptor>() {
+            public MetricDescriptor answer(InvocationOnMock invocation) {
+                descriptorRegistrySpy
+                    .injectDescriptor(descriptorTime.getType(), descriptorTime);
+                return descriptorTime;
+            }}).when(timeMockCreateMethod).execute();
+
```
lwander: Extra newline

lwander:
LGTM after comments

kork-stackdriver/src/main/java/com/netflix/spectator/stackdriver/MetricDescriptorCache.java
```diff
@@ -623,13 +623,18 @@ void createDescriptorInServer(Id id, Registry registry, Meter meter)
     descriptor.setDisplayName(id.name());
     descriptor.setType(descriptorType);
     descriptor.setValueType("DOUBLE");
+
+    String untransformedDescriptorType;
```
ttomsu: `String untransformedDescriptorType = descriptorType;`

Then drop L637

ttomsu: Also, if you're modfiying the original `descriptorType`, wouldn't it be better called `transformedDescriptorType` ? Or `modifiedDescriptorType` ?

kork-stackdriver/src/test/java/com/netflix/spectator/stackdriver/MetricDescriptorCacheTest.java
```diff
@@ -203,7 +204,7 @@ public void descriptorTypeAreCompliant() {
   public void descriptorNamesAreCompliant() {
     // https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors
     Assert.assertEquals(
-        "projects/test-project/metricDescriptors/" + cache.idToDescriptorType(idA),
+        projectResourceName + "/metricDescriptors/" + cache.idToDescriptorType(idA),
```
ttomsu: I find it better to keep explicit string values in tests. There was an internal post I'll point you to that states the case nicely

kork-stackdriver/src/test/java/com/netflix/spectator/stackdriver/StackdriverWriterTest.java
ttomsu: +1

kork-stackdriver/src/main/java/com/netflix/spectator/stackdriver/MetricDescriptorCache.java
ewiseblatt: Changed this (in another PR) and added a comment. The transformation already happened before this was called. So the intent here is to perform the inverse transformation back to the original name, which is the name that is the key.

kork-stackdriver/src/test/java/com/netflix/spectator/stackdriver/MetricDescriptorCacheTest.java
ewiseblatt: I agree -- normally.

The problem is that the name is no longer constant. It has to be a specific name determined at runtime. Or at least I'm making it this way for now in order to simplify the implementation. Making a way to mock that particular name is complicated right now.

The issue is that it turns out that Stackdriver requires this to be your project name if you are deployed on GCE. Without an ability to mock the name (which would add back more complexity, though is something I'd like to be able to test at some point), there is no constant string here anymore. Prior to this CL I was telling it which project to use, so it was constant. Now the name comes from a different place that enforces it and isnt mockable. When you are running the test on GCE, the implementation will detect that and use the GCE name as part of its enforcement. I use that runtime computed name to initialize projectResourceName. It's opaque for purposes of testing and a function of the runtime so is not statically known.

ewiseblatt: I think I have an idea on how I can inject a name for testing afterall.

PR #62, opened 2016-10-21
Remove Autowiring and Value in Configurations.

cfieber:
Step towards working with newer Spring versions that have problems working with `@Value` annotated properties, particularly at the `@Configuration` class level.

There are some changes in the lifecycle in Spring 4.3 that manifest in `@Value` annotations not reading from configuration if they have defaults set. This moves all the configuration into `@ConfigurationProperties` beans instead. Additionally it is recommended to keep `@Configuration` classes simple - to that end I've removed `@Autowired` fields in those classes and replaced with adding the appropriate parameter to whatever `@Bean` method was using the autowired value.

Other noise in this PR: removed a `Precondition` class in favor of the same functionality that is native in java 8 (`Objects.requireNonNull`)

### DIFF
--- a/kork-stackdriver/src/main/java/com/netflix/spinnaker/config/StackdriverConfig.java
+++ b/kork-stackdriver/src/main/java/com/netflix/spinnaker/config/StackdriverConfig.java
@@ -22,72 +22,153 @@ import com.netflix.spectator.stackdriver.ConfigParams;
 import com.netflix.spectator.stackdriver.MetricDescriptorCache;
 import com.netflix.spectator.stackdriver.StackdriverWriter;
 
-import com.netflix.spectator.api.DefaultRegistry;
 import com.netflix.spectator.api.Measurement;
 import com.netflix.spectator.api.Registry;
-import com.netflix.spectator.api.Spectator;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Value;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.boot.context.properties.NestedConfigurationProperty;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.context.annotation.Import;
 
 
 import java.io.IOException;
-import java.util.Arrays;
-import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
-import java.util.Map;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Predicate;
 
+import org.springframework.core.env.Environment;
 import rx.Observable;
 import rx.Scheduler;
 import rx.schedulers.Schedulers;
 
 
 @Configuration
-@EnableConfigurationProperties
-@Import({MetricsController.class})
-@ConditionalOnExpression("${spectator.stackdriver.enabled:false}")
-class StackdriverConfig {
+@EnableConfigurationProperties({StackdriverConfig.SpectatorStackdriverConfigurationProperties.class, StackdriverConfig.StackdriverConfigurationHints.class})
+@ConditionalOnProperty("spectator.stackdriver.enabled")
+public class StackdriverConfig {
 
-  @Value("${spectator.applicationName:${spring.application.name}}")
-  private String applicationName;
+  @Bean
+  @ConditionalOnProperty("spectator.webEndpoint.enabled")
+  MetricsController metricsController(Registry registry) {
+    return new MetricsController();
+  }
 
-  @Value("${spectator.stackdriver.credentialsPath:}")
-  private String credentialsPath;
+  @ConfigurationProperties("spectator")
+  public static class SpectatorStackdriverConfigurationProperties {
+    public static class StackdriverProperties {
+      private String credentialsPath = "";
+      private String projectName = "";
+      private boolean uniqueMetricsPerApplication = true;
+      private int period = 60;
 
-  @Value("${spectator.stackdriver.projectName:}")
-  private String projectName;
+      public String getCredentialsPath() {
+        return credentialsPath;
+      }
 
-  @Value("${spectator.stackdriver.uniqueMetricsPerApplication:true}")
-  private boolean uniqueMetricsPerApplication;
+      public void setCredentialsPath(String credentialsPath) {
+        this.credentialsPath = credentialsPath;
+      }
 
-  // If provided, takes precedence over the *Regex filters above.
-  @Value("${spectator.webEndpoint.prototypeFilter.path:}")
-  private String prototypeFilterPath;
+      public String getProjectName() {
+        return projectName;
+      }
 
-  @Value("${spectator.stackdriver.period:60}")
-  private int pushPeriodSecs;
+      public void setProjectName(String projectName) {
+        this.projectName = projectName;
+      }
+
+      public boolean isUniqueMetricsPerApplication() {
+        return uniqueMetricsPerApplication;
+      }
+
+      public void setUniqueMetricsPerApplication(boolean uniqueMetricsPerApplication) {
+        this.uniqueMetricsPerApplication = uniqueMetricsPerApplication;
+      }
+
+      public int getPeriod() {
+        return period;
+      }
+
+      public void setPeriod(int period) {
+        this.period = period;
+      }
+    }
+
+    public static class WebEndpointProperties {
+      public static class PrototypeFilterPath {
+        private String path = "";
+
+        public String getPath() {
+          return path;
+        }
+
+        public void setPath(String path) {
+          this.path = path;
+        }
+      }
+
+      @NestedConfigurationProperty
+      private PrototypeFilterPath prototypeFilter = new PrototypeFilterPath();
+
+      public PrototypeFilterPath getPrototypeFilter() {
+        return prototypeFilter;
+      }
+
+      public void setPrototypeFilter(PrototypeFilterPath prototypeFilter) {
+        this.prototypeFilter = prototypeFilter;
+      }
+    }
+
+    private String applicationName;
+
+    @NestedConfigurationProperty
+    private StackdriverProperties stackdriver = new StackdriverProperties();
+
+    @NestedConfigurationProperty
+    private WebEndpointProperties webEndpoint = new WebEndpointProperties();
+
+    public String getApplicationName(String defaultValue) {
+      if (applicationName == null || applicationName.isEmpty()) {
+        return defaultValue;
+      }
+      return applicationName;
+    }
+
+    public void setApplicationName(String applicationName) {
+      this.applicationName = applicationName;
+    }
+
+    public StackdriverProperties getStackdriver() {
+      return stackdriver;
+    }
+
+    public void setStackdriver(StackdriverProperties stackdriver) {
+      this.stackdriver = stackdriver;
+    }
+
+    public WebEndpointProperties getWebEndpoint() {
+      return webEndpoint;
+    }
+
+    public void setWebEndpoint(WebEndpointProperties webEndpoint) {
+      this.webEndpoint = webEndpoint;
+    }
+  }
 
   private StackdriverWriter stackdriver;
 
-  @Configuration
-  @ConfigurationProperties(prefix="stackdriver")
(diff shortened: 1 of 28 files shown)


cfieber:
Probably my least favorite thing is deeply nested `@ConfigurationProperties` but I followed the existing property conventions so that existing configs don't have to change - the stackdriver bit was maybe the worst case of that, and there might be some additional magic possible in how `@ConfigurationProperties` get initialized that would allow flattening out some of those cases, but for now I'd prefer to just lean towards explicitly mapping out the nested object graph as simple pojos.

cfieber:
@ttomsu @gregturn PTAL

gregturn:
Any benefit using Lombok to simplify some of this (given we don't have Groovy beans)?

At a glance it looks okay. Should we push this onto a branch for each component, and then we can all test that branch?

cfieber:
Not super interested in adding lombok to the mix, the config classes are simple enough if a bit boilerplatey

kork-stackdriver/src/main/java/com/netflix/spinnaker/config/StackdriverConfig.java
```diff
@@ -22,72 +22,147 @@
...
 
+import org.springframework.core.env.Environment;
 import rx.Observable;
 import rx.Scheduler;
 import rx.schedulers.Schedulers;
 
 
 @Configuration
-@EnableConfigurationProperties
+@EnableConfigurationProperties({StackdriverConfig.SpectatorStackdriverConfigurationProperties.class, StackdriverConfig.StackdriverConfigurationHints.class})
```
ewiseblatt: It appears that this does NOT work. When I use spinnaker-dependencies version 0.60.0 with this kork  branch swapped in (and my spectator https://github.com/Netflix/spectator/pull/335) I get the following:

DependencyException: Error creating bean with name 'registryInitializer' defined in class path resource [com/netflix/spinnaker/kork/metrics/SpectatorConfiguration.class]: Unsatisfied dependency expressed through constructor argument with index 1 of type [com.netflix.spinnaker.kork.metrics.SpectatorConfigurationProperties]: : No qualifying bean of type [com.netflix.spinnaker.kork.metrics.SpectatorConfigurationProperties] found for dependency: expected at least 1 bean which qualifies as autowire candidate for this dependency. Dependency annotations: {}; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type [com.netflix.spinnaker.kork.metrics.SpectatorConfigurationProperties] found for dependency: expected at least 1 bean which qualifies as autowire candidate for this dependency. Dependency annotations: {}

Some services started up, but a few gave the above (or seemed to be the same error). I can provide more info if you need.

ewiseblatt: Let me reverify kork/master to confirm that works. I might have picked up another update there when I pulled your branch.

ewiseblatt: Note that I have a config property in spinanker/config/igor.yml. Be sure you also have that set if you try to replicate. There is a chance that config property is causing the problem since some of those that failed were setting that property and some of those that didnt were not. But I dont know for certain if that is it.

ewiseblatt: Once I merely add spectator.stackdriver.enabled=true (no other properties) it breaks with the error. Without it it starts up (but bypasses stackdriver)

kork-core/src/main/java/com/netflix/spinnaker/kork/metrics/SpectatorConfigurationProperties.java
```diff
@@ -0,0 +1,49 @@
...
+package com.netflix.spinnaker.kork.metrics;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.NestedConfigurationProperty;
+
+/**
+ * SpectatorConfigurationProperties.
+ */
+@ConfigurationProperties("spectator")
+public class SpectatorConfigurationProperties {
```
ttomsu: Is it necessary to have the nested config here? Could we just make an outer `@ConfigurationProperties("spectator.gc")`?

kork-stackdriver/src/main/java/com/netflix/spinnaker/config/StackdriverConfig.java
```diff
@@ -22,72 +22,147 @@
...
+          this.path = path;
+        }
+
+        public String getPath() {
+          return path;
+        }
+
+        public void setPath(String path) {
+          this.path = path;
+        }
```
ttomsu: I miss groovy... So. Much. Boilerplate.

ewiseblatt: This module was intended for spectator which isnt groovy. For various reasons needed to put it here for the time being. Given the recent enlightenment on how touch dependency management is within this codebase, I'm willing to abandon that intent and just put this in kork. That would bring Groovy into play. However I'd rather not do that in the short term scope of cameron's PR lifetime since I'm working on other things.

kork-core/src/main/java/com/netflix/spinnaker/kork/metrics/SpectatorConfigurationProperties.java
gregturn: @ConfigurationProperties definitely supports doing that.

kork-stackdriver/src/main/java/com/netflix/spinnaker/config/StackdriverConfig.java
gregturn: @ttomsu When it's Java, I like Lombok. But Cameron's not pulling that in for such a small window of opportunity.

cfieber:
@ttomsu I'm removing `@Autowired` from `@Configuration` because it complicates the component initialization which I feel opens us up to ongoing breakages as we update spring versions and there is almost always no real need to do it other than eliminating repeated `@Bean` method parameters

ttomsu:
My question is more "is it a strict requirement to upgrade Boot"? It's a common practice to include @Autowired fields in runtime-determined @Configuration classes, such as https://github.com/ttomsu/fiat/blob/b0df75cc72a17097a8b66aad48997ccf910b10fe/fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/ResourcesConfig.java#L36-L46

gregturn:
The new recommended approach is to use constructor injection when possible. Spring 4.3 supports autowiring via the constructor without requiring an @Autowired annotation on the constructor PROVIDED there is one clear constructor call to make. 

This is more a style and not a requirement. It forces the code to be in a consistent state of initialization. But @Autowired field injection still works just fine.

I think the fuzzy issue that has beleaguered the code is @Value annotations inside @Configuration classes, especially if there are conditions that could delay WHEN those components are created. I haven't crafted a simple example to reproduce it, but I plan to try. Because I still think Spring Framework proper should trap what would appear to be a backward compatibility breaking issue, or at least have more suitable documentation around the entire use case.

kork-stackdriver/src/main/java/com/netflix/spinnaker/config/StackdriverConfig.java
cfieber: @ewiseblatt  So I can't seem to repro this... I'm using a local build of kork (this PR) in a local build of clouddriver (spinnaker/clouddriver#1170) and if I set 
```

ewiseblatt:
I'm still seeing errors.
I sync'd to spinnaker-dependencies HEAD, which is at 0.63.0 so pushing 0.64.0-SNAPSHOT when I add your kork 1.78.0-SNAPSHOT in and I run everything at 0.64.0-SNAPSHOT.

Gate seems ok.

Clouddriver is failing with redis but could be unrelated even though I just brought down an instead from HEAD that was working fine.

Most fail with something like this
2016-10-26 22:02:20.920  WARN 14609 --- [           main] o.s.boot.SpringApplic\
ation               : Error handling failed (Error creating bean with name 'dis\
coveryStatusPoller' defined in class com.netflix.spinnaker.orca.eureka.Discover\
yPollingConfiguration$NoDiscoveryConfiguration: Initialization of bean failed; \
nested exception is org.springframework.beans.factory.BeanCreationException: Er\
ror creating bean with name 'org.springframework.security.config.annotation.met\
hod.configuration.GlobalMethodSecurityConfiguration': Injection of autowired de\
pendencies failed; nested exception is org.springframework.beans.factory.BeanCr\
eationException: Could not autowire method: public void org.springframework.sec\
urity.config.annotation.method.configuration.GlobalMethodSecurityConfiguration.\
setPermissionEvaluator(java.util.List); nested exception is org.springframework\
.beans.factory.BeanCreationException: Error creating bean with name 'fiatPermis\
sionEvaluator': Injection of autowired dependencies failed; nested exception is\
 org.springframework.beans.factory.BeanCreationException: Could not autowire fi\
eld: private com.netflix.spinnaker.fiat.shared.FiatService com.netflix.spinnake\
r.fiat.shared.FiatPermissionEvaluator.fiatService; nested exception is org.spri\
ngframework.beans.factory.BeanCreationException: Error creating bean with name \
'fiatAuthenticationConfig': Injection of autowired dependencies failed; nested \
exception is org.springframework.beans.factory.BeanCreationException: Could not\
 autowire field: private com.fasterxml.jackson.databind.ObjectMapper com.netfli\
x.spinnaker.config.FiatAuthenticationConfig.objectMapper; nested exception is o\
rg.springframework.beans.factory.BeanCreationException: Error creating bean wit\
h name 'org.springframework.boot.autoconfigure.data.rest.SpringBootRepositoryRe\
stMvcConfiguration': Injection of autowired dependencies failed; nested excepti\
on is org.springframework.beans.factory.BeanCreationException: Could not autowi\
re field: private org.springframework.http.converter.json.Jackson2ObjectMapperB\
uilder org.springframework.boot.autoconfigure.data.rest.SpringBootRepositoryRes\
tMvcConfiguration.objectMapperBuilder; nested exception is org.springframework.\
beans.factory.BeanCreationException: Error creating bean with name 'jacksonObje\
ctMapperBuilder' defined in class path resource [org/springframework/boot/autoc\
onfigure/jackson/JacksonAutoConfiguration$JacksonObjectMapperBuilderConfigurati\
on.class]: Bean instantiation via factory method failed; nested exception is or\
g.springframework.beans.BeanInstantiationException: Failed to instantiate [org.\
springframework.http.converter.json.Jackson2ObjectMapperBuilder]: Factory metho\
d 'jacksonObjectMapperBuilder' threw exception; nested exception is java.lang.I\
llegalStateException: org.springframework.boot.context.embedded.AnnotationConfi\
-UU-:----F1  orca.log       73% L??    (Fundamental) ---------------------------

org.springframework.beans.factory.UnsatisfiedDependencyException: Error creatin\
g bean with name 'registryInitializer' defined in class com.netflix.spinnaker.k\
ork.metrics.SpectatorConfiguration: Unsatisfied dependency expressed through co\
nstructor argument with index 1 of type [com.netflix.spinnaker.kork.metrics.Spe\
ctatorConfigurationProperties]: : No qualifying bean of type [com.netflix.spinn\
aker.kork.metrics.SpectatorConfigurationProperties] found for dependency: expec\
ted at least 1 bean which qualifies as autowire candidate for this dependency. \
Dependency annotations: {}; nested exception is org.springframework.beans.facto\
ry.NoSuchBeanDefinitionException: No qualifying bean of type [com.netflix.spinn\
aker.kork.metrics.SpectatorConfigurationProperties] found for dependency: expec\
ted at least 1 bean which qualifies as autowire candidate for this dependency. \
Dependency annotations: {}
        at org.springframework.beans.factory.support.ConstructorResolver.create\
ArgumentArray(ConstructorResolver.java:749)
        at org.springframework.beans.factory.support.ConstructorResolver.instan\
tiateUsingFactoryMethod(ConstructorResolver.java:464)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBea\
nFactory.instantiateUsingFactoryMethod(AbstractAutowireCapableBeanFactory.java:\
1117)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBea\
nFactory.createBeanInstance(AbstractAutowireCapableBeanFactory.java:1012)

ewiseblatt:
Thanks. I havent gotten to checking yet probably will not until later this
evening.

cfieber:
@ewiseblatt I've verified that my last push fixed the init problem you were seeing (Error creating `registryInitializer` bean)

ewiseblatt:
Sorry, I havent gotten to looking at this yet. I'm in the midst of
debugging a subtle front50 problem. I ultimately found it but noticed some
other problems along the way. I want to reach a conclusion on that before I
lose my context. I still expect to get to this before tomorrow.

ewiseblatt:
I confirmed this seems to be working.
Thanks for fixing it.

PR #64, opened 2016-11-05
Set Stackdriver user agent.

ewiseblatt:
@duftler


kork-stackdriver/src/main/java/com/netflix/spectator/stackdriver/ConfigParams.java
```diff
@@ -145,9 +145,10 @@ public ConfigParams build() {
           JsonFactory jsonFactory = JacksonFactory.getDefaultInstance();
           GoogleCredential credential = loadCredential(transport, jsonFactory,
                                                        credentialsPath);
+          String version = getClass().getPackage().getImplementationVersion();
```
cfieber: maybe want a default value here ("unknown?") - I think that can return null

ewiseblatt:
@cfieber ping
@duftler FYI if you didnt see it, this is the change I mentioned in the recent thread.

cfieber:
LGTM

PR #67, opened 2016-11-18
Rewrote stackdriver integration

ewiseblatt:
Instead of being optimistic that I can create informed Metric Descriptors specifying
the labels that will be needed for the different use cases, this approach is
pessimistic and assumes that the definition is eventually going to be discovered
to be wrong and fixes the definition when it encounters a problem.

I'm not convinced that this will not lose data -- it does not at the moment,
however the issue exists either way. At least this strategy adds resilience
and makes it easier to have a judas deployment easily flush out any new labels.
The underlying descriptors can be extracted to proactively define them for new
deployments as a better starting point.

This removes the need for actually caching descriptors or having "hints" as a
workaround.

@lwander 
This is similar to the python server changes I sent you the other day, though a little different because it is interacting with spectator directly.

### DIFF
--- a/kork-stackdriver/src/main/java/com/netflix/spectator/stackdriver/MetricDescriptorCache.java
+++ b/kork-stackdriver/src/main/java/com/netflix/spectator/stackdriver/MetricDescriptorCache.java
@@ -17,7 +17,6 @@
 package com.netflix.spectator.stackdriver;
 
 import com.google.api.services.monitoring.v3.Monitoring;
-import com.google.api.services.monitoring.v3.model.ListMetricDescriptorsResponse;
 import com.google.api.services.monitoring.v3.model.LabelDescriptor;
 import com.google.api.services.monitoring.v3.model.MetricDescriptor;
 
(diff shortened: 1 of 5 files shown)


kork-stackdriver/src/main/java/com/netflix/spectator/stackdriver/MetricDescriptorCache.java
```diff
@@ -793,4 +224,68 @@ public boolean meterIsTimer(Registry registry, Meter meter) {
         }
         });
   }
+
+  /**
+   * Determine the Stackdriver Custom Metric Desctiptor Kind to use.
+   */
+  public String descriptorTypeToKind(String descriptorType, Registry registry, Meter meter) {
+    return typeToKind.computeIfAbsent(descriptorType, k -> {
+            if (meterIsTimer(registry, meter)) {
```
lwander: Why not have this logic in `meterToKind(..)`?

kork-stackdriver/src/main/java/com/netflix/spectator/stackdriver/MetricDescriptorCache.java
```diff
@@ -793,4 +224,68 @@ public boolean meterIsTimer(Registry registry, Meter meter) {
...
+    descriptor.getLabels().add(labelDescriptor);
+
+    /* Catch but ignore errors. Assume there is another process making mods too.
+     * We may overwrite theirs, but they'll just have to make them again later.
+     */
+    try {
+      log.info("Deleting existing stackdriver descriptor {}", descriptorName);
+      service.projects().metricDescriptors().delete(descriptorName).execute();
+    } catch (IOException ioex) {
+        log.info("Ignoring error " + ioex);
```
lwander: Indentation is off

kork-stackdriver/src/main/java/com/netflix/spectator/stackdriver/MetricDescriptorCache.java
```diff
@@ -793,4 +224,68 @@ public boolean meterIsTimer(Registry registry, Meter meter) {
...
+    }
+
+    try {
+      log.info("Adding new definition for {}", descriptorName);
```
lwander: Shouldn't this read "adding new descriptor for {}"?

kork-stackdriver/src/main/java/com/netflix/spectator/stackdriver/MetricDescriptorCache.java
```diff
@@ -793,4 +224,68 @@ public boolean meterIsTimer(Registry registry, Meter meter) {
...
+              return "CUMULATIVE";
+            }
+            return meterToKind(registry, meter);
+        });
+  }
+
+  /**
+   * Hack another label into an existing Stackdriver Custom Metric Descriptor.
+   *
+   * This may lose historic time series data.
```
lwander: It's not clear to me why adding labels could lose historical data.

kork-stackdriver/src/main/java/com/netflix/spectator/stackdriver/MetricDescriptorCache.java
```diff
@@ -793,4 +224,68 @@ public boolean meterIsTimer(Registry registry, Meter meter) {
...
+      }
+    }
+
+    LabelDescriptor labelDescriptor = new LabelDescriptor();
+    labelDescriptor.setKey(newLabel);
+    labelDescriptor.setValueType("STRING");
+    descriptor.getLabels().add(labelDescriptor);
+
+    /* Catch but ignore errors. Assume there is another process making mods too.
+     * We may overwrite theirs, but they'll just have to make them again later.
```
lwander: How do we know they will correctly make theirs again later? Unless does this periodically try to reinsert this label?

kork-stackdriver/src/main/java/com/netflix/spectator/stackdriver/MetricDescriptorCache.java
ewiseblatt: because to add a label requires deleting the old descriptor and recreating it.
It isnt clear this is even supported and certainly could go south if the label were bad, program crashed, etc. It seems to work in practice, but not so long ago deleting descriptors caused historic data to be deleted. Currently deleting the descriptor loses access to the data, but creating the new descriptor brings it back.

kork-stackdriver/src/main/java/com/netflix/spectator/stackdriver/MetricDescriptorCache.java
ewiseblatt: Stackdriver will create the descriptor on demand. There are different types of race conditions here. However if bad things happen, this heuristic will eventually restore it if the processes are still active. An obscure label could be lost for a while until it is needed again. Depending on the interface du'jour you might lose the ability to see it in the UI. The blank caveat emptor covers all that. I expect stackdriver to eventually fix their design to accomodate evolving metrics.

kork-stackdriver/src/main/java/com/netflix/spectator/stackdriver/MetricDescriptorCache.java
ewiseblatt: oops

PR #70, opened 2016-12-09
Always store stackdriver metrics unique per application.

ewiseblatt:
Also, change the default instanceid to be <server>:<port> which is more
readable than a UUID but still unique at any given point in time.

@ttomsu


ttomsu:
LGTM

PR #75, opened 2017-03-10
fix(tls) - configure tomcat TLS settings

cfieber:
uses the same tls versions and cipher suites as used to configure okhttp clients

### DIFF
--- a/kork-web/src/main/groovy/com/netflix/spinnaker/config/TomcatConfiguration.groovy
+++ b/kork-web/src/main/groovy/com/netflix/spinnaker/config/TomcatConfiguration.groovy
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.config
 
+import com.netflix.spinnaker.okhttp.OkHttpClientConfigurationProperties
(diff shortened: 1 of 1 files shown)


kork-web/src/main/groovy/com/netflix/spinnaker/config/TomcatConfiguration.groovy
```diff
@@ -45,7 +46,8 @@ class TomcatConfiguration {
    */
   @Bean
   @ConditionalOnExpression('${server.ssl.enabled:false}')
-  EmbeddedServletContainerCustomizer containerCustomizer(TomcatConfigurationProperties tomcatConfigurationProperties,
+  EmbeddedServletContainerCustomizer containerCustomizer(OkHttpClientConfigurationProperties okHttpClientConfigurationProperties,
```
cfieber: It feels a bit dirty to inject the Ok config here (and supercede some of what could come in from server.ssl config) but we've already defined a configuration namespace for SSL configuration and a good set of defaults, and I didn't really want to duplicate that

robzienert:
👍 Seems fine to me. Not the worst to reuse the Ok config.

PR #77, opened 2017-03-27
fix(ssl): improve TLS configuration

cfieber:
* update default cipher suites to mozilla modern TLS recommendation.
* prefer server cipher suite order
* add jdk.tls System properties to Tomcat initializer (reject client renegotiation and increase DH key size)


kork-web/src/main/groovy/com/netflix/spinnaker/config/TomcatConfiguration.groovy
```diff
@@ -49,6 +49,9 @@ class TomcatConfiguration {
   EmbeddedServletContainerCustomizer containerCustomizer(OkHttpClientConfigurationProperties okHttpClientConfigurationProperties,
                                                          TomcatConfigurationProperties tomcatConfigurationProperties,
                                                          SslExtensionConfigurationProperties sslExtensionConfigurationProperties) throws Exception {
+    System.setProperty("jdk.tls.rejectClientInitiatedRenegotiation", "true")
```
cfieber: I had added these in gate Main - relocating here so allthethings will pick them up

robzienert:
👍 Down with RC4.

PR #78, opened 2017-04-24
[chore]: change kork-security classes to Java

robfletcher:
The `AuthenticatedRequest` API was a pain to use from non-Groovy code as it exposes `Closure` in its API so I went crazy and converted the class to Java.


robfletcher:
There shouldn't be because anything that was passing a Closure there will continue to do so and any closed over scope will be contained in the caller. It just means if we're building anything using non-Groovy languages we don't have a direct compile-time dependency on Groovy.

PR #84, opened 2017-07-11
feat(web): Generic exception handling

ajordens:
Ported from spinnaker/gate#423

### DIFF
--- a/kork-web/src/main/java/com/netflix/spinnaker/kork/web/exceptions/GenericExceptionHandlers.java
+++ b/kork-web/src/main/java/com/netflix/spinnaker/kork/web/exceptions/GenericExceptionHandlers.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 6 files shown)


kork-web/src/main/java/com/netflix/spinnaker/kork/web/exceptions/GenericExceptionHandlers.java
```diff
@@ -0,0 +1,62 @@
...
+
+  @ExceptionHandler(Exception.class)
+  public void handleException(Exception e, HttpServletResponse response, HttpServletRequest request) throws IOException {
+    logger.error("Internal Server Error", e);
+
+    storeException(request, response, e);
+
+    ResponseStatus responseStatus = e.getClass().getAnnotation(ResponseStatus.class);
+    if (responseStatus != null) {
+      response.sendError(responseStatus.value().value(), e.getMessage());
```
cfieber: worth including responseStatus.reason() here if present?

ajordens: Debated doing that but `.reason()` is almost certainly worse than the message on the exception itself.

As middle ground I'll use `.reason()` iff the exception doesn't have a message itself (unlikely but who knows!)

PR #87, opened 2017-07-15
fix(web): Support @ResponseStatus on parent classes/interfaces

ajordens:
Previously the GenericExceptionHandler only looked for the `@ResponseStatus`
annotation on the exception class itself.

With this change set, the handler will look on the exception class
_as well as_ any implemented interfaces or super classes.

An HTTP 500 will be returned iff no `@ResponseStatus` annotation is found.

### DIFF
--- a/kork-web/src/main/java/com/netflix/spinnaker/kork/web/exceptions/GenericExceptionHandlers.java
+++ b/kork-web/src/main/java/com/netflix/spinnaker/kork/web/exceptions/GenericExceptionHandlers.java
@@ -21,6 +21,7 @@ import com.google.common.io.CharStreams;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.boot.autoconfigure.web.DefaultErrorAttributes;
+import org.springframework.core.annotation.AnnotationUtils;
 import org.springframework.http.HttpStatus;
 import org.springframework.security.access.AccessDeniedException;
 import org.springframework.web.bind.annotation.ControllerAdvice;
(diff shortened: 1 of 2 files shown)


kork-web/src/test/groovy/com/netflix/spinnaker/kork/web/exceptions/GenericExceptionHandlersSpec.groovy
```diff
@@ -39,23 +38,74 @@ class GenericExceptionHandlersSpec extends Specification {
...
-    new MyException("My Message") || "My Message"
+    exception                                || expectedStatusCode || expectedReason
+    new E1()                                 || 404                || "Default Reason!"
+    new E1("")                               || 404                || "Default Reason!"
+    new E1("  ")                             || 404                || "Default Reason!"
+    new E1("E1 Reason")                      || 404                || "E1 Reason"
+    new E2()                                 || 404                || "Default Reason!"
+    new E2("E2 Reason")                      || 404                || "E2 Reason"
+    new E3()                                 || 404                || "Default Reason!"
+    new E4()                                 || 400                || "My Other Reason!" // favor @ResponseStatus on interface over super class
```
ajordens: It's unlikely we'll have an `Interface` in our exception hierarchy but _it could happen_.

In such a scenario, we'll use the `@ResponseStatus` on the interface rather than one on a super class.

kork-web/src/main/java/com/netflix/spinnaker/kork/web/exceptions/GenericExceptionHandlers.java
```diff
@@ -94,21 +96,49 @@ public void handleRetrofitError(RetrofitError e, HttpServletResponse response, H
 
   @ExceptionHandler(Exception.class)
   public void handleException(Exception e, HttpServletResponse response, HttpServletRequest request) throws IOException {
-    logger.error("Internal Server Error", e);
-
     storeException(request, response, e);
 
-    ResponseStatus responseStatus = e.getClass().getAnnotation(ResponseStatus.class);
+    ResponseStatus responseStatus = getAncestors(e.getClass())
```
cfieber: worth using some of the stuff in here?

http://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/annotation/AnnotatedElementUtils.html

Would give consistency with the default framework behaviour

ajordens:
That's fair ... I'll check the spring docs as soon as the cloudflare errors go away.

ajordens:
`AnnotationElementUtils` is rather weak sauce in Spring 4.1.9 so not really a proper fit.

Revised to use `AnnotationUtils.findAnnotation()`, existing tests pass so the behavior should be as expected.

PR #94, opened 2017-08-30
feat(artifacts): Added generic artifact model class.

jtk54:
@gardleopard @spinnaker/netflix-reviewers FYI. Required to continue pub/sub work in Echo, needs a release and Kork version bump there.


kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/GenericArtifact.java
```diff
@@ -0,0 +1,40 @@
...
+@NoArgsConstructor
+@AllArgsConstructor
+public class GenericArtifact {
+
+  private String type;
+  private String name;
+  private String version;
+  private String reference;
+  private Map<String, String> metadata;
+  private Map<String, Object> auth;
```
lwander: `string -> object` seems overly ambitious here... I think all we need is a "storage account" name that will be used to fetch the artifact.

lwander: @ewiseblatt what were you thinking of calling the storage/artifact accounts? let's use that name here.

jtk54: Heh that was the only field I wasn't sure about.

ewiseblatt: I didnt have a name. We were calling them storage accounts for the time being. "Accounts" might be overloaded because these are not "deployment" accounts in providers.

ewiseblatt: @lwander I dont know what "auth" is for, but when you say to "fetch" the artifact then that is implying a context/use case but this sounds like a generic data model. Woudnt those credentials be given in the request? The artifact would need some protections that fiat could map to accounts in a one-to-many if not many-to-many manner (e.g. access roles on the resource and particular accounts that have those access roles).

ewiseblatt: Or the artifact isnt protected here but relies on the intrinsic system security for the substrated it is stored in.

lwander: By putting the credentials into the request do you mean something like username/password? That feels dangerous.

If the artifact (like a GCS file) isn't publicly readable, we need some abstraction to download the contents as well as check the permissions against fiat. Having some name, e.g. artifactAccount, act as a reference to some fiat-protected credentials solves this.

jtk54:
I think we arrived on a name and type reasonable enough for a first pass.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/GenericArtifact.java
ewiseblatt: not the actual credentials, but a reference to them (e.g. storage account name)

PR #99, opened 2017-09-27
feat(artifacts): Adds location field

lwander:


kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/Artifact.java
```diff
@@ -31,6 +31,7 @@
   private String type;
   private String name;
   private String version;
+  private String location;
```
jtk54: What is the distinction between `location` and `reference`?

lwander: Reference is a URI, location is namespace, region, zone, etc...

PR #100, opened 2017-09-27
feat(artifacts): Expected artifact definition

lwander:
I wound up leaving the artifact resolution out of this since I think the validator leaves the usage clear-enough.

### DIFF
--- a/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/ExpectedArtifact.java
+++ b/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/ExpectedArtifact.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/ExpectedArtifact.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+@Data
+public class ExpectedArtifact {
+  private List<ArtifactField> fields = new ArrayList<>();
+
+  private static Set<String> artifactFields = Arrays.stream(Artifact.class.getFields())
```
jtk54: Probably `final` as well.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/ExpectedArtifact.java
```diff
@@ -0,0 +1,101 @@
...
+        case MUST_MATCH:
+          matchCondition = true;
+          break;
+        case FIND_IF_MISSING:
+          if (field.getMissingPolicy() == null) {
+            throw new IllegalStateException("When fieldType == FIND_IF_MISSING, a policy must be provided");
+          }
+
+          switch (field.getMissingPolicy()) {
+            case EXPRESSION:
```
jtk54: So the idea is that we'd `validate` before a pipeline runs and then rely on Orca to to the fetching/resolving in the case of `PRIOR_PIPELINE`, etc?

lwander: We'd likely validate before a pipeline is saved.

PR #105, opened 2017-10-13
feat(artifacts): Simplify artifact matching

lwander:


lwander:
The idea is now
1. Match against incoming artifact using supplied method
2. If no match is found & usePriorArtifact is true, check the prior pipeline
3. If no match is found & defaultArtifact is specified, use that
4. If no match is found (either 2 & 3 are disabled or failed) fail the pipeline

lwander:
I'll tie this into echo next & then update the UI

PR #110, opened 2017-11-02
fix(artifacts): metadata must be `Map<String, Object>`

ajordens:


ajordens:
This should address what was identified @ https://github.com/spinnaker/orca/commit/6b91aa880d5232111265e844b6efc70bf19701b7#commitcomment-25347949

jtk54:
LGTM.

PR #119, opened 2018-01-18
feat(web): Adding basic retrofit client spectator profiler

robzienert:
Would like to get some better metrics out of how long individual service calls are taking between services.

### DIFF
--- a/kork-web/src/main/java/com/netflix/spinnaker/retrofit/RetrofitSpectatorProfiler.java
+++ b/kork-web/src/main/java/com/netflix/spinnaker/retrofit/RetrofitSpectatorProfiler.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


kork-web/src/main/java/com/netflix/spinnaker/retrofit/RetrofitSpectatorProfiler.java
```diff
@@ -0,0 +1,45 @@
...
+    return null;
+  }
+
+  @Override
+  public void afterCall(RequestInformation requestInfo, long elapsedTime, int statusCode, Object beforeCallData) {
+    Id id = registry.createId(String.format("retrofit.%s.requestTimeMs", serviceName))
+      .withTag("method", requestInfo.getMethod())
+      .withTag("endpoint", requestInfo.getRelativePath())
+      .withTag("status", Integer.toString(statusCode));
+    registry.gauge(id, elapsedTime);
```
ajordens: should this be a `gauge` or `timer`?

ajordens: Will this give us anything materially different that what we can get out of the existing controller metrics? 

Is `requestInfo.getRelativePath()` going to have all parameters expanded? If so, that's going to create an awful lot of unique tags.

robzienert: The goal in this PR was to get metrics from the other side of the request/response path. Controller metrics don't record what the source is, which I think is quite important in determining where to focus efforts.

ajordens: You can get the inverse from the existing retrofit log messages ... 

(a) controller metrics tell you what is slow
(b) logs tell you who is making the slow calls

It's a two step approach but it is not impossible to do right now.

PR #120, opened 2018-01-22
fix(jedis): use JedisPools base class for delegate

asher:


kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/JedisClientDelegate.java
```diff
@@ -15,16 +15,17 @@
...
+import redis.clients.util.Pool;
 import redis.clients.jedis.*;
 
 import java.util.function.Consumer;
 import java.util.function.Function;
 
 public class JedisClientDelegate implements RedisClientDelegate {
 
-  private final JedisPool jedisPool;
+  private final Pool<Jedis> jedisPool;
```
asher: Orca uses a mix of Pool<Jedis> and JedisPool (extends Pool<Jedis>), switching to the base class here for compatibility with either.

robzienert: 💯 It turns out this is required for Sentinel support as well, since it has a `SentinelJedisPool` object that implements `Pool<Jedis>`. We should definitely be actively converting any of these cases.

PR #123, opened 2018-02-06
feat(redis): Adding jedis/dynomite client factories

robzienert:
This is mostly just a lift of clouddriver's Redis & Dynomite configs. I'd like to add another layer on top that would be Spring-y that would offer something more like the following, but also have classes (seen in this PR) that can be used to drive our current configs.

```yaml
# clouddriver.yml
cache:
  primary:
    driver: dynomite
    applicationName: clouddriver
    clusterName: dyno_clouddriver_main
  previous:
    driver: redis
    connection: redis://localhost:6379
  

taskRepository:
  primary:
    driver: redis
    connection: redis://localhost:6379
```

Rather than defining a single location for redis configs, so we can better control what we're migrating at certain times or using vanilla Redis where we don't think Dynomite may be worth the overhead, etc.

### DIFF
--- a/kork-dynomite/src/main/java/com/netflix/spinnaker/kork/dynomite/DynomiteClientFactory.java
+++ b/kork-dynomite/src/main/java/com/netflix/spinnaker/kork/dynomite/DynomiteClientFactory.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.dynomite;
+
+import com.netflix.discovery.DiscoveryClient;
+import com.netflix.dyno.connectionpool.impl.ConnectionPoolConfigurationImpl;
+import com.netflix.dyno.jedis.DynoJedisClient;
+import com.netflix.spinnaker.kork.jedis.exception.MissingRequiredConfiguration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Optional;
+
+public class DynomiteClientFactory {
+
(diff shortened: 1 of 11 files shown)


kork-dynomite/src/main/java/com/netflix/spinnaker/kork/dynomite/DynomiteClientFactory.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    Optional<DiscoveryClient> discovery = getDiscoveryClient();
+    if (discovery.isPresent()) {
+      builder.withDiscoveryClient(discovery.get());
+    } else {
+      connectionPoolConfiguration
+        .withTokenSupplier(new StaticTokenMapSupplier(properties.getDynoHostTokens()))
+        .setLocalDataCenter(properties.localDatacenter)
+        .setLocalRack(properties.localRack);
```
asher: For a static host config, I think `.withHashTag()` should be added to the CPC if one is specified in config.

kork-dynomite/src/main/java/com/netflix/spinnaker/kork/dynomite/DynomiteConfigurationProperties.java
```diff
@@ -0,0 +1,69 @@
...
+
+  static class DynoHost {
+    public String hostname;
+    public String ipAddress;
+    public int port = DEFAULT_PORT;
+    public Status status = Status.Up;
+    public String rack = LOCAL_RACK;
+    public String datacenter = LOCAL_DATACENTER;
+    public Long token = 1000000L;
+    public String hashtag = "{}";
```
asher: May want to make this a root level property since it is also set on the connection pool and shouldn't differ across hosts.

kork-dynomite/src/main/java/com/netflix/spinnaker/kork/dynomite/DynomiteHealthIndicatorFactory.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package com.netflix.spinnaker.kork.dynomite;
+
+import org.springframework.boot.actuate.health.Health;
+import org.springframework.boot.actuate.health.HealthIndicator;
+
+public class DynomiteHealthIndicatorFactory {
+
+  public static HealthIndicator build() {
+    // TODO rz - Haven't really figured out a good healthcheck strategy for dynomite yet.
```
asher: I'm tempted to request a d_ping command that fans out to all hosts and returns pong only if $quroum instances per shard respond.

robzienert: Good thought. I don't care to open any of the other admin commands, ping would be handy, though.

PR #124, opened 2018-02-07
feat(redis): Distributed lock recipe

robzienert:
Updated from early commit-only feedback. Also altered the algo a little bit to be more Redis-native. Tests on the way.

### DIFF
--- a/kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/recipe/RedisDistributedLock.java
+++ b/kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/recipe/RedisDistributedLock.java
@@ -1,0 +1,374 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.jedis.recipe;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.spectator.api.Id;
+import com.netflix.spectator.api.Registry;
+import com.netflix.spinnaker.kork.jedis.RedisClientDelegate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.time.Clock;
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.UUID;
+
+import static com.netflix.spinnaker.kork.jedis.recipe.RedisDistributedLock.LockStatus.*;
+
+/**
+ * Simple distributed lock recipe that supports both Redis & Dynomite.
+ *
+ * When a lock is acquired, it's expected that the owner of the lock will continue to
+ * send heartbeats before the end of its lease duration to continue extending the lock
+ * as long as it needs. Upon completing work, a client can either stop sending heartbeats,
+ * or explicitly release the lock. If a release is not called, the lock will expire at
+ * the end of lease.
+ *
+ * Each operation will take one or two version numbers: The Previous Record Version (PRV)
+ * and the New Record Version (NRV). If a PRV is required, the action will only succeed
+ * if the PRV matches the version number stored with the lock in Redis. When an action
+ * succeeds, the version number inside of Redis will be updated to the NRV.
+ *
+ * If a client has a lock and heartbeats, but the heartbeat fails due to a transfer
+ * of ownership, the client must stop performing work against the locked resource(s).
+ * This behavior is not enforced by this lock class.
+ *
+ * TODO rz - total lock duration timer metric (somehow? lock expiration is db-side)
+ */
+public class RedisDistributedLock {
+
+  /**
+   * Returns the active lock, whether or not the desired lock was acquired.
+   *
+   * ARGS
+   * 1: leaseDurationMillis
+   * 2: owner
+   * 3: ownerSystemTimestamp
+   * 4: newRecordVersion
+   */
+  private static String ACQUIRE_SCRIPT = "" +
+    "local payload = cjson.encode({" +
+    "  ['leaseDurationMillis']=ARGS[1]," +
+    "  ['owner']=ARGS[2]," +
+    "  ['ownerSystemTimestamp']=ARGS[3]," +
+    "  ['version']=ARGS[4]," +
+    "})" +
+    "if redis.call('SET', KEYS[1], payload, 'NX', 'EX', ARGS[1]) == 'OK' then" +
+    "  return payload" +
+    "end" +
+    "return redis.call('GET', KEYS[1])";
+
+  /**
+   * Returns 1 if heartbeat was successful, -1 if the lock no longer exists,
+   * 0 if the lock is now owned by someone else or is a different version.
+   *
+   * If the heartbeat is successful, update the lock with the NRV and updated
+   * owner system timestamp.
+   *
+   * ARGS
+   * 1: owner
+   * 2: previousRecordVersion
+   * 3: newRecordVersion
+   * 4: updatedOwnerSystemTimestamp
+   */
+  private static String HEARTBEAT_SCRIPT = "" +
+    "if redis.call('EXISTS', KEYS[1]) == 1 then" +
+    "  local payload = redis.call('GET', KEYS[1])" +
+    "  local lock = cjson.decode(payload)" +
+    "  if lock['owner'] == ARGS[1] and lock['version'] == ARGS[2] then" +
+    "    lock['version'] = ARGS[3]" +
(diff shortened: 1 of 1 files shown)


kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/recipe/RedisDistributedLock.java
```diff
@@ -0,0 +1,374 @@
...
+    this.registry = registry;
+    this.clock = clock;
+
+    acquireId = registry.createId("redis.distributedLock.acquire");
+    heartbeatId = registry.createId("redis.distributedLock.heartbeat");
+    releaseId = registry.createId("redis.distributedLock.release");
+  }
+
+  /**
+   * Does not need to use a PVN, as the lock will only be acquired if the key does
```
dreynaud: PVN... PRV?

robzienert: ono. I didn't find/replace good enough. PRV is the thing.

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/recipe/RedisDistributedLock.java
```diff
@@ -0,0 +1,374 @@
...
+import java.time.Clock;
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.UUID;
+
+import static com.netflix.spinnaker.kork.jedis.recipe.RedisDistributedLock.LockStatus.*;
+
+/**
+ * Simple distributed lock recipe that supports both Redis & Dynomite.
```
dreynaud: is this a common pattern or an original recipe? I find that the heartbeat and record versions are a bit hard to wrap my head around

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/recipe/RedisDistributedLock.java
```diff
@@ -0,0 +1,374 @@
...
+   * owner system timestamp.
+   *
+   * ARGS
+   * 1: owner
+   * 2: previousRecordVersion
+   * 3: newRecordVersion
+   * 4: updatedOwnerSystemTimestamp
+   */
+  private static String HEARTBEAT_SCRIPT = "" +
+    "if redis.call('EXISTS', KEYS[1]) == 1 then" +
```
dreynaud: couldn't this be a single redis call? Something like

  local payload = redis.call('GET', KEYS[1])
  if payload != null then
    local lock = cjson.decode(payload)
    ...

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/recipe/RedisDistributedLock.java
```diff
@@ -0,0 +1,374 @@
...
+        Collections.singletonList(lockKey(normalizedName)),
+        Arrays.asList(
+          ownerName,
+          previousVersion,
+          newVersion,
+          Long.toString(clock.millis())
+        )
+      ).toString());
+
+      Id lockHeartbeat = heartbeatId.withTag("lockName", normalizedName);
```
dreynaud: shouldn't the counter for this be incremented?

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/recipe/RedisDistributedLock.java
asher: If I understand correctly, redlock's complexity is to properly support multi-master redis environments. We've avoided that pattern operating redis for spinnaker at nflx and while we're moving to dynomite, I don't think it has the problem redlock is trying to solve (_"In the distributed version of the algorithm we assume we have N Redis masters. Those nodes are totally independent, so we don’t use replication or any other implicit coordination system."_)

If true, I think we could KISS just doing something like https://redislabs.com/ebook/part-2-core-concepts/chapter-6-application-components-in-redis/6-2-distributed-locking/6-2-5-locks-with-timeouts/

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/recipe/RedisDistributedLock.java
dreynaud: never mind, I see the incremented counters below

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/recipe/RedisDistributedLock.java
```diff
@@ -0,0 +1,374 @@
...
+
+      switch (result) {
+        case 1:
+          log.info("Refreshed lease for lock " + lockVersion(normalizedName, newVersion));
+          registry.counter(lockHeartbeat.withTag("status", "SUCCESS")).increment();
+          return true;
+        case 0:
+          log.warn("Failed refreshing lock " + normalizedName + ": Not owner");
```
dreynaud: that log message is a bit misleading, it could be an owner mismatch or a version mismatch... and in any case the mismatch isn't visible. I would feel better about this if instead of relying on enum-style error codes (0, 1, -1) the HEARTBEAT_SCRIPT actually returned the content of the lock so that we could display more meaningful stuff. Does that make sense?

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/recipe/RedisDistributedLock.java
```diff
@@ -0,0 +1,374 @@
...
+import java.util.UUID;
+
+import static com.netflix.spinnaker.kork.jedis.recipe.RedisDistributedLock.LockStatus.*;
+
+/**
+ * Simple distributed lock recipe that supports both Redis & Dynomite.
+ *
+ * When a lock is acquired, it's expected that the owner of the lock will continue to
+ * send heartbeats before the end of its lease duration to continue extending the lock
+ * as long as it needs. Upon completing work, a client can either stop sending heartbeats,
```
dreynaud: aren't you concerned about a client potentially going heartbeat-happy and hogging the lock, starving everybody else? If we didn't have the heartbeat, then all nodes would get the chance to compete for the lock.

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/recipe/RedisDistributedLock.java
```diff
@@ -0,0 +1,374 @@
...
+ * or explicitly release the lock. If a release is not called, the lock will expire at
+ * the end of lease.
+ *
+ * Each operation will take one or two version numbers: The Previous Record Version (PRV)
+ * and the New Record Version (NRV). If a PRV is required, the action will only succeed
+ * if the PRV matches the version number stored with the lock in Redis. When an action
+ * succeeds, the version number inside of Redis will be updated to the NRV.
+ *
+ * If a client has a lock and heartbeats, but the heartbeat fails due to a transfer
```
dreynaud: if I acquire a lock and do things, is there an API for knowing if the lock is still valid? (e.g. how do I know if it has expired on the db?)

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/recipe/RedisDistributedLock.java
robzienert: This pattern was lifted from the DynamoDB AWS client recipe.

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/recipe/RedisDistributedLock.java
robzienert: Yes.

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/recipe/RedisDistributedLock.java
robzienert: Yes, it does. I chose the enum codes so we didn't need to inject an object mapper to deserialize the lock in Java. Can switch out of that, of course.

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/recipe/RedisDistributedLock.java
robzienert: That's a concern, yes, but for argument, the same condition of a client going heartbeat-happy, the client may also just continue to work and fall out of bounds of the lock. Having a heartbeat would allow for a client to complete its work before another client comes in to step on it. (More thoughts on next comment)...

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/recipe/RedisDistributedLock.java
dreynaud: But there's already one that parses the value out of the acquire script, no?

`authoritativeLock = objectMapper.readValue(payload.toString(), Lock.class);`

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/recipe/RedisDistributedLock.java
robzienert: No, there isn't, but there definitely should be.

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/recipe/RedisDistributedLock.java
robzienert: Additional thought here, that function would probably best also have a `cancel` method on it so if the operation doesn't complete in time, it could be killed off.

robzienert:
Closing in favor of #150

PR #128, opened 2018-02-13
refactor(aws): pull bastion config to kork

emjburns:
Pulled from https://github.com/spinnaker/clouddriver/tree/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/bastion

Since `clouddriver`, `front50`, and soon `echo` will be using this I'm pulling it here. @ajordens and I will refactor `clouddriver` and `front50` after this is merged.

### DIFF
--- a/kork-aws/kork-aws.gradle
+++ b/kork-aws/kork-aws.gradle
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+dependencies {
+  compile spinnaker.dependency('groovy')
+  compile spinnaker.dependency('bootAutoConfigure')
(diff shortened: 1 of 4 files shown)


kork-aws/kork-aws.gradle
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+dependencies {
+  compile spinnaker.dependency('groovy')
+  compile project(":kork-core")
```
ajordens: out of curiosity ... are you actually using anything from `kork-core`?

ajordens:
LGTM (just a straight copy right?)

At some point I might do a quick pass and see if it's easy enough to convert that ssh trickery to kotlin.

emjburns:
@ajordens straight copy pasta. Sounds good!

kork-aws/kork-aws.gradle
emjburns: The spring annotations are in kork-core

cfieber: think you could just add ` compile spinnaker.dependency('bootAutoConfigure')` and get the spring annotations

cfieber: and whatever bit of the aws-sdk this needs to compile as well

emjburns: @cfieber is that `compile spinnaker.dependency('aws')`, or is there a smaller subset that provides `com.amazonaws.auth.*`? Like https://github.com/spinnaker/spinnaker-dependencies/blob/master/src/spinnaker-dependencies.yml#L72?

emjburns: Updated the dependencies. Merging shortly.

PR #136, opened 2018-03-08
feat(hystrix): Java 8 support for the hystrix command wrapper

jervi:


jervi:
@spinnaker/reviewers PTAL

PR #138, opened 2018-03-12
feat(artifacts): Put "unknown" fields into the metadata map

jervi:
Artifacts from Igor contains fields that can't be directly mapped to the fields in Kork's `Artifact` class. This commit will put the data in these fields into the metadata map.

### DIFF
--- a/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/Artifact.java
+++ b/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/Artifact.java
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.kork.artifacts.model;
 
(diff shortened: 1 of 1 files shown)


kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/Artifact.java
```diff
@@ -37,4 +38,10 @@
   private String artifactAccount;
   private String provenance;
   private String uuid;
+
+  // Add extra, unknown data to the metadata map:
+  @JsonAnySetter
+  public void putMetadata(String key, Object value) {
+    metadata.put(key, value);
```
jtk54: https://github.com/spinnaker/echo/pull/240 handled the case when `metadata` was null, does that check need to go here also?

PR #140, opened 2018-03-17
feat(redis): Add scan command facade

robzienert:
Dynomite's scan methods don't implement the Jedis interfaces. This method puts a small facade around barebones functionality of scan. 

```java
client.withScan("mypattern*", 100, (scan) -> {
  log.info(
    "This is called for every scan iteration and has {} items", 
    scan.getResults().size()
  );
});
```

Something like that.

### DIFF
--- a/kork-dynomite/src/main/java/com/netflix/spinnaker/kork/dynomite/DynomiteClientDelegate.java
+++ b/kork-dynomite/src/main/java/com/netflix/spinnaker/kork/dynomite/DynomiteClientDelegate.java
@@ -15,13 +15,16 @@
  */
 package com.netflix.spinnaker.kork.dynomite;
 
+import com.netflix.dyno.connectionpool.CursorBasedResult;
 import com.netflix.dyno.connectionpool.exception.DynoException;
 import com.netflix.dyno.jedis.DynoJedisClient;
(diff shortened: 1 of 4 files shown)


kork-dynomite/src/main/java/com/netflix/spinnaker/kork/dynomite/DynomiteClientDelegate.java
```diff
@@ -147,6 +150,23 @@ public void withScriptingClient(Consumer<ScriptingCommands> f) {
     return f.apply(client);
   }
 
+  @Override
+  public void withKeyScan(String pattern, int count, Consumer<RedisScanResult> f) {
+    CursorBasedResult<String> result = null;
+
+    do {
+      if (result == null) {
+        result = client.dyno_scan(count, pattern);
```
asher: `dyno_scan(int, String)` calls `dyno_scan(null, int, string)` so should be ok to drop the if statement and always call `result = client.dyno_scan(result, count, pattern);`

robzienert: Neat, thanks for the check.

PR #141, opened 2018-03-30
feat(retry): Adds exponential base config support

robzienert:


ajordens:
The impl looks fine but tbh I don't really feel like we're going to use this very often ... at least with a base of 2 the math is sorta easy to do in your head. 

In a world where there's no guarantee everyone is sticking to the same rules (ie. all apps at Netflix talking to aws apis -- it's more than just Spinnaker), I would still err on a static backoff of a second or two for a longer period of time. 

Again, with respect to the AWS SDK, it is already retrying exponentially internally (where it decides that it can) ... so it makes sense to externally balance that with non-exponential retries.

asher:
I like it. An exponential base of 2 is easy to compute in your head but grows aggressively enough that we advise against using it. I'd be happy to see a static base of 1.5 if we don't want to make it adjustable, or retry intervals of fibonacci numbers in seconds. If there's consensus to stick with static backoff intervals, let's remove the exponential boolean option.

robzienert:
I'm open to removing the exponential flag entirely. It seems like it's not something we actually want to support, @ajordens?

PR #142, opened 2018-03-30
fix(mdc): propagate execId across threads

asher:
- This fixes an issue where ExecutionId wasn't propagating to clouddriver task threads, ensuring it's added to the User-Agent string for AWS calls.

### DIFF
--- a/kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequest.java
+++ b/kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequest.java
@@ -49,6 +49,7 @@ public class AuthenticatedRequest {
    */
   public static <V> Callable<V> propagate(Callable<V> closure, boolean restoreOriginalContext, Object principal) {
     String spinnakerUser = getSpinnakerUser(principal).orElse(null);
+    String executionId = getSpinnakerExecutionId().orElse(null);
     if (spinnakerUser == null) {
       return () -> {
         MDC.remove(SPINNAKER_USER);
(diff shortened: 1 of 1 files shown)


robzienert:
That's the good stuff.

kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequest.java
```diff
@@ -49,6 +49,7 @@
    */
   public static <V> Callable<V> propagate(Callable<V> closure, boolean restoreOriginalContext, Object principal) {
     String spinnakerUser = getSpinnakerUser(principal).orElse(null);
+    String executionId = getSpinnakerExecutionId().orElse(null);
```
ajordens: I wonder if we should just auto propagate anything in MDC that is `X-SPINNAKER-*`?

asher: Good idea. I'd like to better understand our use of principal though - theoretically, we aren't always propagating `X-SPINNAKER-USER`, in that if principal contains a valid `User` object, it's preferred over what's in MDC:

```
  public static Optional<String> getSpinnakerUser(Object principal) {
    Object spinnakerUser = MDC.get(SPINNAKER_USER);

    if (principal != null && principal instanceof User) {
      spinnakerUser = ((User) principal).getUsername();
    }
```

Do we support user delegation anywhere, or a similar usecase where these could differ? If not, we can do as you propose without special casing `spinnakerUser`.

PR #143, opened 2018-04-06
fix(artifacts): correctly deserialize artifactAccount

lwander:
Some of our default deserializers are turning camelCase into underscore_case


lwander:
Sure, might as well

PR #144, opened 2018-04-10
feat(redis): Instrumented Jedis client & Pool

robzienert:
Instruments:

* The latency for every command
* The payload size for (most) writes
* The error rate for individual commands
* The combined error rate for the client

### DIFF
--- a/kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/JedisPoolFactory.java
+++ b/kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/JedisPoolFactory.java
@@ -15,7 +15,10 @@
  */
 package com.netflix.spinnaker.kork.jedis;
 
+import com.netflix.spectator.api.NoopRegistry;
+import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.kork.jedis.exception.MissingRequiredConfiguration;
+import com.netflix.spinnaker.kork.jedis.telemetry.InstrumentedJedisPool;
 import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
 import redis.clients.jedis.Jedis;
 import redis.clients.jedis.JedisPool;
@@ -25,6 +28,16 @@ import redis.clients.util.Pool;
 import java.net.URI;
 
 public class JedisPoolFactory {
+
+  private final Registry registry;
+
+  public JedisPoolFactory() {
+    this(new NoopRegistry());
+  }
+
+  public JedisPoolFactory(Registry registry) {
+    this.registry = registry;
+  }
 
   public Pool<Jedis> build(String name, JedisDriverProperties properties) {
     if (properties.connection == null || "".equals(properties.connection)) {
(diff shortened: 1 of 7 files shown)


kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/telemetry/InstrumentedJedis.java
```diff
@@ -0,0 +1,6648 @@
...
+  public Jedis unwrap() {
+    return delegated;
+  }
+
+  @Override
+  public String set(String key, String value) {
+    String command = "set";
+    registry.distributionSummary(TelemetryHelper.payloadSizeId(idPrefix, command)).record(payloadSize(value));
+    try {
+      return registry.timer(timerId(idPrefix, command)).record(() ->
```
asher: PercentileTimer (http://netflix.github.io/spectator/en/latest/javadoc/spectator-api/com/netflix/spectator/api/histogram/PercentileTimer.html) please, it enables things `name,redis.command.set,:eq,(,50,95,99,),:percentiles` in atlas

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/JedisPoolFactory.java
```diff
@@ -39,7 +43,11 @@
     String password = parsePassword(redisConnection.getUserInfo());
     GenericObjectPoolConfig objectPoolConfig = properties.poolConfig;
 
-    return new JedisPool(objectPoolConfig, host, port, properties.timeoutMs, password, database, name);
+    return new InstrumentedJedisPool(
+      new NoopRegistry(),
```
cfieber: The one usage of JedisPoolFactory I saw in clouddriver we do `new JedisPoolFactory().build(name, properties)`

I'd be somewhat tempted to move JedisPoolFactory to a component we autowire in where we are using it, in which case making Registry a constructor arg for JedisPoolFactory and defaulting the no arg constructor to use NoopRegistry.  (Otherwise wherever we are using it we have to autowire in the Registry anyways)

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/telemetry/InstrumentedJedis.java
```diff
@@ -0,0 +1,6648 @@
...
+
+  public InstrumentedJedis(Registry registry, Jedis delegated) {
+    this(registry, delegated, DEFAULT_ID_PREFIX);
+  }
+
+
+  public InstrumentedJedis(Registry registry, Jedis delegated, String idPrefix) {
+    this.registry = registry;
+    this.delegated = delegated;
+    this.idPrefix = idPrefix;
```
cfieber: Wondering about having a separate Id for each pool vs having the pool name as a tag on each Id.

We gain the ability to do some aggregation of metrics across pools and/or some more generic dashboarding if we had more static metric names with the dynamic component being a tag

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/JedisPoolFactory.java
robzienert: That's a good idea. It's intended to be used through `@Import(JedisClientConfiguration.class)`, but in migrating to the new stuffs, some of the services are a little more artisanally wired up: https://github.com/spinnaker/kork/blob/master/kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/JedisClientConfiguration.java

I will change the signature for the factory to include the registry, however.

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/telemetry/InstrumentedJedis.java
robzienert: I don't believe the `PercentileTimer` is actually exposed through the API? All I see on our registry is `timer`. I was under the impression that timers would be automatically percentiled?

PR #146, opened 2018-04-17
feat(web): move serviceSelector to kork for general use

asher:


asher:
This is in preparation for using `ServiceSelector` in gate, in order to shard the orca queue by dimensions such as application. In this use case, we may route requests to start an execution or orchestration to an orca shard with a different queue namespace within a common redis store.

PR #148, opened 2018-04-20
support rediss:// (redis+ssl) URIs

dblackdblack:
AWS Elasticache only supports redis with AUTH when used in combination with "in-transit encryption" (a/k/a SSL).  Redis SSL URIs have the scheme `rediss://`.  Jedis already supports SSL, but we need to pass through the SSL-ness after parsing the URI.


dblackdblack:
@robzienert

dblackdblack:
ping @robzienert

PR #149, opened 2018-04-26
feat(core): TransientConfigService

robzienert:
Offers a pluggable dynamic / transient config service interface, and a default Spring-backed implementation. Two primary APIs are offered: A generic `getTransientConfig` and a `isEnabled` feature flag. Both of which support passing a `Supplier<Boolean>` predicate that must satisfy to `true` before returning the actual value. If `false`, the `defaultValue` is returned.

An additional layer of logic was added to feature flags that allows us to pass in canned `ScopedCriteria` according to our usual domain filters: `application`, `region`, `cloudProvider` and `account`. Precedence (highest to lowest) is currently: `region -> account -> cloudProvider -> application -> global`.

Given a feature flag `myFeature`, the configs for the Spring-backed implementation would be:

* `myFeature.enabled` (global)
* `myFeature.{scope}.{value}` (e.g. `myFeature.application.orca`)

Example usage:

```kotlin
// No extra logic usage
transientConfigService.isEnabled("myFeature", false)

// Scoped criteria usage
val criteria = ScopedCriteria.Builder().withApplication("orca").build()
transientConfigService.isEnabled("myFeature", false, criteria)

// Predicate usage
transientConfigService.isEnabled("myFeature", false, { myPredicationLogic() })

// Config usage
transientConfigService.getTransientConfig<String>("myConfig", "Hello!")
```

### DIFF
--- a/kork-core/kork-core.gradle
+++ b/kork-core/kork-core.gradle
@@ -10,6 +10,8 @@ dependencies {
     compile spinnaker.dependency('spectatorJvm')
     compile spinnaker.dependency('spectatorWebSpring')
 
-    testCompile spinnaker.dependency('junit')
+
+    testCompile spinnaker.dependency('groovy')
+    spinnaker.group('spockBase')
     testRuntime spinnaker.dependency('slf4jSimple')
 }
--- a/kork-core/src/main/java/com/netflix/spinnaker/kork/PlatformComponents.java
+++ b/kork-core/src/main/java/com/netflix/spinnaker/kork/PlatformComponents.java
@@ -19,12 +19,18 @@ package com.netflix.spinnaker.kork;
 import com.netflix.spinnaker.kork.archaius.ArchaiusConfiguration;
 import com.netflix.spinnaker.kork.aws.AwsComponents;
 import com.netflix.spinnaker.kork.eureka.EurekaComponents;
+import com.netflix.spinnaker.kork.transientconfig.TransientConfigConfiguration;
 import com.netflix.spinnaker.kork.metrics.SpectatorConfiguration;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingClass;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
 
 @Configuration
-@Import({ArchaiusConfiguration.class, EurekaComponents.class, SpectatorConfiguration.class, AwsComponents.class})
+@Import({
+  ArchaiusConfiguration.class,
+  TransientConfigConfiguration.class,
+  EurekaComponents.class,
+  SpectatorConfiguration.class,
+  AwsComponents.class
+})
 public class PlatformComponents {
 }
--- a/kork-core/src/main/java/com/netflix/spinnaker/kork/transientconfig/ScopedCriteria.java
+++ b/kork-core/src/main/java/com/netflix/spinnaker/kork/transientconfig/ScopedCriteria.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.transientconfig;
+
+public class ScopedCriteria {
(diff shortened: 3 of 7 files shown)


kork-core/src/main/java/com/netflix/spinnaker/kork/feature/FeatureFlagService.java
```diff
@@ -0,0 +1,26 @@
...
+package com.netflix.spinnaker.kork.feature;
+
+import javax.annotation.Nonnull;
+
+/**
+ * A simple interface for interacting with dynamic Spring properties in the scope of feature flags.
+ */
+public interface FeatureFlagService {
+
+  boolean isEnabled(@Nonnull String flagName, boolean defaultValue);
```
tomaslin: would be great if there was a list method so we can see available flags and states

robzienert: How do you suggest going about that? In the context of a Spring environment, there's really no way to flag a feature flag, unless we force all flags under the same namespace, which wouldn't be consistent with all of the flags we have today.

kork-core/src/main/java/com/netflix/spinnaker/kork/feature/SpringFeatureFlagService.java
```diff
@@ -0,0 +1,42 @@
...
+public class SpringFeatureFlagService implements FeatureFlagService, EnvironmentAware {
+
+  private Environment environment;
+
+  @Override
+  public boolean isEnabled(@Nonnull String flagName, boolean defaultValue) {
+    if (environment == null) {
+      return defaultValue;
+    }
+    return Boolean.valueOf(environment.getProperty(flagName, String.valueOf(defaultValue)));
```
cfieber: `environment.getProperty(flagName, Boolean.class, defaultValue)` for the generic'd one

robzienert: Oh neat. Missed that interface.

kork-core/src/main/java/com/netflix/spinnaker/kork/feature/FeatureFlagService.java
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.feature;
+
+import javax.annotation.Nonnull;
+
+/**
+ * A simple interface for interacting with dynamic Spring properties in the scope of feature flags.
+ */
+public interface FeatureFlagService {
```
cfieber: I wonder if there's a good way to establish a property convention that enables incremental roll out of a configurable option within an instance of a service in a common way. (Not scoping the property in the config store by profiles etc but rather scoping whether the flag should be on or not given some set of constraints)

As a concrete example, if we had some new feature in clouddriver and wanted to incrementally enable it in various accounts having something like

`isEnabled(String flagName, List<String> scopes, boolean defaultValue)`

and on the calling side in say a caching agent could do 

```
boolean useNewFeature = 
  featureFlagService.isEnabled("my.new.feature", [region.name, account.name], false)
```

And maybe that means on the FeatureFlagService end we do a 
`isEnabled('my.new.feature.<region.name>') ?: isEnabled('my.new.feature.<account.name?) ?: isEnabled('my.new.feature', defaultValue)`

cfieber: I don't know if I mildly like this suggestion or really hate it but its one of those two things

robzienert: I like the direction. Here's my stab at it...

```
<T> T getDynamicConfig(Class<T> configType, String configName, T defaultValue, Supplier<Boolean> predicate);

boolean isEnabled(String flagName, boolean defaultValue, Supplier<Boolean> predicate) {
  return getDynamicConfig(Boolean.class, flagName, defaultValue, predicate);
}
```

Then we could do something sorta like...

```
boolean useNewFeature =
  dynamicConfigService.isEnabled("my.new.feature", true, new MyNewFeaturePredicate(region.name, account.name));
```

Where in the predicate supplier could be any sort of arbitrary logic to determine if the flag should be considered or not. If the predicate fails, the `defaultValue` would be used.

¯\\\_(ツ)\_/¯

robzienert: Oh, heh... yeah... my solution totally skipped over the whole incremental rollout bit. If we go the property convention path, I could see us wanting to roll out a feature to a group of applications as well. I'll update the PR with something we can play with.

robzienert: @cfieber Updated the interface for what, I think, roughly matches your ideas. Lemme know what you think.

jrsquared:
🥈

ajordens:
I dig it, could possibly see us adding some auth details (user or cert) and having an override down to that level of granularity.

PR #150, opened 2018-05-04
feat(redis): Generic Locking support

jeyrschabu:
- `LockManager` defines a basic locking interface. 
- `RefreshableLock` as an extension of `LockManager`. Defines lock heartbeat functionality.
- Simple usage:
``` 
val acquireLockResponse = lockManager.acquireLock(locksName, maxLockDuration), {
    // Callback to run when lock is acquired
    // Lock will be kept alive with periodic heartbeats for `maxLockDuration`.
})

log.info("Status", acquireLockResponse.lockStatus)
- Heartbeats are handled by an internal single-threaded scheduler executor. The run frequency of the scheduler is configurable as well as the default lease duration.

### DIFF
--- a/kork-core/src/main/java/com/netflix/spinnaker/kork/lock/LockManager.java
+++ b/kork-core/src/main/java/com/netflix/spinnaker/kork/lock/LockManager.java
@@ -1,0 +1,329 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.kork.lock;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.apache.http.util.Asserts;
+
+import javax.annotation.Nonnull;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.time.Duration;
+import java.util.*;
+import java.util.concurrent.Callable;
+
+public interface LockManager {
+  <R> AcquireLockResponse<R> acquireLock(@Nonnull final LockOptions lockOptions,
+                                         @Nonnull final Callable<R> onLockAcquiredCallback);
+
+  AcquireLockResponse<Void> acquireLock(@Nonnull final LockOptions lockOptions,
+                                        @Nonnull final Runnable onLockAcquiredCallback);
+
+  <R> AcquireLockResponse<R> acquireLock(@Nonnull final String lockName,
+                                         final long maximumLockDurationMillis,
+                                         @Nonnull final Callable<R> onLockAcquiredCallback);
+
+  AcquireLockResponse<Void> acquireLock(@Nonnull final String lockName,
+                                        final long maximumLockDurationMillis,
+                                        @Nonnull final Runnable onLockAcquiredCallback);
+
+  boolean releaseLock(@Nonnull final Lock lock);
+
+  String NAME_FALLBACK = UUID.randomUUID().toString();
+
+  /**
+   * Used only if an ownerName is not provided in the constructor.
+   */
+  default String getOwnerName() {
+    try {
+      return InetAddress.getLocalHost().getHostName();
+    } catch (UnknownHostException e) {
+      return NAME_FALLBACK;
+    }
+  }
+
+  default String lockKey(String name) {
+    return String.format("{korkLock:%s}", name.toLowerCase());
+  }
+
+  class AcquireLockResponse<R> {
+    private final Lock lock;
+    private final R onLockAcquiredCallbackResult;
+    private final LockStatus lockStatus;
+    private final Exception exception;
+    private boolean released;
+
+    public AcquireLockResponse(final Lock lock,
+                               final R onLockAcquiredCallbackResult,
+                               final LockStatus lockStatus,
+                               final Exception exception,
+                               final boolean released) {
+      this.lock = lock;
+      this.onLockAcquiredCallbackResult = onLockAcquiredCallbackResult;
+      this.lockStatus = lockStatus;
+      this.exception = exception;
+      this.released = released;
+    }
+
+    public Lock getLock() {
+      return lock;
+    }
+
+    public R getOnLockAcquiredCallbackResult() {
+      return onLockAcquiredCallbackResult;
+    }
+
+    public LockStatus getLockStatus() {
+      return lockStatus;
+    }
+
+    public Exception getException() {
+      return exception;
+    }
+
+    public boolean isReleased() {
+      return released;
+    }
+  }
+
+  enum LockStatus {
+    ACQUIRED,
+    TAKEN,
+    ERROR,
+    EXPIRED
+  }
+
+  interface LockReleaseStatus {
+    String SUCCESS = "SUCCESS";
+    String SUCCESS_GONE = "SUCCESS_GONE"; //lock no longer exists
+    String FAILED_NOT_OWNER = "FAILED_NOT_OWNER"; //found lock but belongs to someone else
+  }
+
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  class Lock implements Named {
+    private final String name;
+    private final String ownerName;
+    private final long leaseDurationMillis;
+    private final long version;
+    private final long ownerSystemTimestamp;
+    private final String attributes; //arbitrary string to store data along side the lock
+
+    @JsonCreator
+    public Lock(@JsonProperty("name") String name,
+                @JsonProperty("ownerName") String ownerName,
+                @JsonProperty("version") long version,
+                @JsonProperty("leaseDurationMillis") long leaseDurationMillis,
+                @JsonProperty("ownerSystemTimestamp") long ownerSystemTimestamp,
+                @JsonProperty("attributes") String attributes) {
+      this.name = name;
+      this.ownerName = ownerName;
+      this.leaseDurationMillis = leaseDurationMillis;
+      this.ownerSystemTimestamp = ownerSystemTimestamp;
+      this.version = version;
+      this.attributes = attributes;
+    }
+
+    @Override
+    public String getName() {
+      return name;
+    }
+
+    public String getOwnerName() {
+      return ownerName;
+    }
+
+    public long getLeaseDurationMillis() {
+      return leaseDurationMillis;
+    }
+
+    public long getVersion() {
+      return version;
+    }
+
+    public long nextVersion() {
+      return version + 1;
+    }
+
+    public long getOwnerSystemTimestamp() {
+      return ownerSystemTimestamp;
+    }
+
+    public String getAttributes() {
+      return attributes;
+    }
+
+    @Override
+    public String toString() {
+      return "Lock{" +
+        "name='" + name + '\'' +
+        ", ownerName='" + ownerName + '\'' +
+        ", leaseDurationMillis=" + leaseDurationMillis +
+        ", version='" + version + '\'' +
+        ", ownerSystemTimestamp=" + ownerSystemTimestamp +
+        ", attributes=" + attributes +
+        '}';
+    }
+
+    @Override
+    public boolean equals(Object o) {
+      // Two locks are identical if the lock name, owner and version match.
+      if (this == o) return true;
+      if (o == null || getClass() != o.getClass()) return false;
+      Lock lock = (Lock) o;
+      return Objects.equals(name, lock.name) &&
+        Objects.equals(ownerName, lock.ownerName) &&
+        Objects.equals(version, lock.version);
+    }
+
+    @Override
+    public int hashCode() {
+      // Two locks are equal if lockName, ownerName and version match.
+      return Objects.hash(name, ownerName, version);
+    }
+  }
+
+  interface Named {
+    String getName();
+  }
+
+  class LockOptions {
+    private String lockName;
+    private Duration maximumLockDuration;
+    private long version;
+    private List<String> attributes = new ArrayList<>(); // the list will be joined with a ';' delimiter for brevity
+
+    public LockOptions withLockName(String name) {
+      this.lockName = name;
+      return this;
+    }
+
+    public LockOptions withMaximumLockDuration(Duration duration) {
+      this.maximumLockDuration = duration;
+      return this;
+    }
+
+    public LockOptions withAttributes(List<String> attributes) {
+      this.attributes = attributes;
+      return this;
+    }
+
+    public LockOptions withVersion(long version) {
+      this.version = version;
+      return this;
+    }
+
+    public String getLockName() {
+      return lockName;
+    }
+
+    public Duration getMaximumLockDuration() {
+      return maximumLockDuration;
+    }
+
+    public List<String> getAttributes() {
+      return attributes;
+    }
+
+    public long getVersion() {
+      return version;
+    }
+
+    public void validateInputs() {
+      Asserts.check(this.lockName.matches("^[a-zA-Z0-9.-]+$"),
+        "Lock name must be alphanumeric, may contain dots");
+
+      Objects.requireNonNull(this.lockName, "Lock name must be provided");
+      Objects.requireNonNull(this.maximumLockDuration, "Lock max duration must be provided");
+    }
+
+    public void setVersion(long version) {
+      this.version = version;
+    }
+
+    @Override
+    public String toString() {
+      return "LockOptions{" +
+        "lockName='" + lockName + '\'' +
+        ", maximumLockDuration=" + maximumLockDuration +
+        ", version=" + version +
+        ", attributes=" + attributes +
+        '}';
+    }
+  }
+
+  class LockException extends RuntimeException {
+    public LockException() {
+      super();
+    }
+
+    public LockException(String message) {
+      super(message);
+    }
+
+    public LockException(String message, Throwable cause) {
+      super(message, cause);
+    }
+
+    public LockException(Throwable cause) {
+      super(cause);
+    }
+  }
+
+  class LockCallbackException extends LockException {
+    public LockCallbackException(String message) {
+      super(message);
+    }
+
+    public LockCallbackException(String message, Throwable cause) {
+      super(message, cause);
+    }
+
+    public LockCallbackException(Throwable cause) {
+      super(cause);
+    }
+  }
+
+  class LockNotAcquiredException extends LockException {
+    public LockNotAcquiredException(String message) {
+      super(message);
+    }
+
+    public LockNotAcquiredException(String message, Throwable cause) {
+      super(message, cause);
+    }
+
+    public LockNotAcquiredException(Throwable cause) {
+      super(cause);
+    }
+  }
+
+  class LockExpiredException extends LockException {
+    public LockExpiredException(String message) {
+      super(message);
+    }
+
+    public LockExpiredException(String message, Throwable cause) {
+      super(message, cause);
+    }
+
+    public LockExpiredException(Throwable cause) {
+      super(cause);
+    }
+  }
+}
--- a/kork-jedis/kork-jedis.gradle
+++ b/kork-jedis/kork-jedis.gradle
@@ -1,8 +1,10 @@
 dependencies {
-  compile spinnaker.dependency('bootActuator')
+  compile project(":kork-core")
   compile spinnaker.dependency('spectatorApi')
+  compile spinnaker.dependency('bootActuator')
   compile spinnaker.dependency('jedis')
 
   testCompile spinnaker.dependency('groovy')
+  testCompile project(":kork-jedis-test")
   spinnaker.group('spockBase')
 }
--- a/kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
+++ b/kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
@@ -1,0 +1,482 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.kork.jedis.lock;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.spectator.api.Id;
+import com.netflix.spectator.api.LongTaskTimer;
+import com.netflix.spectator.api.Registry;
+import com.netflix.spinnaker.kork.jedis.RedisClientDelegate;
+import com.netflix.spinnaker.kork.lock.RefreshableLockManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nonnull;
+import javax.annotation.PreDestroy;
+import java.io.IOException;
+import java.time.Clock;
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Optional;
+import java.util.List;
+import java.util.concurrent.*;
+
+import static com.netflix.spinnaker.kork.jedis.lock.RedisLockManager.LockScripts.*;
+import static com.netflix.spinnaker.kork.lock.LockManager.LockReleaseStatus.*;
+
+public class RedisLockManager implements RefreshableLockManager {
+  private static final Logger log = LoggerFactory.getLogger(RedisLockManager.class);
+  private static final long DEFAULT_HEARTBEAT_RATE_MILLIS = 5000L;
+  private static final long DEFAULT_TTL_MILLIS = 10000L;
+
+  private final String ownerName;
+  private final Clock clock;
+  private final Registry registry;
+  private final ObjectMapper objectMapper;
+  private final RedisClientDelegate redisClientDelegate;
+  private final ScheduledExecutorService scheduledExecutorService;
+
+  private final Id acquireId;
+  private final Id releaseId;
+  private final Id heartbeatId;
+
+  private long heartbeatRateMillis;
+  private long leaseDurationMillis;
+  private BlockingDeque<HeartbeatLockRequest> heartbeatQueue;
+
+  public RedisLockManager(String ownerName,
+                          Clock clock,
+                          Registry registry,
+                          ObjectMapper objectMapper,
+                          RedisClientDelegate redisClientDelegate,
+                          Optional<Long> heartbeatRateMillis,
+                          Optional<Long> leaseDurationMillis
+  ) {
+    this.ownerName = Optional.ofNullable(ownerName).orElse(getOwnerName());
+    this.clock = clock;
+    this.registry = registry;
+    this.objectMapper = objectMapper;
+    this.redisClientDelegate = redisClientDelegate;
+    this.scheduledExecutorService = Executors.newSingleThreadScheduledExecutor();
+    this.heartbeatQueue = new LinkedBlockingDeque<>();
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

**Comment by jeyrschabu** [comment]
-- WORK IN PROGRESS --
- Supports simple use case of acquiring a simple lock:
```acquireLock(lockName, ttlMillis, {
        //Do Work
})```

- Supports configuring lock options: heartbeat config, timeout...
- Added basic support for heartbeat refresh (Though not happy with current impl)

TODO:
- Revisit heartbeat monitoring and lease renew
- Support interruptions when lock vanishes or expired

File: kork-core/src/main/java/com/netflix/spinnaker/kork/lock/LockManager.java [hunk 1]
```diff
@@ -0,0 +1,283 @@
...
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.Callable;
+
+public interface LockManager {
+  <R> AcquireLockResponse<R> acquireLock(final LockOptions lockOptions, final Callable<R> onLockAcquired);
+  <R> AcquireLockResponse<R> acquireLock(final String lockName, final long leaseDurationMillis, final Callable<R> onLockAcquired);
+  <R> AcquireLockResponse<R> acquireLock(final String lockName, final long leaseDurationMillis, int maxHeartbeatCount, final Callable<R> onLockAcquired);
```
robzienert: Let's get some newlines on this guy so we're not leaking over 120 char.

kork-core/src/main/java/com/netflix/spinnaker/kork/lock/LockManager.java
```diff
@@ -0,0 +1,283 @@
...
+  boolean releaseLock(final String name, final String previousVersion);
+
+
+  default String lockKey(String name) {
```
robzienert: Too many new lines, tho.

kork-core/src/main/java/com/netflix/spinnaker/kork/lock/LockManager.java
```diff
@@ -0,0 +1,283 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.http.util.Asserts;
+import java.io.IOException;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.Callable;
+
+public interface LockManager {
```
robzienert: I think we should avoid having default methods for Redis-specific implementation. Ideally with this interface, we'll be able to switch out Redis as our consensus backend for ZK in the future, or OSS could drop in etcd / consul, etc.

kork-core/src/main/java/com/netflix/spinnaker/kork/lock/LockManager.java
```diff
@@ -0,0 +1,283 @@
...
+      this.ownerName = ownerName;
+      this.leaseDurationMillis = leaseDurationMillis;
+      this.ownerSystemTimestamp = ownerSystemTimestamp;
+      this.version = version;
+      this.attributes = attributes;
+    }
+
+    @JsonCreator
+    public static Lock create(String json) throws IOException {
+      return new ObjectMapper().readValue(json, Lock.class);
```
robzienert: It's pretty inefficient to create a new object mapper each time we want to create a lock.

kork-core/src/main/java/com/netflix/spinnaker/kork/lock/LockManager.java
```diff
@@ -0,0 +1,283 @@
...
+    public LockException(String message, Throwable cause) {
+      super(message, cause);
+    }
+
+    public LockException(Throwable cause) {
+      super(cause);
+    }
+  }
+
+  class LockCallbackException extends RuntimeException {
```
robzienert: Should this extend `LockException`?

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,425 @@
...
+  private final Clock clock;
+  private final Registry registry;
+  private final ObjectMapper objectMapper;
+  private final RedisClientDelegate redisClientDelegate;
+  private final ScheduledExecutorService executorService;
+  private final Id acquireId;
+  private final Id releaseId;
+  private final Id heartbeatId;
+
+  private BlockingDeque<RefreshLockRequest> heartbeatQueue = new LinkedBlockingDeque<>();
```
robzienert: Need a newline here.

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,425 @@
...
+      LockStatus status = LockStatus.TAKEN;
+      boolean released = false;
+      try {
+        if (lockOptions.getVersion() == null) {
+          lockOptions.setVersion(String.format("%s:%s", ownerName, clock.millis()));
+        }
+
+        Object payload = Optional.ofNullable(getAcquireLockScriptPayload(lockOptions, client))
+          .orElseThrow(
+            () -> new LockAcquireFailedException(
```
robzienert: Maybe `AcquireLockFailedException` instead, so it reads a little more naturally?

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,425 @@
...
+          } catch (Exception e) {
+            LOGGER.error("Function call failed", e);
+            throw new LockCallbackException(e);
+          }
+        }
+
+        acquireLockResponse = new AcquireLockResponse<>(lock, workResult, status, null, released);
+      } catch (Exception e) {
+        if (e instanceof LockCallbackException) {
+          throw new LockCallbackException(e);
```
robzienert: Should just `throw e;` instead of wrapping the exception in the same exception class.

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,425 @@
...
+          status = LockStatus.ACQUIRED;
+          LOGGER.info("Acquired Lock {}. Processing work started...", lock);
+
+          RefreshLockRequest refreshLockRequest = new RefreshLockRequest(lock, lockOptions.getMaxHeartbeatCount());
+          if (!heartbeatQueue.contains(refreshLockRequest)) {
+            if (lockOptions.getMaxHeartbeatCount() > 0) {
+              LOGGER.info("Lock {} will get {} heartbeats", lock, lockOptions.getMaxHeartbeatCount());
+              heartbeatQueue.add(refreshLockRequest);
+            }
+          }
```
robzienert: Neat. Does this heartbeat queue then mean we the LockManager must be a singleton, though? I've no particular problems with that, but it should be documented at the interface-level if that's the behavior we expect.

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,425 @@
...
+    LockOptions lockOptions = new LockOptions()
+      .withLockName(lockName)
+      .withOwnerSystemTimestamp(clock.millis())
+      .withLeaseDuration(Duration.ofMillis(leaseDurationMillis));
+
+    return acquireLock(lockOptions, onLockAcquired);
+  }
+
+  @Override
+  public <R> AcquireLockResponse<R> acquireLock(String lockName, long leaseDurationMillis, int maxHeartbeatCount, Callable<R> onLockAcquired) {
```
robzienert: More 120 character formatting stuff.

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,425 @@
...
+    this.executorService = executorService;
+    this.ownerName = ownerName;
+    this.clock = clock;
+    this.registry = registry;
+    this.objectMapper = objectMapper;
+    this.redisClientDelegate = redisClientDelegate;
+
+    acquireId = registry.createId("redis.lock.acquire");
+    releaseId = registry.createId("redis.lock.release");
+    heartbeatId = registry.createId("redis.lock.heartbeat");
```
robzienert: I think it would be valuable to have a timer metric for how long a lock is actually held for.

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,425 @@
...
+      .withOwnerSystemTimestamp(clock.millis())
+      .withMaxHeartbeatCount(maxHeartbeatCount)
+      .withLeaseDuration(Duration.ofMillis(leaseDurationMillis));
+
+    return acquireLock(lockOptions, onLockAcquired);
+  }
+
+  @Override
+  public boolean releaseLock(final String name, final String previousVersion) {
+    String normalizedName = name.toLowerCase();
```
robzienert: It just occurred to me that we should probably do a little more on name normalization. Since the name is used as a metric tag, we need to make sure it doesn't have illegal characters for Atlas. Perhaps we just just keep `[a-zA-Z0-9]`?

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,425 @@
...
+      }
+    });
+  }
+
+  private void scheduleHeartbeats() {
+    executorService.scheduleAtFixedRate(
+      () -> {
+        while (!heartbeatQueue.isEmpty()) {
+          RefreshLockRequest refreshLockRequest = heartbeatQueue.getFirst();
+          if (refreshLockRequest.remainingHeartbeatCount >= 0) {
```
robzienert: "If the remaining heartbeat count is 0, do a heartbeat"? Maybe better just `gt` comparison instead?

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,425 @@
...
+            heartbeat(refreshLockRequest.lock);
+            refreshLockRequest.decrementRemainingHeartbeat();
+          }
+
+        }
+      }, 0, HEARTBEAT_RATE_MILLIS, MILLISECONDS);
```
robzienert: We should probably have a pre-destroy action to cleanup the executor service / release any locks that this instance of the LockManager is currently managing.

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,425 @@
...
+      )
+    );
+  }
+
+  private boolean releaseLockQuietly(Lock lock) {
+    if (lock != null) {
+      try {
+        return releaseLock(lock.getName(), lock.getVersion());
+      } catch (Exception e) {
+        LOGGER.warn("Attempt to release lock {} failed", lock);
```
robzienert: Should also include the exception as part of the log message.

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,425 @@
...
+        LOGGER.error("Error running heartbeat script {}", lock.getName(), e);
+        registry.counter(lockHeartbeat.withTag("status", "ERROR")).increment();
+        throw new LockHeartbeatFailedException("Failed to send heartbeat for lock" + lock.getVersion());
+      }
+    });
+  }
+
+  private IllegalStateException lockGoneException(Id lockHeartbeat) {
+    registry.counter(lockHeartbeat.withTag("status", "LOCK_GONE")).increment();
+    return new IllegalStateException("Lock not found.");
```
robzienert: We need to include more context in this exception message. Like, what is the lock name that is no longer there?

jeyrschabu:
@robzienert PT-Another-L.

kork-core/src/main/java/com/netflix/spinnaker/kork/lock/LockManager.java
```diff
@@ -0,0 +1,325 @@
...
+import java.util.UUID;
+import java.util.concurrent.Callable;
+
+public interface LockManager {
+  <R> AcquireLockResponse<R> acquireLock(@Nonnull final LockOptions lockOptions,
+                                         @Nonnull final Callable<R> onLockAcquiredCallback);
+
+  <R> AcquireLockResponse<R> acquireLock(@Nonnull final String lockName,
+                                         final long leaseDurationMillis,
+                                         @Nonnull final Callable<R> onLockAcquiredCallback);
```
robzienert: Would like interfaces for `AcquireLockResponse<Void> acquireLock(..., @Nonnull final Runnable onLockAcquiredCallback)`.

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,442 @@
...
+      try {
+        lock = tryCreateLock(lockOptions);
+      } catch (IOException e) {
+        throw new IllegalStateException("Failed to deserialize lock content", e);
+      }
+
+      if (!matchesLock(lockOptions, lock)) {
+        log.info("Could not acquire already taken lock {}", lock);
+      } else {
+        LongTaskTimer acquireDurationTimer = registry.longTaskTimer(String.format("redis.lock.acquireDuration.%s", lock.getName()));
```
robzienert: Line too long.

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,442 @@
...
+        try {
+          HeartbeatResponse heartbeatResponse = heartbeat(heartbeatLockRequest.getLock());
+          switch (heartbeatResponse.getLockStatus()) {
+            case EXPIRED:
+            case ERROR:
+              log.warn("Lock status {} for {}", heartbeatResponse.getLockStatus(), heartbeatResponse.getLock());
+              heartbeatQueue.remove(heartbeatLockRequest);
+              break;
+            default:
+              log.info("Remaining heartbeat count: {}. Refreshed lock {}", remainingHeartbeat, heartbeatResponse.getLock());
```
robzienert: More long lines; keep lt 120.

kork-jedis/src/test/groovy/com/netflix/spinnaker/kork/jedis/locking/RedisLockManagerSpec.groovy
```diff
@@ -0,0 +1,141 @@
...
+import static com.netflix.spinnaker.kork.lock.RefreshableLockManager.*
+
+class RedisLockManagerSpec extends Specification {
+  def embeddedRedis = EmbeddedRedis.embed()
+  def jedisPool = embeddedRedis.getPool() as JedisPool
+  def objectMapper = new ObjectMapper()
+  def clock = Clock.systemDefaultZone()
+  def registry = new NoopRegistry()
+  def redisClientDelegate = new JedisClientDelegate(jedisPool)
+  def lockLeaseDurationMillis = 6000
```
robzienert: Let's make this lease duration shorter in the tests. Is there any difference between 20ms lock time and 6000ms?

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,488 @@
...
+    HeartbeatLockRequest heartbeatLockRequest = null;
+    lockOptions.setVersion(clock.millis());
+    try {
+      try {
+        lock = tryCreateLock(lockOptions);
+      } catch (IOException e) {
+        throw new IllegalStateException("Failed to deserialize lock content", e);
+      }
+
+      if (!matchesLock(lockOptions, lock)) {
```
dreynaud: I would prefer this:

```
if (!matchesLock(lockOptions, lock)) {
    log.warn("Could not acquire already taken lock {}", lock);
    return new AcquireLockResponse<>(lock, workResult, status, null, tryLockReleaseQuietly(lock));
}

// keep going with the "matches lock" case, no need for an else block
LongTaskTimer acquireDurationTimer = registry.longTaskTimer(
...
```

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,488 @@
...
+    scheduledExecutorService
+      .scheduleAtFixedRate(this::sendHeartbeats, 0, heartbeatRateMillis, TimeUnit.MILLISECONDS);
+  }
+
+  /**
+   * Send heartbeats to queued locks. Monitors maximum heartbeat count when provided.
+   * If a max heartbeat is provided, the underlying lock will receive at most the provided maximum heartbeats.
+   */
+  private void sendHeartbeats() {
+    if (!heartbeatQueue.isEmpty()) {
```
dreynaud: I would prefer:

```
if (heartbeatQueue.isEmpty()) {
    return;
}

// go on with real work here, no else block
HeartbeatLockRequest heartbeatLockRequest = heartbeatQueue.getFirst();
...
```

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,488 @@
...
+              heartbeatQueue.remove(heartbeatLockRequest);
+              break;
+            default:
+              log.info("Remaining heartbeat count: {}. Refreshed lock {}",
+                remainingHeartbeat, heartbeatResponse.getLock());
+              heartbeatLockRequest.setLock(heartbeatResponse.getLock());
+          }
+        } catch (Exception e) {
+          log.error("Heartbeat {} for {} failed", heartbeatLockRequest, heartbeatLockRequest.getLock(), e);
+          heartbeatQueue.remove(heartbeatLockRequest);
```
dreynaud: does that mean we would stop sending heartbeats if there is even something like a temporary exception?

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,488 @@
...
+
+  /**
+   * Send heartbeats to queued locks. Monitors maximum heartbeat count when provided.
+   * If a max heartbeat is provided, the underlying lock will receive at most the provided maximum heartbeats.
+   */
+  private void sendHeartbeats() {
+    if (!heartbeatQueue.isEmpty()) {
+      HeartbeatLockRequest heartbeatLockRequest = heartbeatQueue.getFirst();
+      int remainingHeartbeat = Math.max(heartbeatLockRequest.getRemainingHeartbeatCount().get(), 0);
+      if (remainingHeartbeat == 0) {
```
dreynaud: simpler version I think:

```
int remainingHeartbeat = heartbeatLockRequest.getRemainingHeartbeatCount().get();
if (remainingHeartbeat <= 0) {
    ...
```

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,488 @@
...
+    }
+  }
+
+  /**
+   * A heartbeat will only be accepted if the provided version matches the
+   * version stored in Redis. If a heartbeat is accepted, a new version value will
+   * be stored with the lock along side a renewed lease and the system timestamp.
+   */
+  private HeartbeatResponse heartbeat(final Lock lock) {
+    Id lockHeartbeat = heartbeatId.withTag("lockName", lock.getName().toLowerCase());
```
dreynaud: why lower case though?

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,488 @@
...
+    });
+
+    if (payload == null) {
+      throw new LockNotAcquiredException(String.format("Lock not acquired %s", lockOptions));
+    }
+
+    return objectMapper.readValue(payload.toString(), Lock.class);
+  }
+
+  private String tryRemoveLock(final Lock lock) {
```
dreynaud: `tryReleaseLock` ?

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,488 @@
...
+     * completed (no longer the owner, different version), 2 if the lock no
+     * longer exists.
+     * <p>
+     * ARGS
+     * 1: owner
+     * 2: previousRecordVersion
+     * 3: newRecordVersion
+     */
+    String RELEASE_SCRIPT = "" +
+      "if redis.call('EXISTS', KEYS[1]) == 1 then" +
```
dreynaud: isn't it possible to fold the `EXISTS` and `GET` in one operation? As in:

```
local payload = redis.call('GET', KEYS[1]))
if payload != nil then
   ...
```

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,488 @@
...
+     * ARGS
+     * 1: ownerName
+     * 2: previousRecordVersion
+     * 3: newRecordVersion
+     * 4: newleaseDurationMillis
+     * 5: updatedOwnerSystemTimestamp
+     */
+    String HEARTBEAT_SCRIPT = "" +
+      "if redis.call('EXISTS', KEYS[1]) == 1 then" +
+      "  local payload = redis.call('GET', KEYS[1])" +
```
dreynaud: same comment about potentially folding `EXISTS` and `GET` together

kork-core/src/main/java/com/netflix/spinnaker/kork/lock/LockManager.java
```diff
@@ -0,0 +1,331 @@
...
+  default String getOwnerName() {
+    try {
+      return InetAddress.getLocalHost().getHostName();
+    } catch (UnknownHostException e) {
+      return NAME_FALLBACK;
+    }
+  }
+
+  default String lockKey(String name) {
+    return String.format("{korkLock:%s}", name.toLowerCase());
```
dreynaud: 👍

kork-jedis/src/test/groovy/com/netflix/spinnaker/kork/jedis/locking/RedisLockManagerSpec.groovy
```diff
@@ -0,0 +1,157 @@
...
+    when:
+    def result = redisLockManager.acquireLock(lockOptions, {
+      // simulates long running task
+      Thread.sleep(100)
+      "Done"
+    } as Callable<String>)
+
+    then:
+    result.released == true
+    result.lock.leaseDurationMillis > lockLeaseDurationMillis
```
dreynaud: not sure I understand this, why would the final `leaseDurationMillis` necessarily be bigger than the initial value?

kork-jedis/src/test/groovy/com/netflix/spinnaker/kork/jedis/locking/RedisLockManagerSpec.groovy
```diff
@@ -0,0 +1,157 @@
...
+      return "Run after lock is safely acquired."
+    } as Callable<String>)
+
+    then:
+    result.lockStatus == ACQUIRED
+    result.released == true
+    result.onLockAcquiredCallbackResult == "Run after lock is safely acquired."
+  }
+
+  def "should fail to acquire an already taken lock"() {
```
dreynaud: would be neat to also test lock expiration: acquire, wait (e.g. caller is dead), next acquire succeeds
also would be extra neat to test what happens when a long task is submitted and we run out of heartbeats

kork-jedis/src/test/groovy/com/netflix/spinnaker/kork/jedis/locking/RedisLockManagerSpec.groovy
```diff
@@ -0,0 +1,157 @@
...
+import spock.lang.Specification
+
+import java.time.Clock
+import java.time.Duration
+import java.util.concurrent.Callable
+
+import static com.netflix.spinnaker.kork.lock.LockManager.LockStatus.*
+import static com.netflix.spinnaker.kork.lock.RefreshableLockManager.*
+
+class RedisLockManagerSpec extends Specification {
```
dreynaud: very cool test! it pulls everything together very nicely 👏

kork-jedis/kork-jedis.gradle
```diff
@@ -1,8 +1,8 @@
 dependencies {
-  compile spinnaker.dependency('bootActuator')
-  compile spinnaker.dependency('spectatorApi')
```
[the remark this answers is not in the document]
jeyrschabu: @dreynaud `kork-core` already includes those. bringing in core and removing dupes dependencies

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,488 @@
...
+  @Override
+  public AcquireLockResponse<Void> acquireLock(@Nonnull LockOptions lockOptions,
+                                               @Nonnull Runnable onLockAcquiredCallback) {
+    return acquire(lockOptions, onLockAcquiredCallback);
+  }
+
+  @Override
+  public boolean releaseLock(@Nonnull final Lock lock) {
+    try {
+      Integer result = Integer.valueOf(tryRemoveLock(lock));
```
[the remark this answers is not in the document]
jeyrschabu: Fine Daniel!, just for you though. 
That 's a reasonable ask though. I will make it so.

kork-jedis/src/test/groovy/com/netflix/spinnaker/kork/jedis/locking/RedisLockManagerSpec.groovy
jeyrschabu: The heartbeat increases the lease duration.

kork-jedis/src/test/groovy/com/netflix/spinnaker/kork/jedis/locking/RedisLockManagerSpec.groovy
jeyrschabu: The second part is covered in the test already. Currently display a warning saying the lock will no longer get heartbeats, and that the lock would be up for grabs when expired.

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
```diff
@@ -0,0 +1,488 @@
...
+          heartbeatLockRequest.getLock(), heartbeatLockRequest.getRemainingHeartbeatCount());
+        heartbeatQueue.add(heartbeatLockRequest);
+      }
+    }
+  }
+
+  private <R> AcquireLockResponse<R> doAcquire(@Nonnull final LockOptions lockOptions,
+                                               final Optional<Callable<R>> onLockAcquiredCallbackCallable,
+                                               final Optional<Runnable> onLockAcquiredCallbackRunnable) {
+    int maxHeartbeatCount = Integer.MAX_VALUE;
```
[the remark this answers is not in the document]
jeyrschabu: By default a lock will get heartbeats forever, thus the int max.

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
jeyrschabu: That's correct. Though, as a second thought, we could just give it another chance until remaining heartbeats are up. However this could be problematic if `maxHeartbeatCount` is not provided. It would try for a veryyy long time.

kork-jedis/kork-jedis.gradle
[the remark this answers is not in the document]
dreynaud: I know, but it's better to have explicit dependencies. If one day `kork-core` drops `spectatorApi`, then `kork-jedis` will fail. Explicitly declaring your direct dependencies protects you against this.

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/lock/RedisLockManager.java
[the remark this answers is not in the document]
dreynaud: Isn't that introducing a risk though? If the callback gets somehow blocked, then there will be no recovery. I'd be more comfortable with a tighter default (minutes at most), especially since you can override it for longer-running tasks

PR #151, opened 2018-05-07
config(okhttp3): add okhttp3 client configuration

dreynaud:


dreynaud:
I want to start experimenting with this in orca so I'll merge early, but please comment away

PR #152, opened 2018-05-08
adds syncPipeline to RedisClientDelegates

cfieber:
also incidentally:

* updates to latest spinnaker-gradle-plugin and gradle 4.7
* fixes a flaky cassandra test that was failing due to a port conflict


kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/RedisClientDelegate.java
```diff
@@ -45,6 +45,8 @@
 
   <R> R withPipeline(Function<RedisPipeline, R> f);
 
+  void syncPipeline(RedisPipeline p);
```
cfieber: because `RedisPipeline` doesn't expose `sync` directly, users of withPipeline typically have to cast to a concrete implementation in order to sync at the end, which isn't great if something that doesn't know the implementation wants to do a pipelined operation

sort of deliberatly only added sync and not syncAndReturnAll on this

robzienert: Do like. Yeah, it's annoying that this wasn't included in the interface.

PR #158, opened 2018-05-30
feat(core): Optional application profile config property sources

robzienert:
Not terribly different than what we're doing for Metatron.


emjburns:
Awesome!

PR #159, opened 2018-06-06
fix(locking): Use Arrays.asList(..) over Collections.singletonList(..)

jeyrschabu:
- Collections.singletonList(..) causes a ClassCastException when using the dyno redis driver
- Support retrying heartbeat on failure

### DIFF
--- a/kork-core/src/main/java/com/netflix/spinnaker/kork/lock/RefreshableLockManager.java
+++ b/kork-core/src/main/java/com/netflix/spinnaker/kork/lock/RefreshableLockManager.java
@@ -20,6 +20,7 @@ import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Objects;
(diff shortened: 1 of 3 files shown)


kork-core/src/main/java/com/netflix/spinnaker/kork/lock/RefreshableLockManager.java
```diff
@@ -66,20 +72,36 @@ public Duration getRemainingLockDuration() {
       return Duration.between(startedAt, clock.instant()).minus(heartbeatDuration).abs();
     }
 
+    public boolean shouldRetry() {
+      return retriesOnFailure.decrementAndGet() > 0;
```
dreynaud: shouldn't that be `return retriesOnFailure.decrementAndGet() >= 0;`? 
It returns the updated value, so retriesOnFailure = 1 would lead to no retries

PR #163, opened 2018-06-12
refactor(locking): Renaming metrics names to be more generic

jeyrschabu:
- added a generic lock names

### DIFF
--- a/kork-core/src/main/java/com/netflix/spinnaker/kork/lock/LockManager.java
+++ b/kork-core/src/main/java/com/netflix/spinnaker/kork/lock/LockManager.java
@@ -202,6 +202,13 @@ public interface LockManager {
     String getName();
   }
 
+  interface LockMetricsConstants {
+    String ACQUIRE = "kork.lock.acquire";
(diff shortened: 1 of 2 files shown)


kork-core/src/main/java/com/netflix/spinnaker/kork/lock/LockMetricsConstants.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.kork.lock;
+
+public interface LockMetricsConstants {
```
robzienert: Do we like it as its own interface, or should they just go on `LockManager`? Forgot that strings could be defined on interfaces.

PR #166, opened 2018-06-18
feat(web): Add request content length to metrics

robzienert:
Some prep work for Q3 pubsub work. I want to know what sort of payload sizes we're shipping around between services, as different sizes will dictate different approaches. Hard use case: If a request payload `orca -> echo` is greater than 256KB, we won't be able to ship the actual message via SQS and would instead need to store it in S3, or maybe use a log stream instead, or solution Z.

### DIFF
--- a/kork-web/src/main/java/com/netflix/spinnaker/kork/web/interceptors/MetricsInterceptor.java
+++ b/kork-web/src/main/java/com/netflix/spinnaker/kork/web/interceptors/MetricsInterceptor.java
@@ -30,6 +30,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
+import static java.lang.String.format;
+
 /**
  * An interceptor that logs Controller metrics to an underlying {@link com.netflix.spectator.api.Registry}.
  * <p>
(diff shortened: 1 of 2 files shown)


kork-web/src/main/java/com/netflix/spinnaker/kork/web/interceptors/MetricsInterceptor.java
```diff
@@ -100,7 +100,8 @@ public void afterCompletion(HttpServletRequest request,
         .withTag("controller", controller)
         .withTag("method", handlerMethod.getMethod().getName())
         .withTag("status", status.toString().charAt(0) + "xx")
-        .withTag("statusCode", status.toString());
+        .withTag("statusCode", status.toString())
+        .withTag("reqContentLength", Long.toString(request.getContentLengthLong()));
```
cfieber: I think we need a separate metric (distribution summary probably?) rather than adding a tag to the existing timer. Adding the tag will cause the number of metrics to explode per discrete content length

robzienert: omg I'm dumb

PR #171, opened 2018-06-25
feat(okhttp): A simple metrics network interceptor for okhttp/okhttp3

ajordens:
Reports a timer with sufficient tag detail to identify communication
issues between services.

ie) requests from `orca` to `echo` are failing

Only the host is identified, see the `controller.invocations` tag on the
destination service if you'd like more information about specific
endpoints that may be problematic.


ajordens:
Somewhat close approximation of what the `EurekaOkClient` is doing in `gate`.

Duplicate code between `okhttp` and `okhttp3` variants, but I suspect the former will disappear at some point.

This will be pulled into `orca` initially.

PR #176, opened 2018-07-09
fix(hystrix): Avoid an `NaN` when publishing `hystrix.currentTime`

ajordens:
Same as spinnaker/gate#558


ewiseblatt:
LGTM. Thanks.

PR #179, opened 2018-07-13
 feat(artifacts): Artifact matching should use entire string

ezimanyi:
BREAKING CHANGE: This change affects the way in which artifacts injected
into a pipeline are matched to expected artifacts. Prior to this change,
a match was successful if the regular expression in the expected artifact
matched any substring of the corresponding field in the incoming artifact;
after this change, the regular expression in the expected artifact must
match the entire corresponding field incoming artifact.

As an example, consider an incoming artifact with name 'artifact123'. Before
this change, it would match expected artifacts with names 'artifact' and
'artifact.*'.  After this change, the incoming artifact will no longer
match an expected artifact with name 'artifact' but will continue to match
'artifact*'; users will need to explicitly add a wildcard (or some more
complex regular expression) if substring matching is desired.

### DIFF
--- a/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/ExpectedArtifact.java
+++ b/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/ExpectedArtifact.java
@@ -46,31 +46,31 @@ public class ExpectedArtifact {
   public boolean matches(Artifact other) {
     String thisType = matchArtifact.getType();
     String otherType = other.getType();
-    if (StringUtils.isNotEmpty(thisType) && (otherType == null || !matches(thisType, otherType))) {
+    if (!matches(thisType, otherType)) {
       return false;
     }
 
     String thisName = matchArtifact.getName();
     String otherName = other.getName();
-    if (StringUtils.isNotEmpty(thisName) && (otherName == null || !matches(thisName, otherName))) {
+    if (!matches(thisName, otherName)) {
       return false;
     }
 
     String thisVersion = matchArtifact.getVersion();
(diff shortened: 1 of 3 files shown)


kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/ExpectedArtifact.java
```diff
@@ -80,6 +80,10 @@ public boolean matches(Artifact other) {
   }
 
   private boolean matches(String us, String other) {
-    return Pattern.compile(us).asPredicate().test(other);
+    return StringUtils.isEmpty(us) || (other != null && patternMatches(us, other));
```
lwander: Nice

kork-artifacts/src/test/groovy/com/netflix/spinnaker/kork/artifacts/model/ExpectedArtifactSpec.groovy
```diff
@@ -0,0 +1,99 @@
...
+class ExepectedArtifactSpec extends Specification {
+  private static final EXPECTED_STRING = 'abc.*'
+  private static final MATCH_STRING = 'abcd'
+  private static final NO_MATCH_STRING = 'zzz'
+
+  @Shared typeFactory = { String input ->
+    Artifact.builder().type(input).build()
+  }
+
+  @Shared nameFactory = {String input ->
```
lwander: All of these (except the first) are missing a space after the `{`

ezimanyi:
https://github.com/spinnaker/spinnaker/issues/3011

PR #182, opened 2018-08-15
feat(redis): Support backwards-compat connection configs

robzienert:


kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/RedisClientConfiguration.java
```diff
@@ -57,27 +57,48 @@
...
+    }
     return getClientFactoryForDriver(driver).build(name, properties);
   }
 
+  private static void inheritLegacyRedisConfig(Map<String, Object> properties,
+                                               ClientConfigurationWrapper rootConfig,
+                                               String configKey) {
+    Optional.ofNullable(rootConfig.connection).map(v -> {
+      if (!Optional.ofNullable(properties.get(configKey)).isPresent()) {
+        properties.put(configKey, v);
```
emjburns: This puts 
```connection: redis://localhost:6379
timeoutMs: redis://localhost:6379
``` into `properties`

Comment (emjburns) [reply kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/RedisClientConfiguration.java]: probably by pulling this out of a method and doing the below on ln 85?

```
Optional.ofNullable(rootConfig.connection).map(v -> {
      if (!Optional.ofNullable(properties.get("connection")).isPresent()) {
        properties.put("connection", v);
      }
      return v;
    });
Optional.ofNullable(rootConfig.timeoutms).map(v -> {
      if (!Optional.ofNullable(properties.get("timeoutMs")).isPresent()) {
        properties.put("timeoutMs", v);
      }
      return v;
    });
```

File: kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/RedisClientConfiguration.java [hunk 2]
```diff
@@ -57,27 +57,48 @@
         clients.add(createClient(
```
emjburns: Above this for each need to add 

```
    if (redisClientConfigurations.clients == null || redisClientConfigurations.clients.isEmpty()) {
      clients.add(createClient("primaryDefault", Driver.REDIS, new HashMap<>(), redisClientConfigurations));
    }
```
so that we create a "primaryDefault" client when only `redis.connection` appears in the config

kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/RedisClientConfiguration.java
robzienert: lol. Whoops.

PR #187, opened 2018-09-06
fix(web): Avoid extracting `spinnakerUser` from cert if available in header

ajordens:
Handles a situation where a certificate used for service to service
communication was created with an RFC822 email address.

### DIFF
--- a/kork-web/src/main/groovy/com/netflix/spinnaker/filters/AuthenticatedRequestFilter.groovy
+++ b/kork-web/src/main/groovy/com/netflix/spinnaker/filters/AuthenticatedRequestFilter.groovy
@@ -74,27 +74,15 @@ class AuthenticatedRequestFilter implements Filter {
     def spinnakerExecutionId = null
 
     try {
-      if (request.isSecure()) {
-        ((X509Certificate[]) request.getAttribute(X509_CERTIFICATE))?.each {
-          def emailSubjectName = it.getSubjectAlternativeNames().find {
-            it.find { it.toString() == RFC822_NAME_ID }
-          }?.get(1)
-
-          spinnakerUser = spinnakerUser ?: emailSubjectName
-        }
+      def session = ((HttpServletRequest) request).getSession(false)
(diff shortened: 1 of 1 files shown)


kork-web/src/main/groovy/com/netflix/spinnaker/filters/AuthenticatedRequestFilter.groovy
```diff
@@ -115,6 +103,17 @@ class AuthenticatedRequestFilter implements Filter {
       spinnakerRequestId = UUID.randomUUID().toString()
     }
 
+    // only extract from the x509 certificate if `spinnakerUser` has not been supplied as a header
+    if (request.isSecure() && !spinnakerUser) {
```
cfieber: would anything prevent somebody hitting gate with a valid cert but supplying `X-SPINNAKER-USER` to masquerade as someone else in this case?

ajordens: `gate` explicitly does not extract spinnaker headers.

https://github.com/spinnaker/gate/blob/master/gate-web/src/main/groovy/com/netflix/spinnaker/gate/config/GateConfig.groovy#L320

iiro:
Sorry for barging in on here - this discussion just reminded me of an issue from yesterday... I just bumped into an interesting behaviour when using email as a CommonName (when used together with SAML at least): if I did a request with roer using this certificate - and tried to edit any applications' permissions - it would list this CommonName-email address in a permission-dropdown - and remove the other SAML groups which my user does not belong to from the dropdown. Sounds like a bug...? I had to change the CommonName into something else for now - then things are fine. Sounds like ticket? I'll create one of course if...

ajordens:
@iiro This PR shouldn't impact any of the described behavior (which may or may not be a bug --- I'm not too familiar with what roer is doing) ... it's a change in priority in the places we looked for an email address within a certificate or request. 

It specifically affects environments where you've got mTLS enabled between Spinnaker services and enforces that the email address of the authenticated user will be used rather than any email address encoded within the client certificate.

PR #188, opened 2018-09-11
fix(jedis): Allow services to override connection pool settings

robzienert:
While working on a fix for https://github.com/spinnaker/spinnaker/issues/3309, I noticed a few bugs / missing items.

* Adds a `GenericObjectPoolConfig` as a fallback bean to the `JedisClientConfiguration`. This will now allow kork to use any service-provided override but fallback to default settings if not provided. `GenericObjectPoolConfig` can still be provided on a per-client basis.
* Fixed a bug where two `primaryDefault` clients could be created. For example, Orca currently creates its own `primaryDefault` client, then kork was creating its own as well. Kork will now only create these default clients if one has not already been created by a service.
* Backwards compatibility for `redis.connectionPrevious` was not present. I added this following the same semantics as the `primaryDefault` fix above.
* `InstrumentedJedisPool` had a reference to its own `GenericObjectPoolConfig` that wasn't really being used. While I haven't seen any cases of this pool being used, I overrode the other methods on the `JedisPool` and `Pool<Jedis>` interfaces to use the delegate's internal pool instead.

There will be a followup PR that also cleans up Orca's configuration. Bullets 2 & 3 will allow me to remove all of the `@Deprecated` bean methods within Orca, pushing all redis client setup down to kork.

### DIFF
--- a/kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/JedisClientConfiguration.java
+++ b/kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/JedisClientConfiguration.java
@@ -17,21 +17,40 @@ package com.netflix.spinnaker.kork.jedis;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.netflix.spectator.api.Registry;
+import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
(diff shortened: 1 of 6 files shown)


ezimanyi:
Did a quick smoke test and the test case I had that was failing due to connection throttling works now if I pull this into Orca.

PR #193, opened 2018-09-18
feat(okhttp): add the metrics interceptor to the okhttp client configs

dreynaud:


dreynaud:
I guess this is the right place to have the metrics interceptor added by default, but I am not entirely confident this won't have weird side effects. Any strong opinions? For one thing, I am concerned that other services (like gate) will double publish the metrics since they take the config from kork and tack the metrics interceptor on top of it artisanally: https://github.com/spinnaker/gate/blob/master/gate-web/src/main/groovy/com/netflix/spinnaker/gate/config/RetrofitConfig.groovy#L51

ajordens:
In hindsight, maybe I should have added it here in the first place ... if we merge this than when `spinnaker-dependencies` is being bumped in each service we should just remove the second interceptor addition.

Should just be this set --> https://github.com/search?q=org%3Aspinnaker+OkHttpMetricsInterceptor&type=Code

kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttp3ClientConfiguration.groovy
```diff
@@ -39,10 +41,13 @@ import java.util.concurrent.TimeUnit
 @Component
 class OkHttp3ClientConfiguration {
   private final OkHttpClientConfigurationProperties okHttpClientConfigurationProperties
+  private final OkHttp3MetricsInterceptor okHttp3MetricsInterceptor
```
ajordens: Did you vet that this is actually created somewhere? Should be able to composite build between `gate` and `kork` and ran this new combination locally.

I recall explicitly `new OkHttpMetricsInterceptor(registry)` in each of the services.

dreynaud: What do you mean? Orca is currently using `OkHttp3MetricsInterceptor` in some places:
https://github.com/spinnaker/orca/blob/master/orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/RetrofitConfiguration.groovy#L76

PR #198, opened 2018-09-27
feat(telemetry): Instrumented dynamic proxy

robzienert:


kork-telemetry/src/main/java/com/netflix/spinnaker/kork/telemetry/InstrumentedProxy.java
```diff
@@ -0,0 +1,186 @@
...
+              invocationId(method)
+            ));
+          }
+          processed = true;
+          break;
+        }
+      }
+
+      if (!processed) {
+        addInstrumentedMethod(methodMetricIds, method, new MetricIds(
```
cfieber: feel like we either want to name the metrics after the proxied object type, and use a tag for the method name, or keep current naming and add the proxied object type as a tag

kork-telemetry/src/main/java/com/netflix/spinnaker/kork/telemetry/InstrumentedProxy.java
```diff
@@ -0,0 +1,186 @@
...
+  private final static String INVOCATIONS = "invocations";
+  private final static String TIMING = "timing";
+
+  private final Registry registry;
+  private final Object target;
+  private final String metricNamespace;
+
+  private final Map<Method, MetricIds> instrumentedMethodMetricIds;
+
+  public InstrumentedProxy(Registry registry, Object target, String metricNamespace) {
```
cfieber: might allow to pass in a set of tags to add to the generated metrics here

kork-telemetry/src/main/java/com/netflix/spinnaker/kork/telemetry/Instrumented.java
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.telemetry;
+
+import java.lang.annotation.*;
+
+@Documented
+@Target({ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Instrumented {
```
cfieber: might allow an array of strings of tagName, tagValue to add to the metrics on here

robzienert:
Added a test, had to do a bit of a refactor to where it now builds the metric IDs etc on-demand (caching them after first use), effectively the same functionality. Also added optional tags at both initialization-time and in the `@Instrumented` annotation.

PR #157, opened 2018-05-29
feat(build): separate release management opinions

cfieber:
see spinnaker/spinnaker#2844 for context

see spinnaker/spinnaker-gradle-project#39 for plugin changes

Release management opinions are in a separate spinnaker publishing plugin applied via an init script in the travis CI build shell script rather than in the core build.gradle

PR #199, opened 2018-10-01
fix(builds): Pulls in unopinionated spinnaker-gradle-project plugin

ttomsu:
Do not merge with release candidate libraries - needs https://github.com/spinnaker/spinnaker-gradle-project/pull/47 to be released.

Supercedes https://github.com/spinnaker/kork/pull/157

Ref: https://github.com/spinnaker/spinnaker/issues/2844

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,21 +15,21 @@
  */
 
 buildscript {
-
   repositories {
+    mavenCentral()
     jcenter()
     maven { url 'http://dl.bintray.com/spinnaker/gradle/' }
     maven { url "https://plugins.gradle.org/m2/" }
   }
   dependencies {
-    classpath 'com.netflix.spinnaker.gradle:spinnaker-gradle-project:4.0.0'
+    classpath 'com.netflix.spinnaker.gradle:spinnaker-dev-plugin:4.4.0-rc.5-composite'
     classpath "com.netflix.nebula:nebula-kotlin-plugin:$kotlinVersion"
     classpath "org.jetbrains.kotlin:kotlin-allopen:$kotlinVersion"
   }
 }
 
 allprojects {
-  apply plugin: 'spinnaker.project'
+  apply plugin: 'spinnaker.base-project'
   apply plugin: 'groovy'
 
   group = 'com.netflix.spinnaker.kork'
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -57,4 +57,11 @@ allprojects {
   spinnaker {
     dependenciesVersion = spinnakerDependenciesVersion
   }
+
+  //c&p this because NetflixOss reverts it to 1.7 and ends up getting applied last..
+  project.plugins.withType(JavaBasePlugin) {
```
dibyom: Does this have to be added to each project now?

gradle/buildViaTravis.sh
```diff
@@ -1,24 +1,27 @@
 #!/bin/bash
 # This script will build the project.
 
+GRADLE="./gradlew -I gradle/init-publish.gradle --no-daemon --max-workers=1"
```
dibyom: Should we be hardcode `max-workers` to 1? The default seems to be the number of CPUs

kork-web/src/main/groovy/com/netflix/spinnaker/config/CompositeBuildTest.groovy
```diff
@@ -0,0 +1,16 @@
+package com.netflix.spinnaker.config
+
+import groovy.util.logging.Slf4j
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+
```
dibyom: Is this file meant to be checked in?

build.gradle
ttomsu: I'm actually hoping Cam can point me to where I need to fix this in the plugin to not need the c&p

gradle/buildViaTravis.sh
ttomsu: For OSS Travis CI builds, I don't think you get more than 1, but I'm not 100% sure.

kork-web/src/main/groovy/com/netflix/spinnaker/config/CompositeBuildTest.groovy
ttomsu: How would you _know_ if your super special Kork change worked?!

ttomsu:
Going to merge this and release as an RC to test that publishing still works.

PR #205, opened 2018-11-05
fix(telemetry): Unwrap InvocationTargetException

ezimanyi:
The change to use an instrumeted proxy for Jenkins calls is masking error codes received from the Jenkins server. This is because exceptions thrown in InvocationHandler.invoke are wrapped in the checked exception InvocationTargetException before being returned to the caller. This ultimately leads to downstream code receiving an UndeclaredThrowableException rather than the particular exception.

### DIFF
--- a/kork-telemetry/src/main/java/com/netflix/spinnaker/kork/telemetry/InstrumentedProxy.java
+++ b/kork-telemetry/src/main/java/com/netflix/spinnaker/kork/telemetry/InstrumentedProxy.java
@@ -20,8 +20,10 @@ import com.netflix.spectator.api.Id;
 import com.netflix.spectator.api.Registry;
 import com.netflix.spectator.api.histogram.PercentileTimer;
 
+import com.netflix.spinnaker.kork.telemetry.MetricTags.ResultValue;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.InvocationTargetException;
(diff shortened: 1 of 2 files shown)


ezimanyi:
I noticed this as a change to return codes when making calls to get Jenkins info in Igor.  For example, consider the following where you pass a non-existent job number:
`curl -s http://<igor-endpoint>/builds/status/<number>/<jenkins-server>/<job> | json_pp`

Before adding instrumentation:
```
{
   "message" : "404 Not Found",
   "timestamp" : "2018-11-05T19:30:40.399+0000",
   "error" : "Not Found",
   "url" : "http://<jenkins-ip>/job/<job>/<build-number>/api/xml?exclude=/*/action[not(totalCount)]&tree=actions[failCount,skipCount,totalCount,urlName],duration,number,timestamp,result,building,url,fullDisplayName,artifacts[displayPath,fileName,relativePath]",
   "status" : 404,
   "exception" : "com.netflix.spinnaker.kork.web.exceptions.GenericExceptionHandlers$RetrofitErrorWrapper"
}
```
After adding instrumentation:
```
{
   "status" : 500,
   "exception" : "java.lang.reflect.UndeclaredThrowableException",
   "timestamp" : "2018-11-05T19:45:54.555+0000",
   "error" : "Internal Server Error",
   "message" : "No message available"
}
```

This change restores the return status to the first example above.

PR #209, opened 2018-11-28
chore(*): Applying codestyles

robzienert:
I haven't made a PR for spinnaker-gradle-project yet for this, but I've written a new plugin that formats Groovy, Java and Kotlin. For Groovy and Java, I've used _basically_ the Google Code Styles, with the small departure that continuation indentations are 2 spaces instead of 4.

There are two problems, one an annoyance, one an actual issue with a workaround:

1. For both Groovy and Java, I'm using the Eclipse formatter, which comes with one annoyance: Groovy formatting is _mostly_ correct, except when dealing with anonymous classes, it will create a continuation indent of 4 spaces, even though Java handles it perfectly fine. My theory here is that it's something to do with Groovy just not having a very easy to parse syntax. Example:

```groovy
// The way I would write it without formatting...
project.plugins.withType(SpotlessPlugin) { SpotlessPlugin spotless ->
  spotless.extension.java(new Action<JavaExtension>() {
    @Override
    void execute(JavaExtension javaExtension) {
      javaExtension.eclipse("4.8.0").configFile(project.rootProject.file("codestyle.xml"))
    }
  })
}

// The way that the formatter mangles it...
project.plugins.withType(SpotlessPlugin) { SpotlessPlugin spotless ->
  spotless.extension.java(new Action<JavaExtension>() {
        @Override
        void execute(JavaExtension javaExtension) {
          javaExtension.eclipse("4.8.0").configFile(project.rootProject.file("codestyle.xml"))
        }
      })
}
```

¯\\_(ツ)\_/¯

2. The Groovy formatter _will_ format Java classes that are in the `groovy` source set, unfortunately it will treat Java files with the Groovy formatter and the Groovy syntax is not backwards compatible with Java, specifically Groovy interfaces can't have method bodies like Java can with default methods. The fallout from this is that exceptions get thrown and everything is horrible. The fix is either: 1) Move the Java classes into the `java` source set, which is what I did in this PR and was only an issue in one spot. 2) Disable checking Java extension files in the `groovy` source set.

I don't like the second option, disabling Java. I can create specific rules for exclusion of files and expose that to our projects but I don't like the idea of making exceptions.

Take a look, there's a lot here, let me know if what it's generating looks acceptable. I'll try to tweak things to match what we agree on, but I'm leaning towards calling this "stable". The next step would be to create an IntelliJ formatter that uses these exact settings.

### DIFF
--- a/kork-telemetry/src/main/java/com/netflix/spinnaker/kork/telemetry/InstrumentedProxy.java
+++ b/kork-telemetry/src/main/java/com/netflix/spinnaker/kork/telemetry/InstrumentedProxy.java
@@ -20,8 +20,10 @@ import com.netflix.spectator.api.Id;
 import com.netflix.spectator.api.Registry;
 import com.netflix.spectator.api.histogram.PercentileTimer;
 
+import com.netflix.spinnaker.kork.telemetry.MetricTags.ResultValue;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.*;
@@ -79,20 +81,20 @@ public class InstrumentedProxy implements InvocationHandler {
   @Override
   public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
     MethodMetrics methodMetrics = getMethodMetrics(method);
-    if (methodMetrics != null) {
-      final long start = System.currentTimeMillis();
-      try {
-        Object result = method.invoke(target, args);
-        registry.counter(methodMetrics.invocationsId.withTag(RESULT_KEY, SUCCESS.toString())).increment();
-        recordTiming(methodMetrics.timingId.withTag(RESULT_KEY, SUCCESS.toString()), start);
-        return result;
-      } catch (Exception e) {
-        registry.counter(methodMetrics.invocationsId.withTag(RESULT_KEY, FAILURE.toString())).increment();
-        recordTiming(methodMetrics.timingId.withTag(RESULT_KEY, FAILURE.toString()), start);
-        throw e;
+    ResultValue resultValue = FAILURE;
+    final long start = System.currentTimeMillis();
+    try {
+      Object result = method.invoke(target, args);
+      resultValue = SUCCESS;
+      return result;
+    } catch (InvocationTargetException e) {
+      throw e.getCause();
+    } finally {
+      if (methodMetrics != null) {
+        registry.counter(methodMetrics.invocationsId.withTag(RESULT_KEY, resultValue.toString()))
+          .increment();
+        recordTiming(methodMetrics.timingId.withTag(RESULT_KEY, resultValue.toString()), start);
       }
-    } else {
-      return method.invoke(target, args);
     }
   }
 
--- a/kork-telemetry/src/test/groovy/com/netflix/spinnaker/kork/telemetry/InstrumentedProxySpec.groovy
+++ b/kork-telemetry/src/test/groovy/com/netflix/spinnaker/kork/telemetry/InstrumentedProxySpec.groovy
@@ -26,20 +26,28 @@ class InstrumentedProxySpec extends Specification {
   @Shared
   Registry registry = new DefaultRegistry()
 
+  MyContract subject = java.lang.reflect.Proxy.newProxyInstance(
+     getClass().classLoader,
+     [MyContract] as Class[],
+     new InstrumentedProxy(registry, new MyContractImpl(), "myns")
+  )
+
   def "should generate method metrics on init"() {
     when:
-    MyContract subject = java.lang.reflect.Proxy.newProxyInstance(
-      getClass().classLoader,
-      [MyContract] as Class[],
-      new InstrumentedProxy(registry, new MyContractImpl(), "myns")
-    )
-
     def ignored = subject.ignored()
     def result = subject.doStuff()
 
     then:
     ignored == "so sad"
     result == "did stuff"
+  }
+
+  def "should preserve the exception class"() {
+    when:
+    subject.throwError()
+
+    then:
+    thrown MyException
   }
 }
 
@@ -54,6 +62,8 @@ interface MyContract {
 
   @Instrumented(metricName = "sig1Long", tags = ["foo", "bar"])
   String sig1(Long p1);
+
+  void throwError()
 }
 
 class MyContractImpl implements MyContract {
@@ -77,4 +87,11 @@ class MyContractImpl implements MyContract {
   String sig1(Long p1) {
     return null
   }
+
+  @Override
+  void throwError() {
+    throw new MyException()
+  }
 }
+
+class MyException extends RuntimeException {}


robzienert:
A few things I'm noticing now that I've opened this:

* Constructor & method continuation indentation is gross.
* I need to figure out how to ignore the license headers

The alternative is adopting a different codestyle thing that really just adheres to something super basic like `.editorconfig` and calling it a day. I suspect this won't be sufficient, however.

kork-aws/src/main/groovy/com/netflix/spinnaker/kork/aws/bastion/BastionCredentialsProvider.groovy
```diff
@@ -70,14 +70,12 @@ class BastionCredentialsProvider implements AWSCredentialsProvider {
 
   private AWSCredentials getRemoteCredentials() {
     SimpleDateFormat format = new SimpleDateFormat(
-      "yyyy-MM-dd'T'HH:mm:ss'Z'", Locale.US);
+        "yyyy-MM-dd'T'HH:mm:ss'Z'", Locale.US);
```
ajordens: Q: Any control over the closing `);`

```
SimpleDateFormat format = new SimpleDateFormat(
    "yyyy-MM-dd'T'HH:mm:ss'Z'", Locale.US
);
```

kork-aws/src/main/groovy/com/netflix/spinnaker/kork/aws/pubsub/PubSubUtils.java
```diff
@@ -34,9 +31,9 @@
   private static final Logger log = LoggerFactory.getLogger(PubSubUtils.class);
 
   public static String ensureQueueExists(AmazonSQS amazonSQS,
```
ajordens: Do we have control over alignment of method args when split over multiple lines?

kork-aws/src/main/groovy/com/netflix/spinnaker/kork/aws/pubsub/PubSubUtils.java
```diff
@@ -45,8 +42,7 @@ public static String ensureQueueExists(AmazonSQS amazonSQS,
     attributes.put("MessageRetentionPeriod", Integer.toString(sqsMessageRetentionPeriodSeconds));
     amazonSQS.setQueueAttributes(
       queueUrl,
-      attributes
-    );
+      attributes);
```
ajordens: was it a conscious config to pull the `);` up?

kork-core/src/main/java/com/netflix/spinnaker/kork/metrics/SpectatorMetricWriter.java
```diff
@@ -86,8 +83,8 @@ private AtomicDouble getGaugeStorage(Id id) {
   }
 
   private AtomicLong getCounterStorage(Id id) {
-    final AtomicLong newCounter= new AtomicLong(0);
-    final AtomicLong existingCounter= counters.putIfAbsent(id, newCounter);
+    final AtomicLong newCounter = new AtomicLong(0);
```
ajordens: thank you!

kork-aws/src/main/groovy/com/netflix/spinnaker/kork/aws/bastion/BastionCredentialsProvider.groovy
robzienert: I didn't see anything about that, but I'll keep hunting that down. It was also bothering me.

kork-aws/src/main/groovy/com/netflix/spinnaker/kork/aws/pubsub/PubSubUtils.java
robzienert: Yes, I guess my change got reverted, that'll be fixed up real quick.

kork-aws/src/main/groovy/com/netflix/spinnaker/kork/aws/ARN.java
```diff
@@ -1,17 +1,15 @@
 /*
  * Copyright 2018 Netflix, Inc.
  *
- * Licensed under the Apache License, Version 2.0 (the "License")
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
+ * Licensed under the Apache License, Version 2.0 (the "License") you may not use this file except
```
ttomsu: It seems like we'd need to update the Spinnaker plugin that is adding this header to adjust the line breaks. 

https://github.com/nebula-plugins/gradle-netflixoss-project-plugin/blob/master/src/main/resources/netflixoss/HEADER

kork-core/src/main/java/com/netflix/spinnaker/kork/archaius/ArchaiusConfiguration.java
```diff
@@ -43,40 +40,49 @@
...
+  static class ArchaiusInitializingBeanPostProcessor implements BeanPostProcessor, Ordered {
     private final ConfigurableApplicationContext applicationContext;
     private final AbstractPollingScheduler pollingScheduler;
     private final SpringEnvironmentPolledConfigurationSource polledConfigurationSource;
     private final List<ClasspathPropertySource> propertyBindings;
     private final DynamicConfiguration configurationInstance;
 
-    public ArchaiusInitializingBeanPostProcessor(ConfigurableApplicationContext applicationContext, AbstractPollingScheduler pollingScheduler, SpringEnvironmentPolledConfigurationSource polledConfigurationSource, List<ClasspathPropertySource> propertyBindings) {
+    public ArchaiusInitializingBeanPostProcessor(ConfigurableApplicationContext applicationContext,
+      AbstractPollingScheduler pollingScheduler,
```
ttomsu: Thoughts on 4-space indent on new-lined parameter lists? To differentiate between the parameter list and the first code line.

ttomsu:
My preference would probably be for 120 char limit.

codestyle.xml
```diff
@@ -0,0 +1,318 @@
...
+<setting id="org.eclipse.jdt.core.formatter.insert_space_after_postfix_operator" value="do not insert"/>
+<setting id="org.eclipse.jdt.core.formatter.parentheses_positions_in_enum_constant_declaration" value="common_lines"/>
+<setting id="org.eclipse.jdt.core.formatter.insert_space_after_comma_in_for_increments" value="insert"/>
+<setting id="org.eclipse.jdt.core.formatter.insert_space_after_comma_in_type_arguments" value="insert"/>
+<setting id="org.eclipse.jdt.core.formatter.insert_space_before_comma_in_for_inits" value="do not insert"/>
+<setting id="org.eclipse.jdt.core.formatter.insert_new_line_in_empty_anonymous_type_declaration" value="do not insert"/>
+<setting id="org.eclipse.jdt.core.formatter.insert_space_after_semicolon_in_for" value="insert"/>
+<setting id="org.eclipse.jdt.core.formatter.align_with_spaces" value="false"/>
+<setting id="org.eclipse.jdt.core.formatter.disabling_tag" value="@formatter:off"/>
+<setting id="org.eclipse.jdt.core.formatter.continuation_indentation" value="1"/>
```
ttomsu: Maybe this is the value to tweak to 2 for constructor/method param lists?

PR #214, opened 2019-01-04
fix(artifacts): Ignore kind in serialized artifact

ezimanyi:
The kind field in artifacts is used only in the UI, and is not intended to be part of the server-side specification. Currently the artifact deserializer is putting the kind into the metadata hash; instead let's just ignore it as this is not intended to be used server-side at all.


ezimanyi:
Addresses one of the issues in https://github.com/spinnaker/spinnaker/issues/2654

PR #216, opened 2019-01-09
fix(artifacts): Add kind to artifact model

ezimanyi:
While artifact kind was initially intended to only exist on the front-end, we have been persisting it, and actually depend on its existence in the back-end to differentiate between artifacts configured as 'custom' with all fields visible and artifacts configured by selecting a particular type.

Let's officially add this field to the back-end so that persistence doesn't break in workflows that deserialize as an Artifact (instead of just a Map).


ezimanyi:
Closing for now; considering adding a boolean `isCustom` instead.

PR #220, opened 2019-01-16
feat(kork/secrets): Add storage secret engine and s3 implementation

ncknt:
First secret engine implementation


kork-secrets/kork-secrets.gradle
```diff
@@ -6,6 +6,7 @@ dependencies {
 
   compile spinnaker.dependency('spectatorWebSpring')
   compile spinnaker.dependency('lombok')
+  compile spinnaker.dependency('awsS3')
```
robzienert: This ain't gonna fly, unless you want to force any user of this new secrets system to use S3 - nonstarter.

`kork-secrets` should be the API contracts and like... maybe a default file-based (?) implementation or something. Then for S3, you'd wind up making a `kork-secrets-aws` that has an S3 implementation.

kork-secrets/src/main/java/com/netflix/spinnaker/config/secrets/engines/AbstractStorageSecretEngine.java
```diff
@@ -0,0 +1,140 @@
...
+import org.springframework.stereotype.Component;
+import org.yaml.snakeyaml.Yaml;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.*;
+
+@Component
```
robzienert: Why is this a Component? I don't think you can make abstract beans.

kork-secrets/src/main/java/com/netflix/spinnaker/config/secrets/engines/AbstractStorageSecretEngine.java
```diff
@@ -0,0 +1,140 @@
...
+public abstract class AbstractStorageSecretEngine implements SecretEngine {
+  protected final static String STORAGE_BUCKET = "b";
+  protected final static String STORAGE_REGION = "r";
+  protected final static String STORAGE_FILE_URI = "f";
+  protected final static String STORAGE_PROP_KEY = "k";
+
+  protected Map<String, Map<String,Object>> cache = new HashMap<>();
+
+  @Autowired
+  protected Yaml yamlParser;
```
robzienert: Avoid property-based autowiring, always prefer autowired constructors.

kork-secrets/src/main/java/com/netflix/spinnaker/config/secrets/engines/AbstractStorageSecretEngine.java
```diff
@@ -0,0 +1,140 @@
...
+
+
+  public String decrypt(EncryptedSecret encryptedSecret) throws SecretDecryptionException {
```
robzienert: This is a personal preference thing, but I really don't like checked exceptions. Do we need it?

kork-secrets/src/main/java/com/netflix/spinnaker/config/secrets/engines/AbstractStorageSecretEngine.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+  }
+
+  public EncryptedSecret encrypt(String secretToEncrypt) throws UnsupportedOperationException {
+    throw new UnsupportedOperationException("This operation is not supported");
+  }
+
+  protected abstract InputStream downloadRemoteFile(EncryptedSecret encryptedSecret) throws IOException;
+
+
```
robzienert: Mind whitespace.

kork-secrets/src/main/java/com/netflix/spinnaker/config/secrets/engines/AbstractStorageSecretEngine.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+  }
+
+  protected void parseAsYaml(String fileURI, InputStream inputStream) {
+    Map<String,Object> parsed = (Map<String, Object>) yamlParser.load(inputStream);
+    cache.put(fileURI, parsed);
+  }
+
+  protected String getParsedValue(String fileURI, String yamlPath) throws SecretDecryptionException {
+    String[] pathElts = yamlPath.split("\\.");
```
robzienert: You should probably use a `Splitter` instead.

kork-secrets/kork-secrets.gradle
ncknt: Well, that was the first engine with more to come. The engines register themselves into the registry but the user decides where things are stored. If we move it to a different module, don't we still have the issue of making that jar available at the service startup? Do you want to pick and choose which engines are present when you build your service?

kork-secrets/src/main/java/com/netflix/spinnaker/config/secrets/engines/AbstractStorageSecretEngine.java
ncknt: Oversight and refactor! Fixing...

kork-secrets/src/main/java/com/netflix/spinnaker/config/secrets/engines/AbstractStorageSecretEngine.java
ncknt: This is changing to a simple constructor b/c of the way it's used in the services where we try to instantiate this as early as possible in the Spring lifecycle. So changing too...

kork-secrets/src/main/java/com/netflix/spinnaker/config/secrets/engines/AbstractStorageSecretEngine.java
ncknt: Oversight. That exception should be unchecked. Fixing...

kork-secrets/kork-secrets.gradle
ncknt: Added `kork-secrets-aws`. Will add the runtime dependency to `kork-core` in another PR

kork-secrets/src/main/java/com/netflix/spinnaker/config/secrets/engines/AbstractStorageSecretEngine.java
ncknt: Fixed

kork-secrets/src/main/java/com/netflix/spinnaker/config/secrets/engines/AbstractStorageSecretEngine.java
ncknt: Added `Splitter`

kork-secrets-aws/kork-secrets-aws.gradle
```diff
@@ -0,0 +1,20 @@
+dependencies {
+  compile spinnaker.dependency('bootAutoConfigure')
+  compile spinnaker.dependency('bootActuator')
+  compile spinnaker.dependency('spectatorGc')
+  compile spinnaker.dependency('spectatorJvm')
+
+  compile spinnaker.dependency('spectatorWebSpring')
+  compile spinnaker.dependency('lombok')
+  compile spinnaker.dependency('awsS3')
```
robzienert: I'd just say depend on `kork-core` rather than duplicate all of these.

ncknt: Updated

PR #221, opened 2019-01-16
feat(kork/secrets): Add nop secret engine for testing

ncknt:


kork-secrets/src/main/java/com/netflix/spinnaker/config/secrets/engines/NopSecretEngine.java
```diff
@@ -0,0 +1,44 @@
...
+
+import com.netflix.spinnaker.config.secrets.EncryptedSecret;
+import com.netflix.spinnaker.config.secrets.SecretEngine;
+import org.springframework.stereotype.Component;
+
+/**
+ * Secret engine that returns its value ("v") parameter which can be null if not provided.
+ * Used for testing.
+ */
+@Component
```
robzienert: I think some additional considerations will need to be taken to wire these up. In this pattern, Spring will wire this up all of the time which isn't desirable if there's a different engine. We'll need some conditional configuration to ensure we're only loading up a single `SecretEngine`.

kork-secrets/src/main/java/com/netflix/spinnaker/config/secrets/engines/NopSecretEngine.java
```diff
@@ -0,0 +1,44 @@
...
+public class NopSecretEngine implements SecretEngine {
```
robzienert: `Noop`

PR #226, opened 2019-01-28
feat(config): add noop DynamicConfigService.

cfieber:
Mostly for unit testing as I've reimplemented this in a few different spots now.


ajordens:
LGTM

robfletcher:
Could just be Java 8 default method impls

cfieber:
@robfletcher I had a concern that would cause something bad in an existing implementation in groovy (where I've mostly made the equivalent in the past in spock tests. I've seen some strangeness previously with groovy + java8 interfaces and wasn't sure it would be entirely runtime compatible to pick up a new version with things as defaults.

PR #229, opened 2019-02-05
fix(secrets): return original file path if not encrypted

KathrynLewis:
`decryptAsFile()` should have been returning the original string passed in if that string is unencrypted. `decrypt()` also behaves this way, so the change here just checks if the string returned matches the string passed in.

### DIFF
--- a/kork-secrets/src/main/java/com/netflix/spinnaker/config/secrets/SecretManager.java
+++ b/kork-secrets/src/main/java/com/netflix/spinnaker/config/secrets/SecretManager.java
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.config.secrets;
 
+import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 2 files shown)


kork-secrets/src/main/java/com/netflix/spinnaker/config/secrets/SecretManager.java
```diff
@@ -91,8 +92,8 @@ public Path decryptAsFile(String filePathOrEncrypted) {
     }
 
     String decryptedContents = decrypt(filePathOrEncrypted);
-    if (decryptedContents == null) {
-      return null;
+    if (decryptedContents.equals(filePathOrEncrypted)) {
```
ttomsu: This will throw an NPE if decryptedContents is null - probably wrap in StringUtils.equals()

KathrynLewis: Ok, fixed!

PR #232, opened 2019-02-08
feat(core/secrets): decrypt application secrets

KathrynLewis:
Part of secrets management project: [spinnaker/spinnaker#3649](https://github.com/spinnaker/spinnaker/issues/3649)

Adds a Spring PropertySource to allow for decryption of secrets in the services when configs are read in.

### DIFF
--- a/kork-core/kork-core.gradle
+++ b/kork-core/kork-core.gradle
@@ -9,9 +9,10 @@ dependencies {
     compile spinnaker.dependency('spectatorGc')
     compile spinnaker.dependency('spectatorJvm')
     compile spinnaker.dependency('spectatorWebSpring')
-
+    compile project(":kork-secrets-aws")
 
     testCompile spinnaker.dependency('groovy')
     spinnaker.group('spockBase')
     testRuntime spinnaker.dependency('slf4jSimple')
+    testCompile(group: 'org.mockito', name: 'mockito-core', version: '2.23.4')
 }
--- a/kork-core/src/main/java/com/netflix/spinnaker/kork/secrets/SecretApplicationContextInitializer.java
+++ b/kork-core/src/main/java/com/netflix/spinnaker/kork/secrets/SecretApplicationContextInitializer.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2019 Armory, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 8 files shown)


kork-core/kork-core.gradle
```diff
@@ -9,7 +9,7 @@ dependencies {
     compile spinnaker.dependency('spectatorGc')
     compile spinnaker.dependency('spectatorJvm')
     compile spinnaker.dependency('spectatorWebSpring')
-
+    compile project(":kork-secrets-aws")
```
jkschneider: `kork-core` depends on `kork-secrets-aws`?

KathrynLewis: It's not required to build kork itself, but it's required when the services try to use the s3 decryption engine.

kork-core/src/main/java/com/netflix/spinnaker/kork/secrets/SecretApplicationContextInitializer.java
```diff
@@ -0,0 +1,63 @@
...
+  public void initialize(ConfigurableApplicationContext applicationContext) {
+    applicationContext.getBeanFactory().addBeanPostProcessor(new BeanPostProcessor() {
+      @Override
+      public Object postProcessBeforeInitialization(Object bean, String beanName) throws BeansException {
+        return bean;
+      }
+
+      @Override
+      public Object postProcessAfterInitialization(Object bean, String beanName) throws BeansException {
+        if (beanName.equals("secretManager")) {
```
robzienert: Prefer `"secretManager".equals(beanName)`. I doubt `beanName` is ever going to be null, but I tend to err on the side of absolute caution in this case.

kork-core/kork-core.gradle
robzienert: 🤔 I'm not sure of a better option than what we're doing here, but I'm not enthusiastic about bringing the AWS dependency tree along for the ride into all services even if they're not going to be using the S3 decryption engine.

@ttomsu @cfieber -- Any ideas here? Has this ship already sailed? As more decryption engines come along (GCS, Vault, ...) we're gonna get pretty bloated.

kork-core/src/main/java/com/netflix/spinnaker/kork/secrets/SecretApplicationContextInitializer.java
KathrynLewis: fixed

KathrynLewis:
Added tests

kork-core/kork-core.gradle
robzienert: Gonna just say we shipit as is.

PR #233, opened 2019-02-13
feat(sql): Support for named connection pools and dynamic routing

robzienert:


ajordens:
LGTM — I like it!

kork-sql/src/main/kotlin/com/netflix/spinnaker/kork/sql/routing/StaticDataSourceLookup.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.springframework.jdbc.datasource.lookup.DataSourceLookup
+import javax.sql.DataSource
+
+class StaticDataSourceLookup(
+  private val dataSources: Map<String, DataSource>
+) : DataSourceLookup {
+
+  override fun getDataSource(dataSourceName: String): DataSource {
+    return dataSources[dataSourceName]
+      ?: throw DataSourceRoutingException("Could not find data source by name $dataSourceName")
```
asher: Will `withPool("undefined")` throw a `DataSourceRoutingException` or use the default?

robzienert: 🤔 Yeah this is wrong. It shouldn't throw.

robzienert:
Fairly significant refactor of the originally submitted PR. Boils down to mainly introducing a `DataSourceFactory` interface that simplifies the process of overriding use of HikariCP. This is important especially for Netflix since we use MariaDB & Aurora - we can drop in our own `MariaDbDataSourceFactory` and get all of the routing functionality without any copy/paste.

Please take another look.

PR #236, opened 2019-02-25
fix(locking): Fixing race condition causing intermittent test failures

jeyrschabu:
- ensures fetching most up to date lock before attempting to release it
- added a requestId property on heartbeat requests for convenience


jeyrschabu:
This issue manifests itself when attempting to auto release a lock that has been assigned a different version via the heartbeat mechanism. The failure is due to a drift between the lock held in memory and the authoritative version stored in redis.

This fix ensures we are operating on the most up to date version of the lock.

PR #237, opened 2019-02-27
feat(secrets): clear secret caches at the end of each session

KathrynLewis:
We need to clear the secret caches at the end of each session so the code picks up any user changes to secret fields or files.

Part of https://github.com/spinnaker/spinnaker/issues/3649


KathrynLewis:
@robzienert thanks!

PR #238, opened 2019-02-27
feat(web): use histograms to track controller times

asher:


emjburns:
LGTM

ajordens:
FYI @ewiseblatt -- concerns with changing the type of this metric?

ewiseblatt:
I dont think so because percentile timers are compatible with regular
timers.
The existing dashboards and monitoring converts these timers into a pair of
counters
   <meter>__count and <meter>__totalTime depending on the "statistic" tag
value.
The percentile timer will add a 3rd counter <meter>__percentile which will
be ignored.

How do we interpret the statistic=percentile values? We would like the
refinement this change suggests.

PR #241, opened 2019-03-06
Hikari API change, H2 mysql compatibility mode

robfletcher:


kork-sql/kork-sql.gradle
```diff
@@ -8,7 +8,7 @@ dependencies {
   compile "org.springframework:spring-tx:${spinnaker.version("spring")}"
   compile spinnaker.dependency('jooq')
   compile "org.liquibase:liquibase-core:3.6.1"
-  compile 'com.zaxxer:HikariCP:2.5.1'
+  compile "com.zaxxer:HikariCP:${spinnaker.version('hikari')}"
```
robzienert: Just do `compile spinnaker.dependency('hikari')`

kork-sql/kork-sql.gradle
```diff
@@ -8,7 +8,7 @@ dependencies {
   compile "org.springframework:spring-tx:${spinnaker.version("spring")}"
   compile spinnaker.dependency('jooq')
   compile "org.liquibase:liquibase-core:3.6.1"
-  compile 'com.zaxxer:HikariCP:2.5.1'
```
robzienert: Also, I bet Spring has a different liquibase version...

kork-sql/kork-sql.gradle
robfletcher: I was too lazy to define the entire dependency. I just put the version in there.

PR #244, opened 2019-03-07
refactor(artifacts): Move artifact extraction to kork-core

ezimanyi:


ezimanyi:
I left all the copyright headers as they were, as this is just copied/refactored from Echo.

PR #247, opened 2019-03-08
 feat(secrets): move secret session and caching to kork

KathrynLewis:
Moving the SecretSession from halyard into kork to simplify secrets caching and cleanup. Follow up to https://github.com/spinnaker/kork/pull/237

@robzienert would you mind taking a look?

### DIFF
--- a/kork-secrets/src/main/java/com/netflix/spinnaker/config/secrets/SecretManager.java
+++ b/kork-secrets/src/main/java/com/netflix/spinnaker/config/secrets/SecretManager.java
@@ -16,7 +16,7 @@
 
 package com.netflix.spinnaker.config.secrets;
 
-import org.apache.commons.lang3.StringUtils;
+import lombok.Getter;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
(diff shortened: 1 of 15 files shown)


kork-secrets/src/main/java/com/netflix/spinnaker/config/secrets/SecretManager.java
```diff
@@ -87,23 +76,14 @@ public String decrypt(String configValue) {
...
-    if (decryptedFile != null) {
-      return decryptedFile;
-    }
-
-    String decryptedContents = decrypt(filePathOrEncrypted);
-    if (StringUtils.equals(decryptedContents, filePathOrEncrypted)) {
+    if (!EncryptedSecret.isEncryptedSecret(filePathOrEncrypted)) {
       return Paths.get(filePathOrEncrypted);
+    } else {
+      return createTempFile("tmp", decrypt(filePathOrEncrypted));
```
robzienert: Saving decrypted files to disk. 🤔 Are we _sure_ that's the pattern we want? What's the purpose of this?

KathrynLewis: This is how it was already set up. I was just changing the method name to be more clear. We do avoid creating temp files whenever possible, but sometimes the code needs the secrets to be in file form, such as for kube configs, google cloud json files, etc. We clean up temp files at the end of each thread in halyard.

robzienert: I see. That's disappointing (being forced into using the file system) but makes sense. Thanks!

PR #248, opened 2019-03-13
feat(secrets): improve error messages for S3 secret engine decryption

KathrynLewis:
The current error messages aren't as helpful as they could be. Adding a specific errors for the following:

- bucket unauthorized
- bucket not found
- file not found
- key not found


emjburns:
LGTM, thanks for the improvements!

PR #249, opened 2019-03-13
feat(secrets): decrypt saml keystore in services

KathrynLewis:
The saml keystore doesn't end with "file" or "path" so currently it won't be decrypted in the services. This change enables decryption.


kork-core/src/main/java/com/netflix/spinnaker/kork/secrets/SecretAwarePropertySource.java
```diff
@@ -39,7 +39,7 @@ public Object getProperty(String name) {
       if (secretManager == null) {
         throw new SecretException("No secret manager to decrypt value of " + name);
       }
-      if (name.toLowerCase().endsWith("file") || name.toLowerCase().endsWith("path")) {
+      if (name.toLowerCase().endsWith("file") || name.toLowerCase().endsWith("path") || name.equalsIgnoreCase("keystore")) {
```
robzienert: Consider moving this conditional into a static method if we anticipate it growing further.

PR #252, opened 2019-03-22
feat(secrets): add decryption as bytes for binary secret files

KathrynLewis:
Some secret files such as java keystores will require decryption in the form of bytes instead of strings.


KathrynLewis:
@dibyom

PR #253, opened 2019-03-26
Make lock manager test a reusable TCK

robfletcher:
I didn't end up actually implementing a second version but I don't think this is a bad idea regardless.


robfletcher:
I cannot figure out why this won't compile

PR #254, opened 2019-03-27
Split Retrofit stuff into own module, new module for Retrofit2

robfletcher:
This splits the Retrofit and OkHttp configuration from `kork-web` and into their own modules; `kork-okhttp3` and `kork-retrofit-legacy`. 
There's also now a new `kork-retrofit` module for Retrofit 2.


kork-retrofit-legacy/src/main/java/com/netflix/spinnaker/config/OkHttpClientConfiguration.java
```diff
@@ -0,0 +1,123 @@
...
+import org.springframework.stereotype.Component;
+import static com.squareup.okhttp.ConnectionSpec.MODERN_TLS;
+import static java.util.stream.Collectors.toList;
+
+/**
+ * @deprecated replaced by {@link OkHttp3ClientConfiguration}
+ */
+@Component
+@Deprecated // see OkHttp3ClientConfiguration
+public class OkHttpClientConfiguration {
```
robfletcher: Wondering if we can just drop this old OkHttp stuff. Are any apps still using it?

emjburns: I think definitely apps are still using it. I keep running into it all over the place and I'm never sure which to use.

emjburns:
Could you add a readme with some brief words on when to use which classes here (or an example of a class doing this well)? And maybe some words on the deprecation strategy? Because I'm bad at knowing what to use and I want to carry this good work on to other services.

kork-retrofit-legacy/src/main/java/com/netflix/spinnaker/config/OkHttpClientConfiguration.java
ajordens: Happy to be proven wrong but pretty sure most services are still using it.

kork-retrofit/src/main/java/com/netflix/spinnaker/config/Retrofit2ConfigurationProperties.kt
```diff
@@ -0,0 +1,9 @@
+package com.netflix.spinnaker.config
+
+import okhttp3.logging.HttpLoggingInterceptor
+import org.springframework.boot.context.properties.ConfigurationProperties
+
+@ConfigurationProperties(prefix = "retrofit2")
```
ajordens: Do we really need the `retrofit` vs `retrofit2` prefixes?

kork-okhttp3/kork-okhttp3.gradle
```diff
@@ -0,0 +1,9 @@
+apply plugin: "java-library"
+
+dependencies {
+  api project(":kork-security")
+  api "com.squareup.okhttp3:okhttp:${spinnaker.version("okHttp3")}"
+  api "org.springframework:spring-context:${spinnaker.version("spring")}"
+  api "org.springframework.boot:spring-boot-autoconfigure:${spinnaker.version("springBoot")}"
+  api "com.netflix.spectator:spectator-api:${spinnaker.version("spectator")}"
```
ajordens: api == the old compile scope?

kork-retrofit/kork-retrofit.gradle
```diff
@@ -0,0 +1,10 @@
+apply plugin: "java-library"
+apply from: "$rootDir/gradle/kotlin.gradle"
+
+dependencies {
+  api project(":kork-security")
+  api project(":kork-okhttp3")
+  api "com.squareup.retrofit2:retrofit:${spinnaker.version("retrofit2")}"
+
+  implementation "com.squareup.okhttp3:logging-interceptor:${spinnaker.version('okHttp3')}"
```
ajordens: implementation == the old provided scope?

kork-okhttp3/kork-okhttp3.gradle
robfletcher: Kind of. `api` is compile + transitively exported. `implementation` is compile + runtime only for consumers, `compileOnly` is not transitively exported at all.

kork-retrofit/kork-retrofit.gradle
robfletcher: See above

kork-retrofit/src/main/java/com/netflix/spinnaker/config/Retrofit2ConfigurationProperties.kt
robfletcher: Unfortunately the log level enum is incompatible between the two, so yeah.

cfieber: we could take the log level name as a string and do something to produce both enum values?

kork-retrofit-legacy/src/main/java/com/netflix/spinnaker/config/OkHttpClientConfiguration.java
robzienert: All apps are using it still.

kork-retrofit/src/main/java/com/netflix/spinnaker/config/Retrofit2ConfigurationProperties.kt
robzienert: Yes, we can attach a converter to this to generate the value. https://www.baeldung.com/spring-type-conversions

robfletcher:
I rebased this. @robzienert @cfieber can you just eyeball the two metrics interceptors?

PR #255, opened 2019-04-02
Generate a BOM file instead of using spinnaker-dependencies

robfletcher:
This PR adds 2 new modules to kork that use Gradle's new support for "platforms" (basically Maven Bill-of-Materials or BOM files).

- `spinnaker-dependencies` (the new module here, not the existing project) is a collection of the third-party dependencies we use across the Spinnaker µservice fleet. It also inherits from other BOMs where available which means we can benefit from version alignment done by other projects.
- `kork-bom` transitively exposes `spinnaker-dependencies` as well as specifying the versions of all the kork modules.

The intention is that services can pull in `kork-bom` and then declare dependencies without versions, depending on the version recommendations & alignment done here. In terms of what this looks like in use, take a look at:

- https://github.com/spinnaker/swabbie/pull/181
- https://github.com/spinnaker/keiko/pull/54

The existing spinnaker/spinnaker-dependencies has a number of issues:

- It’s got a circular dependency with kork.
- It doesn’t actually do any version alignment so you still end up debugging transitive dependency conflicts and sticking hacky fixes in individual projects.
- it’s yet another thing you have to rev when you want a new dependency version (I’m not sure this makes it much better unless you’re also changing kork, though).
- BOM files are the standard way of doing this kind of dependency version management and we can gain much value by inheriting from Spring, Kotlin, Jackson, OkHttp BOMs and just augmenting with our own additions. Look at the number of places where we’re doing nasty little bits of Gradle stuff to align Jackson versions for example. It’s copy ’n pasted in almost every Spinnaker service.
- Gradle as of v5 now supports BOMs natively, previously you had to use a plugin or home roll a mechanism like spinnaker-dependencies. It seems sensible to use the standard.
- I find spinnaker-dependencies as it stands can be quite opaque. For example if I pull in `spinnaker.group("jackson")` what am I actually getting?
- The `compile` and `testCompile` scopes used by spinnaker-dependencies are now deprecated in Gradle. This is only an issue if you’re using the `spinnaker.group(whatever)` syntax (which I never do) but it’s worth updating things, I think.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.3.1-bin.zip
```
robfletcher: Needed to upgrade Gradle to get BOM support

kork-artifacts/kork-artifacts.gradle
```diff
@@ -1,13 +1,18 @@
...
+  api(platform(project(":spinnaker-dependencies")))
+  
+  implementation project(":kork-core")
+  implementation "commons-io:commons-io"
+  implementation "com.hubspot.jinjava:jinjava"
 
-  compile spinnaker.dependency('jinjava')
-  compileOnly spinnaker.dependency("lombok")
+  compileOnly "org.projectlombok:lombok"
+  annotationProcessor "org.projectlombok:lombok:1.16.20"
```
robfletcher: For some reason this configuration doesn't pick up from the BOM. Investigating.

kork-bom/kork-bom.gradle
```diff
@@ -0,0 +1,36 @@
...
+
+publishing {
+  publications {
+    nebula(MavenPublication) {
+      from components.javaPlatform
+    }
+  }
+}
+
+dependencies {
```
robfletcher: This build generates `kork-bom` which imports the new `spinnaker-dependencies` BOM and just contains version recommendations for all the Kork modules.

spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -0,0 +1,97 @@
...
+    retrofit2     : "2.5.0",
+    slf4j         : "1.7.25",
+    spectator     : "0.75.0",
+    spek          : "1.1.5",
+    spock         : "1.1-groovy-2.4",
+    testcontainers: "1.10.7",
+  ]
+}
+
+dependencies {
```
robfletcher: This file is a BOM with recommendations for all the dependencies. All the modules depend on this and it is imported by `kork-bom`.

spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -0,0 +1,97 @@
...
+  api(platform("org.jetbrains.kotlin:kotlin-bom:1.3.21"))
+  api(platform("org.junit:junit-bom:5.4.1"))
+  api(platform("com.fasterxml.jackson:jackson-bom:2.9.8"))
+  api(platform("org.springframework.boot:spring-boot-dependencies:1.5.17.RELEASE"))
+  api(platform("io.strikt:strikt-bom:0.19.7"))
+  api(platform("com.squareup.okhttp3:okhttp-bom:3.12.0"))
```
robfletcher: Handily we can also import all these BOMs and benefit from their version recommendations.

robfletcher:
The init script we use for building on Travis is screwing up the Java language setting. Working on it now.

kork-dependencies/kork-dependencies.gradle
```diff
@@ -0,0 +1,97 @@
...
+    api("org.slf4j:jcl-over-slf4j:${versions.slf4j}")
+    api("org.slf4j:slf4j-api:${versions.slf4j}")
+    api("org.slf4j:slf4j-simple:${versions.slf4j}")
+    api("org.spockframework:spock-core:${versions.spock}")
+    api("org.spockframework:spock-spring:${versions.spock}")
+    api("org.testcontainers:mysql:${versions.testcontainers}")
+    api("org.yaml:snakeyaml:1.23")
+    api("redis.clients:jedis:2.9.3")
+  }
+}
```
robfletcher: This file is a BOM with recommendations for all the dependencies. All the modules depend on this and it is imported by `kork-bom`. Downstream projects should not need to depend on this directly as they'll get everything transitively from `kork-bom`.

kork-dependencies/kork-dependencies.gradle
```diff
@@ -0,0 +1,97 @@
...
+  ]
+}
+
+dependencies {
+  api(platform("org.jetbrains.kotlin:kotlin-bom:1.3.21"))
+  api(platform("org.junit:junit-bom:5.4.1"))
+  api(platform("com.fasterxml.jackson:jackson-bom:2.9.8"))
+  api(platform("org.springframework.boot:spring-boot-dependencies:1.5.17.RELEASE"))
+  api(platform("io.strikt:strikt-bom:0.19.7"))
+  api(platform("com.squareup.okhttp3:okhttp-bom:3.12.0"))
```
robfletcher: Handily we can also import all these BOMs and benefit from their version recommendations.

ezimanyi:
Looks interesting...would be interested in a brief overview of how this would change our dependency management approach and what bumping dependencies would look like in the future.

Is the idea that any of the modules expressed as `constraints` in `kork-dependencies.gradle` would be removed from `spinnaker-dependencies`, and microservices would instead:
* declare their dependencies without a version
* import `kork-dependencies` to get the constraints

And then would the plan be to move anything currently in `spinnaker-dependencies` into `kork.dependencies` and deprecate `spinnaker-dependencies`?  So going forward to bump a dependency you'd bump it in `kork-dependencies` and bump `kork` in each microservice (perhaps with the help of `spinnakerbot` for the last part)?

kork-dependencies/kork-dependencies.gradle
```diff
@@ -0,0 +1,97 @@
...
+publishing {
+  publications {
+    nebula(MavenPublication) {
+      from components.javaPlatform
+    }
+  }
+}
+
+ext {
+  versions = [
```
cfieber: thoughts on pulling all the version numbers up to here? Just thinking it might be nicer than having to scan a couple places, also might be easier to automate some version bumping stuff if its all specified in properties (could build with `-Pversions.foo=1.2.3` and emit an updated bom

kork-dependencies/kork-dependencies.gradle
cfieber: thoughts on renaming this to spinnaker-dependencies instead of kork-dependencies? There are likely many things specified in here that aren't actually in use in the kork project, but are in use in some of our downstreams. This is really meant to be the spinnaker platform BOM not just for kork, kork just happens to be the first consumer

cfieber:
@ezimanyi yeah that is exactly the goal, this is just a way better way of implementing the current spinnaker-dependencies

cfieber:
one of the nice side-effects of this change is that one link in the spinnaker dependencies chain goes away and there is no weird circular reference of kork uses spinnaker dependencies but spinnaker dependencies references korks version

one of the (IMO nice, but I think we have to address still) implications is that we shouldn't have other repos of spinnaker exporting shared libraries. (e.g. the fiat client library should probably be relocated into kork as a sub-project) which would let us keep the inter-project dependency graph directed in a sane way

essentially any code we want to share across multiple spinnaker projects should be a kork subproject and not anywhere else

kork-dependencies/kork-dependencies.gradle
robfletcher: I deliberately limited it to just the one I used more than once but I'm open to alternatives.

ezimanyi:
Awesome, yeah this looks like a great improvement over the current system.  CC @dibyom who's done a lot of work on builds + dependency analysis.

On the topic of using our services as libraries, @ttomsu @dibyom & I were just talking today about trying to remove the halyard -> (clouddriver, front50) dependency as this often breaks when people change public methods in those services.  (That one's a bit tricky as halyard depends on more of `clouddriver` than we'd want to push to `kork`, but the end goal remains the same.)

robfletcher:
Yeah, a bom file is the canonical way to do this kind of thing and has been growing in popularity. Now Gradle has added support it seems like a logical step. We’ve introduced a dependency on the Spring bom in a couple of newer projects, for example. The idea here is that other services can depend on `kork-bom` which transitively pulls in `kork-dependencies` as well as other boms that align versions of Jackson, Spring, Kotlin, etc. The leaf services should not need to specify versions for anything specified in `kork-bom`, etc. I think there is a conversation worth having about what dependencies we want to manage in this way as I don’t think it’s always necessary, and individual services may pull in specific versions of libraries where they’re likely the only consumer or where drift is not going to affect other dependencies (the way Jackson or slf4j tend to creep in to everywhere).

kork-dependencies/kork-dependencies.gradle
robfletcher: Sure. I had actually called it that & then renamed it.

robfletcher:
I’m going to cut a release from a branch tomorrow so we can start dogfooding this and iron out the kinks.

kork-dependencies/kork-dependencies.gradle
robfletcher: My only doubt is that the names in that versions map are rather arbitrary and the more there are the less clear it is which things they apply to. I'm considering whether we can extract the versions out to a properties file where we can update them easily but I'd still like there to be a better link to the actual artifact or group.

kork-artifacts/kork-artifacts.gradle
```diff
@@ -1,13 +1,16 @@
-dependencies {
-  compile project(":kork-core")
-
-  compile 'commons-io:commons-io:2.5'
+apply plugin: "java-library"
+apply plugin: "groovy"
+apply plugin: "io.freefair.lombok"
```
ttomsu: What does this plugin do?

kork-aws/kork-aws.gradle
```diff
@@ -13,17 +13,22 @@
...
-  compile spinnaker.dependency('awsCore')
-  compile spinnaker.dependency("aws")
-  compile 'com.jcraft:jsch.agentproxy.jsch:0.0.9'
-  compile 'com.jcraft:jsch.agentproxy.connector-factory:0.0.9'
-  compile 'com.aestasit.infrastructure.sshoogr:sshoogr:0.9.25'
+  api(platform(project(":spinnaker-dependencies")))
 
-  spinnaker.group("spockBase")
-}
+  api "org.springframework.boot:spring-boot-autoconfigure"
```
ttomsu: Are these really all the APIs we want to expose with kork-aws? Seems like the groovy and jinjava stuff could be grouped into something else - not sure what the other stuff is.

kork-core/kork-core.gradle
```diff
@@ -1,18 +1,30 @@
...
+  api "org.springframework.boot:spring-boot-starter-actuator"
+  api "com.netflix.archaius:archaius-core"
+  api "com.netflix.eureka:eureka-client"
+  api "com.amazonaws:aws-java-sdk-core"
+  api "com.netflix.spectator:spectator-api"
+  api "com.netflix.spectator:spectator-ext-aws"
+  api "com.netflix.spectator:spectator-ext-gc"
+  api "com.netflix.spectator:spectator-ext-jvm"
+  api "com.netflix.spectator:spectator-web-spring"
+  api project(":kork-secrets-aws")
```
ttomsu: Seems suspect to be in `core`

kork-core/kork-core.gradle
```diff
@@ -1,18 +1,30 @@
...
-    compile spinnaker.dependency('spectatorJvm')
-    compile spinnaker.dependency('spectatorWebSpring')
-    compile project(":kork-secrets-aws")
+  api(platform(project(":spinnaker-dependencies")))
+
+  api "org.springframework.boot:spring-boot-autoconfigure"
+  api "org.springframework.boot:spring-boot-starter-actuator"
+  api "com.netflix.archaius:archaius-core"
+  api "com.netflix.eureka:eureka-client"
+  api "com.amazonaws:aws-java-sdk-core"
```
ttomsu: Seems suspect to be in core

kork-expressions/kork-expressions.gradle
```diff
@@ -1,15 +1,19 @@
...
-  compile spinnaker.dependency('lombok')
-
-  testCompile "org.assertj:assertj-core:${spinnaker.version('assertj')}"
-  testCompile "org.junit.jupiter:junit-jupiter-api:${spinnaker.version('jupiter')}"
-  testRuntime "org.junit.jupiter:junit-jupiter-engine:${spinnaker.version('jupiter')}"
-  testRuntime spinnaker.dependency('slf4jSimple')
+  api(platform(project(":spinnaker-dependencies")))
+
+  api "com.fasterxml.jackson.core:jackson-databind"
+  api "org.springframework:spring-context"
```
ttomsu: Only having briefly looked at the `api` documentation, it seems we should be very selective and what things we export this way. I understand that you're trying to do the cutover "as is", and that we can clean it up later, but we should shoot to make a concerted effort in the longer term to reduce (eliminate?) the external libraries we export. Thoughts?

spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -0,0 +1,114 @@
...
+  publications {
+    nebula(MavenPublication) {
+      from components.javaPlatform
+    }
+  }
+}
+
+ext {
+  versions = [
+    aws           : "1.11.469",
```
ttomsu: Have you thought about where cloud-provider specific versions would go? If you put `aws` here, then each cloud provider will put (all) of their versioning stuff here. You may want to consider creating a cloud provider pattern here (say `versions.common`, `versions.providers`, `versions.providers.aws` etc)

kork-artifacts/kork-artifacts.gradle
robfletcher: I couldn’t get the `annotationProcessor` scope to pick up the version from the BOM. This plugin just applies the dependency for Lombok so I didn’t have to hard-code the version everywhere. I picked it up from the Lombok homepage here: https://projectlombok.org/setup/gradle

kork-expressions/kork-expressions.gradle
robfletcher: Definitely agree.

spinnaker-dependencies/spinnaker-dependencies.gradle
robfletcher: It’s possible we could split the bom up into multiple. For example we have a bunch of Groovy-related libs and Kotlin-related libs we might not want to pull in everywhere. At the same time pulling in a bom does not automatically bring all the dependencies in, just the constraints on their versions so having an Uber-bom that includes everything might make sense.

I’d also be happy to consider _only_ including the genuinely common things where we should align across all projects and let individual services manage their own versions for everything else. I’m not sure how that sits with the way Google is managing dependencies when bringing things into your monorepo.

robfletcher:
In terms of what this looks like in use. Take a look at:

- https://github.com/spinnaker/swabbie/pull/181
- https://github.com/spinnaker/keiko/pull/54

kork-core/kork-core.gradle
robfletcher: Maybe. I was trying to make the minimal changes to the actual dependencies otherwise I'll spend more time debugging that than actually getting the BOM working in consuming projects. I totally agree we should revisit and rationalize these dependencies at some point.

kork-artifacts/kork-artifacts.gradle
robfletcher: I've taken this back out again as I figured out how to get the `annotationProcessor` scope to get version recommendations from the BOM.

kork-aws/kork-aws.gradle
robfletcher: I'm really just porting over what was already there. I'd be totally on-board with a thorough audit of the way our dependencies transitivize but I feel like it would be better tackled separately from this change.

robfletcher:
@ttomsu Any thoughts as to how happy you guys would be to merge this? I'm aware you've developed some tooling around automatic version bumping that this would disrupt.

robzienert:
I think that tooling (spinnaker-grsdle-project), can easily be updated to autobump on kork release.

robfletcher:
Heads up, I intend to merge this later today unless someone shouts at me to not to.

PR #261, opened 2019-04-17
Removed OkHttp BOM as they are not supporting it going forward

robfletcher:


spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -50,23 +50,27 @@ dependencies {
...
-  api(platform("com.squareup.okhttp3:okhttp-bom:${versions.okhttp3}"))
   api(platform("io.strikt:strikt-bom:0.20.0"))
   api(platform("org.spockframework:spock-bom:1.1-groovy-2.4"))
 
   constraints {
     api("cglib:cglib-nodep:3.2.0")
     api("ch.qos.logback:logback-classic:1.2.3")
+    api("com.aestasit.infrastructure.sshoogr:sshoogr:0.9.25")
     api("com.google.apis:google-api-services-monitoring:v3-rev477-1.25.0")
+    api("com.google.guava:guava:27.1-jre")
```
robfletcher: Looks like we were getting Guava transitively from OkHttp before so I added an explicit reference.

cfieber: FWIW the guava we had in spinnaker-dependencies on the boot2 branch is actually `com.github.ben-manes.caffeine:guava` (which is nice because I'm all about unguava'ing our things in favor of caffeine where we are just using guava cachebuilder). (this looks like a change between master and boot2 spinnaker-dependencies)

PR #262, opened 2019-04-18
fix(kork/secrets): Ensure secret manager is loaded before Tomcat props

ncknt:
Prior to this change, the instantiation of the `SecretManager` was causing the change of the `PropertySource`. This should now happen earlier - before TomcatProperties is instantiated in Gate - and fix keyStore and trustStore issues when using secrets.


cfieber:
instead of bootstrapping secrets via an autoconfigured Configuration class and trying to get the ordering to line up nicely I think you'd be better served by hooking into the spring boot application lifecycle by registering an `ApplicationListener` for `ApplicationEnvironmentPreparedEvent`.

This is what we do for our internal implementation of an extra PropertySource that does in memory secret decryption via our internal crypto service.

Heres a rough scaffolding of how to wire into the application startup:
```java
package example;

public class ExampleSecretsConfigurer implements ApplicationListener<ApplicationEnvironmentPreparedEvent> {
 @Override
  public void onApplicationEvent(ApplicationEnvironmentPreparedEvent event) {
    ConfigurableEnvironment environment = event.getEnvironment();
    //add/replace PropertySources in environment...
  }
}
```

You then register the `ApplicationListener` in `META-INF/spring.factories` (same place you register an autoconfiguration bean):

```properties
org.springframework.context.ApplicationListener=example.ExampleSecretsConfigurer
```

This will get picked up before the main Spring-Boot ApplicationContext starts up, and ensures the Environment is set up to read encrypted properties before any `@ConfigurationProperties` beans or `@Value` annotations are processed.

ncknt:
@cfieber Thanks for the feedback. I tried the `ApplicationListener` approach but it depends on a bean (`SecretManager`) that need to be instantiated before others and could depend on some properties (e.g. when we add assumeRole for the s3 secret engine). I'm not quite sure how to wire up so `SecretManager` could be retrieved either from the ApplicationListener (too early) or lazily from the `PropertySource` (could still be too late). Any ideas?

cfieber:
Hmm yeah good point, in our case we are constructing the various decryption classes in our listener and don't have any `@Bean` dependencies.

ncknt:
@cfieber Yeah so it has to be when bean starting instantiating and early in the process which is what the `SecretConfiguration` controls. 
I missed the test failure -> updated them

PR #265, opened 2019-04-22
fix(bom): remove last bits of Spinnaker dependencies

marchello2000:


kork-core-tck/kork-core-tck.gradle
```diff
@@ -3,7 +3,7 @@ apply plugin: "groovy"
 
 dependencies {
   api(project(":kork-core"))
-
-  api(spinnaker.dependency("groovy"))
-  api(spinnaker.dependency("spock"))
+  
+  api("org.codehaus.groovy:groovy-all")
+  api("org.spockframework:spock-core")
```
marchello2000: @robfletcher i assume versions here aren't needed since they are defined in 
https://github.com/spinnaker/kork/blob/master/spinnaker-dependencies/spinnaker-dependencies.gradle (via spock-bom)?

cfieber: correct!

kork-core-tck/kork-core-tck.gradle
```diff
@@ -3,7 +3,7 @@ apply plugin: "groovy"
 
 dependencies {
   api(project(":kork-core"))
-
-  api(spinnaker.dependency("groovy"))
-  api(spinnaker.dependency("spock"))
+  
+  api("org.codehaus.groovy:groovy-all")
```
cfieber: I'd make groovy-all `implementation` instead of `api`, or remove it entirely from the dependencies if this builds without it explicitly declared.

The spock dependency should bring in a groovy dependency, and groovy-all is maybe something we want to get rid of in favor of the modular groovy dependencies because we are doing some shenannigans in the spinnaker-dependencies BOM module to try and do version alignment on groovy-all because we don't get that one defined in spring-boot-dependencies

kork-core-tck/kork-core-tck.gradle
robfletcher: Yeah, this ended up here due to this module being merged in after I added the bom

kork-core-tck/kork-core-tck.gradle
marchello2000: removing it didn't work, but changed to `implementation`

PR #267, opened 2019-04-22
fix(bom): make sure we get spock 1.1

marchello2000:
Use Spock 1.1 that we explicitly specify vs one provided in spring-dependencies


marchello2000:
> Might want to get @emjburns’ take on this, IIRC she had some issues with the newer Spock.

 1.1 is what we use in orca today - when i pulled in kork-bom which downgraded spock to 1.0 i got a bunch of UT failures. With this they are gone

emjburns:
👍 the newer spock is 1.3, and spock 1.3 groovy 2.5 is what's giving errors. Only a boot2 problem right now.

PR #268, opened 2019-04-23
feat(micrometer): Delegate to Micrometer registry

jonsie:
@robzienert 

Spectator delegates to Micrometer.  Downstream services pull in the registry type(s), for example `io.micrometer:micrometer-registry-statsd`, and add the necessary config:

```
management:
    metrics:
        export:
            statsd:
                enabled: true
                flavor: datadog
                host: localhost
                port: 8125
```


cfieber:
@jonsie can you rebase this on boot2 again, I just pushed a fix for the travis failure (I think, lets find out!)

jonsie:
@cfieber Looks like that worked.

cfieber:
I would like to test this out in another service just to confirm that MicrometerRegistry is equivalent to DefaultRegistry in the absence of any other configuration or micrometer registry type being present (essentially acting like a noop registry) but otherwise this LGTM.

jonsie:
@cfieber Actually, it looks like this implementation causes `/spectator/metrics` to throw a 500 with the exception ```java.lang.IllegalArgumentException: There is already a registered meter of a different type with the same name``` - an exception in Micrometer's `MeterRegistry`.  I thought it was safe to add back `MetricsController` here, but apparently not.  I'll need to circle back to this.

cfieber:
sounds good.

it seems like /spectator/metrics becomes unnecessary if you are going to use the micrometer backend to export metrics

you could potentially leave the existing `@ConditionalOnMissingBean` registry creating the spectator `DefaultRegistry`,  and add another `@Bean` `MicrometerRegistry` `@ConditionalOnProperty('metrics.micrometer.enabled')` and make the `MetricsController` also `@ConditionalOnBean(DefaultRegistry)`

cfieber:
can you reopen this against master now that we merged boot2

PR #270, opened 2019-04-24
refactor(headers): Refactor header propagation logic

marchello2000:
This also adds ability to propagate any `X-SPINNAKER-*` header automatically
And, ability to notify with a metric and a log when a request doesn't have
headers needed to successfully pass identity of the caller
(i.e. `X-SPINNAKER-USER` and `X-SPINNAKER-ACCOUNTS`)

symmetric PR in orca

### DIFF
--- a/kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequest.java
+++ b/kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequest.java
@@ -29,11 +29,64 @@ import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.util.CollectionUtils;
 
 public class AuthenticatedRequest {
-  public static final String SPINNAKER_USER = "X-SPINNAKER-USER";
-  public static final String SPINNAKER_ACCOUNTS = "X-SPINNAKER-ACCOUNTS";
-  public static final String SPINNAKER_USER_ORIGIN = "X-SPINNAKER-USER-ORIGIN";
-  public static final String SPINNAKER_REQUEST_ID = "X-SPINNAKER-REQUEST-ID";
-  public static final String SPINNAKER_EXECUTION_ID = "X-SPINNAKER-EXECUTION-ID";
+  /**
+   * Known X-SPINNAKER headers, but any X-SPINNAKER-* key in the MDC will be automatically
+   * propagated to the HTTP headers.
+   *
+   * <p>Use makeCustomerHeader() to add customer headers
+   */
+  public enum Header {
+    USER("X-SPINNAKER-USER", true),
+    ACCOUNTS("X-SPINNAKER-ACCOUNTS", true),
+    USER_ORIGIN("X-SPINNAKER-USER-ORIGIN", false),
+    REQUEST_ID("X-SPINNAKER-REQUEST-ID", false),
+    EXECUTION_ID("X-SPINNAKER-EXECUTION-ID", false),
+    APPLICATION("X-SPINNAKER-APPLICATION", false);
+
+    private String header;
+    private boolean isRequired;
+
+    Header(String header, boolean isRequired) {
+      this.header = header;
+      this.isRequired = isRequired;
+    }
+
+    public String getHeader() {
+      return header;
+    }
+
+    public boolean isRequired() {
+      return isRequired;
+    }
+
+    public static String XSpinnakerPrefix = "X-SPINNAKER-";
+    public static String XSpinnakerAnonymous = XSpinnakerPrefix + "ANONYMOUS";
+
+    public static String makeCustomHeader(String header) {
+      return XSpinnakerPrefix + header.toUpperCase();
+    }
+  }
+
+  /**
+   * Allow a given HTTP call to be anonymous. Normally, all requests to Spinnaker services should be
+   * authenticated (i.e. include USER & ACCOUNTS HTTP headers). However, in specific cases it is
+   * necessary to make an anonymous call. If an anonymous call is made that is not wrapped in this
+   * method, it will result in a log message and a metric being logged (indicating a potential bug).
+   * Use this method to avoid the log and metric. To make an anonymous call wrap it in this
+   * function, e.g.
+   *
+   * <p>AuthenticatedRequest.allowAnonymous(() -> { // do HTTP call here });
+   */
+  public static <V> V allowAnonymous(Callable<V> closure) throws Exception {
+    String originalValue = MDC.get(Header.XSpinnakerAnonymous);
+    MDC.put(Header.XSpinnakerAnonymous, "anonymous");
+
+    try {
+      return closure.call();
+    } finally {
+      setOrRemoveMdc(Header.XSpinnakerAnonymous, originalValue);
+    }
+  }
 
   public static <V> Callable<V> propagate(Callable<V> closure) {
     return propagate(closure, true, principal());
@@ -55,43 +108,20 @@ public class AuthenticatedRequest {
     String executionId = getSpinnakerExecutionId().orElse(null);
     String requestId = getSpinnakerRequestId().orElse(null);
     String spinnakerAccounts = getSpinnakerAccounts(principal).orElse(null);
+    String spinnakerApp = getSpinnakerApplication().orElse(null);
 
     return () -> {
-      String originalSpinnakerUser = MDC.get(SPINNAKER_USER);
-      String originalSpinnakerUserOrigin = MDC.get(SPINNAKER_USER_ORIGIN);
-      String originalSpinnakerAccounts = MDC.get(SPINNAKER_ACCOUNTS);
-      String originalSpinnakerRequestId = MDC.get(SPINNAKER_REQUEST_ID);
-      String originalSpinnakerExecutionId = MDC.get(SPINNAKER_EXECUTION_ID);
+      // Deal with (set/reset) known X-SPINNAKER headers, all others will just stick around
+      Map originalMdc = MDC.getCopyOfContextMap();
+
       try {
-        if (spinnakerUser != null) {
-          MDC.put(SPINNAKER_USER, spinnakerUser);
-        } else {
-          MDC.remove(SPINNAKER_USER);
-        }
+        setOrRemoveMdc(Header.USER.getHeader(), spinnakerUser);
+        setOrRemoveMdc(Header.USER_ORIGIN.getHeader(), userOrigin);
+        setOrRemoveMdc(Header.ACCOUNTS.getHeader(), spinnakerAccounts);
+        setOrRemoveMdc(Header.REQUEST_ID.getHeader(), executionId);
+        setOrRemoveMdc(Header.EXECUTION_ID.getHeader(), requestId);
+        setOrRemoveMdc(Header.APPLICATION.getHeader(), spinnakerApp);
 
-        if (userOrigin != null) {
-          MDC.put(SPINNAKER_USER_ORIGIN, userOrigin);
-        } else {
-          MDC.remove(SPINNAKER_USER_ORIGIN);
-        }
-
-        if (spinnakerAccounts != null) {
-          MDC.put(SPINNAKER_ACCOUNTS, spinnakerAccounts);
-        } else {
-          MDC.remove(SPINNAKER_ACCOUNTS);
-        }
-
-        if (executionId != null) {
-          MDC.put(SPINNAKER_EXECUTION_ID, executionId);
-        } else {
-          MDC.remove(SPINNAKER_EXECUTION_ID);
-        }
-
-        if (requestId != null) {
-          MDC.put(SPINNAKER_REQUEST_ID, requestId);
-        } else {
-          MDC.remove(SPINNAKER_REQUEST_ID);
-        }
         return closure.call();
       } finally {
         MDC.clear();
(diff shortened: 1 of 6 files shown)


marchello2000:
Still rebasing some code - but ready for the first pass of review

kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
```diff
@@ -0,0 +1,139 @@
...
+      boolean missingAuthHeaders = missingHeaders.size() > 0;
+
+      if (missingAuthHeaders) {
+        List<String> stack = Arrays.stream(Thread.currentThread().getStackTrace())
+          .map(StackTraceElement::toString)
+          .filter(x -> x.contains(packageName))
+          .collect(Collectors.toList());
+
+        String stackTrace = String.join("\n\tat ", stack);
+        log.warn(
```
marchello2000: this is an example of what the log looks like:
```
2019-04-24 09:02:30.661  WARN 46010 --- [      events-17] c.n.s.okhttp.OkHttp3MetricsInterceptor   : [] Request POST:https://echo-prestaging.us-west-2.spinnaker.mgmt.netflix.net/:https://echo-prestaging.us-west-2.spinnaker.mgmt.netflix.net/ does not have authentication headers and will be treated as anonymous.
at com.netflix.spinnaker.okhttp.MetricsInterceptor.intercept(MetricsInterceptor.java:83)
	at com.netflix.spinnaker.okhttp.OkHttp3MetricsInterceptor.intercept(OkHttp3MetricsInterceptor.java:33)
	at com.netflix.spinnaker.orca.echo.spring.EchoNotifyingExecutionListener.beforeExecution(EchoNotifyingExecutionListener.groovy:66)
	at com.netflix.spinnaker.orca.events.ExecutionListenerAdapter.onExecutionStarted(ExecutionListenerAdapter.java:59)
	at com.netflix.spinnaker.orca.events.ExecutionListenerAdapter.onApplicationEvent(ExecutionListenerAdapter.java:48)
	at com.netflix.spinnaker.orca.events.ExecutionListenerAdapter.onApplicationEvent(ExecutionListenerAdapter.java:32)
```

kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
```diff
@@ -0,0 +1,139 @@
...
+ */
+class MetricsInterceptor {
+  private final Registry registry;
+  private final Logger log;
+  private final String packageName;
+
+  MetricsInterceptor(Registry registry) {
+    this.registry = registry;
+    this.log = LoggerFactory.getLogger(getClass());
+    this.packageName = this.getClass().getPackage().getName().split("\\.")[1];
```
marchello2000: not sure how we feel about this, but this is used to filter the call stack on line 89. making it much more paleteable

kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequest.java
```diff
@@ -30,11 +30,64 @@
...
-  public static final String SPINNAKER_USER_ORIGIN = "X-SPINNAKER-USER-ORIGIN";
-  public static final String SPINNAKER_REQUEST_ID = "X-SPINNAKER-REQUEST-ID";
-  public static final String SPINNAKER_EXECUTION_ID = "X-SPINNAKER-EXECUTION-ID";
+  /**
+   * Known X-SPINNAKER headers, but any X-SPINNAKER-* key in the MDC
+   * will be automatically propagated to the HTTP headers.
+   *
+   * Use makeCustomerHeader() to add customer headers
+   */
+  public enum SpinnakerHeader {
```
ajordens: Maybe `s/SpinnakerHeader/Header` to cut down on the verbosity

kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequest.java
```diff
@@ -30,11 +30,64 @@
...
+    USER("X-SPINNAKER-USER", true),
+    ACCOUNTS("X-SPINNAKER-ACCOUNTS", true),
+    USER_ORIGIN("X-SPINNAKER-USER-ORIGIN", false),
+    REQUEST_ID("X-SPINNAKER-REQUEST-ID", false),
+    EXECUTION_ID("X-SPINNAKER-EXECUTION-ID",false),
```
ajordens: space needed

kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequest.java
```diff
@@ -30,11 +30,64 @@
...
+    }
+  }
+
+  /**
+   * Allow a given HTTP call to be anonymous.
+   * Normally, when an HTTP call is missing required headers (USER & ACCOUNTS)
+   * a log message will be generated and a metric logged. If a call is expected
+   * to be anonymous, wrap it in this function, e.g.
+   *
+   * AuthenticatedRequest.allowAnonymous(() -> {
```
ajordens: Need to sleep on whether this is the right wording to use ... vs something else that could imply a more system/internal caller.

The intent is legitimate.

kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
+/**
+ * Sort of generic class to wrap okhttp and okhttp3 interception logic
+ * since we basically do the same thing and it's annoying to duplicate it
+ */
+class MetricsInterceptor {
```
ajordens: Hmmm ... not totally convinced on merging the okhttp and okhttp3 bits, need to sleep on it (given that the former will be going away sooner than later).

Seems more difficult to read having to worry about the two versions in the same method implementation.

kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
ajordens: does it give better results than just a blind `com.netflix.spinnaker` package prefix?

kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
ajordens: this is pretty slick, I like it.

kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
```diff
@@ -0,0 +1,139 @@
...
+        );
+      }
+
+      recordTimer(registry, url, System.nanoTime() - start, statusCode, wasSuccessful, !missingAuthHeaders);
+      recordMissingHeaders(registry, missingHeaders);
+    }
+  }
+
+  private static void recordMissingHeaders(Registry registry, List<String> missingHeaders) {
+    registry.counter("okhttp.requests.missingHeaders",
```
ajordens: this is actually recording anything ... no calls to increment.

kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
```diff
@@ -0,0 +1,139 @@
...
+    int statusCode,
+    boolean wasSuccessful,
+    boolean hasAuthHeaders) {
+    registry.timer(
+      registry.createId("okhttp.requests")
+        .withTag("requestHost", requestUrl.getHost())
+        .withTag("statusCode", String.valueOf(statusCode))
+        .withTag("status", bucket(statusCode))
+        .withTag("success", wasSuccessful)
+        .withTag("hasAuthHeaders", hasAuthHeaders)
```
ajordens: maybe `s/hasAuthHeaders/authenticated`?

kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
marchello2000: I was worried that we might move the package names at some future point...? but happy to change it to `com.netflix.spinnaker` - same result today

kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
marchello2000: doh, thanks!

kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequest.java
marchello2000: thanks! we need @robzienert autoformatter thing..

kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
marchello2000: yeah, i first added it to the two, then i was annoyed that the code was duplicated and thought it would be cleaner as a single one... then it got messier than i wanted. I still think it's a bit cleaner (and only 1 place to fix/mess up). But happy to re-examine

kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequest.java
marchello2000: let me take another stab

kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
ajordens: Ok, let's do `com.netflix.spinnaker` for the explicitness ... changing packages will wreck havoc anyhow.

kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
ajordens: TBH I'm still not a huge fan ... but on the fence as it's not something that really needs to be touched very often.

I'd be more onside if the javadoc on this was a little more "professional"?

```
{@code MetricsInterceptor} is an encapsulation of common interception logic relevant to both
okhttp and okhttp3.

It is implemented as a single class with the expectation that legacy okhttp usage is on the way out and the interceptor logic will not be necessary longterm.
```

kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
```diff
@@ -0,0 +1,144 @@
...
+      if (missingAuthHeaders) {
+        List<String> stack =
+            Arrays.stream(Thread.currentThread().getStackTrace())
+                .map(StackTraceElement::toString)
+                .filter(x -> x.contains("com.netflix.spinnaker"))
+                .collect(Collectors.toList());
+
+        String stackTrace = String.join("\n\tat ", stack);
+        log.warn(
+            String.format(
```
ajordens: Should this output include the missing headers?

kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
```diff
@@ -0,0 +1,144 @@
...
+      }
+
+      recordTimer(
+          registry, url, System.nanoTime() - start, statusCode, wasSuccessful, !missingAuthHeaders);
+      recordMissingHeaders(registry, missingHeaders);
+    }
+  }
+
+  private static void recordMissingHeaders(Registry registry, List<String> missingHeaders) {
+    registry
```
ajordens: Not sure about a metric with the missing headers ... can just get that from logs based on the `authenticated` tag on the `recordTimer` below?

kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
```diff
@@ -0,0 +1,144 @@
...
+class MetricsInterceptor {
+  private final Registry registry;
+  private final Logger log;
+
+  MetricsInterceptor(Registry registry) {
+    this.registry = registry;
+    this.log = LoggerFactory.getLogger(getClass());
+  }
+
+  Object intercept(Object chainObject) throws IOException {
```
ajordens: Maybe rename this method and make it `protected final` ... than it's clearer that the subclasses need to call it directly rather than a magical `super.intercept()`.

kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
marchello2000: I can/will do professional

kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
marchello2000: Will do, i will also rename it - it's weird that it has the same name (but different signature) as `com.squareup.okhttp.Interceptor` i think that will make it clear

kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
marchello2000: yeah, that's a good idea, thanks!

kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
marchello2000: yeah, you're right... prob not useful, will remove

ajordens:
LGTM

PR #271, opened 2019-04-24
updates kork to spring boot 2 becomes the source of spinnaker-dependencies

cfieber:
This is the set of changes that upgraded the components in the kork library to spring-boot 2.

As part of this project we have removed the use of the spinnaker-dependencies yml file in favour of generating a dependency BOM file as this is now natively supported functionality in gradle.

The new spinnaker-dependencies module in kork serves as a list of 3rd party libraries. It extends spring-boot-platform BOM as well as including other BOM files where available for version alignment (jackson, junit, etc).

Additionally there is a kork-bom module that exports version opinions of both the spinnaker-dependencies as well as all the components of kork.

Consumers now import kork-bom for version opinions where they previously would have used the spinnaker-dependency gradle plugin.

We have been publishing releases from this branch, integrating them with boot2 upgrades in other services (fiat, igor, echo, keel have all been successfully running in production for us on boot2) so now it's time to get out of branch management land.


spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -127,5 +135,13 @@ dependencies {
     api("org.threeten:threeten-extra:1.0")
     api("org.yaml:snakeyaml:1.23")
     api("redis.clients:jedis:2.9.3")
+    api("com.google.apis:google-api-services-cloudbuild:v1-rev836-1.25.0")
+    api("com.google.apis:google-api-services-compute:alpha-rev102-1.25.0")
+    api("com.google.api-client:google-api-client:1.28.0")
+    api("rg.apache.commons:commons-exec:1.3")
```
cfieber: that seems rong

kork-stackdriver/src/main/java/com/netflix/spinnaker/config/StackdriverConfig.java
```diff
@@ -183,8 +176,9 @@ public boolean test(Measurement measurement) {
       log.error("Ignoring prototypeFilterPath because it is not yet supported.");
       measurementFilter = null;
       log.info("Configuring stackdriver filter from {}", prototypeFilterPath);
-      measurementFilter = PrototypeMeasurementFilter.loadFromPath(
-           prototypeFilterPath).and(filterNotSpring);
+      // TODO(rz): Make this all work with Boot 2
+//      measurementFilter = PrototypeMeasurementFilter.loadFromPath(
+//           prototypeFilterPath).and(filterNotSpring);
```
robfletcher: Is this something that needs doing?

spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -80,17 +84,20 @@ dependencies {
     api("com.netflix.spinnaker.embedded-redis:embedded-redis:0.8.0")
     api("com.netflix.spinnaker.moniker:moniker:0.2.0")
     api("com.nhaarman:mockito-kotlin:1.5.0")
-    api("com.squareup.okhttp3:logging-interceptor:${versions.okhttp3}")
+    api("com.squareup.okhttp3:logging-interceptor:${versions.okhttp3}") // see https://github.com/square/okhttp/issues/4960
```
robfletcher: Should remove this comment, I think

spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -127,5 +136,12 @@ dependencies {
     api("org.threeten:threeten-extra:1.0")
     api("org.yaml:snakeyaml:1.23")
     api("redis.clients:jedis:2.9.3")
+    api("com.google.apis:google-api-services-cloudbuild:v1-rev836-1.25.0")
+    api("com.google.apis:google-api-services-compute:alpha-rev102-1.25.0")
+    api("com.google.api-client:google-api-client:1.28.0")
+    api("com.amazonaws:aws-java-sdk-s3:${versions.aws}")
+    api("com.amazonaws:aws-java-sdk:${versions.aws}")
```
robfletcher: These should be provided by the AWS BOM, why do we respecify them?

cfieber: The -s3 one definitely is
the `aws-java-sdk` dependency is an unfortunate one because it brings in the entirety of all the other aws sdk libraries. It isn't declared in the bom file.

Ideally we would get rid of uses of aws-java-sdk but that is probably work best left for a post-merged boot2 world

kork-stackdriver/src/main/java/com/netflix/spinnaker/config/StackdriverConfig.java
robzienert: I have no idea. Probably @ewiseblatt ?

spinnaker-dependencies/spinnaker-dependencies.gradle
robzienert: I think I'd also like to break out `kork-core` into `-core`, `-core-aws`, `-core-netflix`, `-core-google`, etc...

robfletcher: I did get rid of a usage of `aws-java-sdk` somewhere (in one of the Kork modules IIRC).

Can we take the `-s3` out of here, then?

kork-stackdriver/src/main/java/com/netflix/spinnaker/config/StackdriverConfig.java
cfieber: I dunno why it was commented out, it should be fine as it is on master

PR #273, opened 2019-04-26
fix(secrets): cleans up kork-secrets dependency

cfieber:
moves `kork-secrets` out of `kork-core`, and moves implementation code out of the
`com.netflix.spinnaker.config` package which is reserved as a `@ComponentScan` location
for plugin bootstrapping.

This change is breaking in that now `kork-secrets` and/or `kork-secrets-aws` needs to be
explicitly added as a dependency, but this should only happen at the final runtime
module (e.g the `-web` module for each service) and not be a part of the dependency graph 
for all projects via kork-core.

There should be no other change required than adding the dependencies at the appropriate
module. `SecretManager` is set up for autoconfiguration, and it has an `@ComponentScan`
to detect any beans under the `com.netflix.spinnaker.kork.secrets` package.

There were a few dependencies that fell out of the graph for `kork-core` as a result of
this change, most notably `guava`. I elected to implement the library functionality instead
of bringing in the missing dependencies since `kork-core` is included early and often in
dependency graphs and the missing library functionality was trivial.


cfieber:
@ncknt @KathrynLewis please review

cfieber:
I'll ensure all the services have `kork-secrets-aws` as an explicit dependency before I merge this

ncknt:
LGTM

PR #275, opened 2019-04-29
fix(spring): Expose actuator endpoints by default

robzienert:


ajordens:
Any fear that this creates a more open / less secure out-of-the-box setup?

kork-core/src/main/java/com/netflix/spinnaker/kork/spring/SpringBoot1CompatibilityApplicationListener.java
```diff
@@ -29,5 +29,8 @@ public void onApplicationEvent(ApplicationEnvironmentPreparedEvent event) {
 
     // Put spring endpoints on / instead of /actuator (for /health backwards compat).
     System.setProperty("management.endpoints.web.base-path", "/");
+
+    // Expose all actuator endpoints.
+    System.setProperty("management.endpoints.web.exposure.include", "*");
```
cfieber: I wonder if this is the best way to set these defaults - I sort of suspect it would be hard to override this value in spring's config hierarchy since it is applying as a system property.

At the very least only setting this value if its not already set would be worthwhile.

Maybe adding a config source to the environment would play nicer with overriding this from yaml and/or system property configuration

robzienert: Sure, that's probably a better way to do it. I'll update.

PR #280, opened 2019-05-02
fix(javadoc): use entities for invalid javadoc.

cfieber:
javadoc task was failing on `&` and `>` characters when running a release build


marchello2000:
ugh, sorry. What do I run to see errors like this?

cfieber:
I'm not sure, it didn't fail in the travis build, just failing on my local machine when trying to locally maven publish (`gw -PenablePublishing=true publishToMavenLocal`)

Maybe its my openjdk version, maybe it is linux, maybe the build gods are just running out of ways to smite me

PR #281, opened 2019-05-03
fix(ssl): update SSL configuration for boot2

cfieber:
this fixes configuring SSL on any additional ports (only really applicable for gate)


kork-web/src/main/groovy/com/netflix/spinnaker/config/TomcatConfiguration.groovy
```diff
@@ -86,9 +87,8 @@ class TomcatConfiguration {
         apiConnector.port = tomcatConfigurationProperties.apiPort
 
         def ssl = tomcatContainerCustomizerUtil.copySslConfigurationWithClientAuth(tomcat)
-        Http11NioProtocol handler = apiConnector.getProtocolHandler() as Http11NioProtocol
-        // TODO(rz): no more:
-        tomcat.configureSsl(handler, ssl)
+        def sslCustomizer = new SslConnectorCustomizer(ssl, tomcat.getSslStoreProvider())
```
cfieber: dirty (but no more so that our accessing configureSsl in boot 1) since SslConnectorCustomizer is package private.

someone set up us the bomb for boot 2.2 upgrade most likely...

PR #283, opened 2019-05-06
fix(requests): Correctly compare header names

marchello2000:
Fixes a bug resulting from https://github.com/spinnaker/kork/pull/270
where X-SPINNAKER-* headers stopped being propagated because they were
not compared in a case insensitive way


marchello2000:
@ajordens: I am going to merge it so I can get a release of clouddriver going for Monday, let me know if you have feedback later

ajordens:
Looks alright to me.

PR #289, opened 2019-05-10
feat(tomcat): allow relaxed URI support

cfieber:
newer tomcat as part of the upgrade to boot2 is stricter about
which characters are allowed in a URI path or query, but does
support relaxed mode for those characters.

this exposes two additional configuration options:
```yaml
default:
  relaxedPathCharacters: "[]"
  relaxedQueryCharacters: ${default.relaxedPathCharacters}
```

These strings consist of characters that are allowable in URIs
even if they don't strictly conform to the RFC.

An example we found was the square bracket characters were
previously acceptable, and were used in pipeline configuration
names. This setting allows us to preserve that capability
until we can migrate callers to properly URLEncode paths and
queries


cfieber:
I'm not so hot to trot on the `default` as the config namespace for these properties but it is where the rest of our tomcat customization properties are already so...

PR #291, opened 2019-05-14
feat(serviceSelector): Added a byLocationServiceLocator

jeyrschabu:
- assists in finding a service by region/location


kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/SelectableService.java
```diff
@@ -81,5 +87,39 @@ public String getExecutionId() {
     public String getOrigin() {
       return origin;
     }
+
+    public String getLocation() {
+      return location;
+    }
+
+    public Criteria withApplication(String application) {
```
jeyrschabu: Added these builder style setters because I suspect the default constructor might get out of control as we add more selectors.

PR #293, opened 2019-05-16
upgrade to boot 2.1.6

robfletcher:


spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -31,7 +31,7 @@ ext {
     retrofit2     : "2.5.0",
     spectator     : "0.75.0",
     spek          : "1.1.5",
-    springBoot    : "2.1.4.RELEASE",
+    springBoot    : "2.1.5.RELEASE",
```
cfieber: I had made this a property because I thought this was versioned in lock-step with the spring-boot version:
https://search.maven.org/artifact/org.springframework.security.oauth.boot/spring-security-oauth2-autoconfigure/2.1.4.RELEASE/jar

I don't currently see a 2.1.5.RELEASE for it so maybe we should unwind my assumption or wait for a release of the oauth2-autoconfigure jar

cfieber: feels like that was maybe a bad assumption:

https://docs.spring.io/spring-security-oauth2-boot/docs/current/reference/htmlsingle/

looks like something we should be striving to get rid of

cfieber: I have a ticket open to port our internal config to oauth2 - I will deal with that as part of that work

robfletcher:
Updated this PR now Boot 2.1.6 was released. I aligned versions of Jackson and Groovy. Guava was unchanged.

spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -32,7 +32,7 @@ ext {
     retrofit2     : "2.6.0",
     spectator     : "0.75.0",
     spek          : "1.1.5",
-    springBoot    : "2.1.4.RELEASE", //check groovy and guava versions after updating this
+    springBoot    : "2.1.6.RELEASE",
     springCloud   : "Greenwich.SR1",
```
robfletcher: Not sure if this needs to change. It's not directly referenced from the Boot BOM

PR #297, opened 2019-05-17
feat(exceptions): Adding tiered root-level exceptions

robzienert:
Adds a hierarchy of root-level Spinnaker exception interface types:

```
SpinnakerException      # root-level interface
|
+- UserException        # exceptional conditions caused by an end-user
|
+- IntegrationException # exceptional conditions caused by non-core Spinnaker
|                       # code (cloud providers, extensions, etc)
|
+- SystemException      # exceptional conditions caused by core Spinnaker
                        # (defects, infra/db errors, etc)
```

Some basic `*RuntimeException` types are included which can be extended;
however it would be preferable to avoid direct usage, preferring concrete,
exception types.

The problem I'm trying to solve here is how difficult it is to classify errors into the three buckets (user vs integration vs core). How does this seem?

### DIFF
--- a/kork-exceptions/src/main/java/com/netflix/spinnaker/kork/exceptions/IntegrationException.java
+++ b/kork-exceptions/src/main/java/com/netflix/spinnaker/kork/exceptions/IntegrationException.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.exceptions;
+
+/** An exception thrown by non-core Spinnaker code (e.g. cloud providers, vendor stages, etc.) */
+public class IntegrationException extends SpinnakerException {
+  public IntegrationException(String message) {
+    super(message);
+  }
+
+  public IntegrationException(String message, Throwable cause) {
+    super(message, cause);
+  }
+
+  public IntegrationException(Throwable cause) {
+    super(cause);
+  }
+
+  public IntegrationException(String message, String userMessage) {
+    super(message, userMessage);
+  }
+
+  public IntegrationException(String message, Throwable cause, String userMessage) {
+    super(message, cause, userMessage);
+  }
+
+  public IntegrationException(Throwable cause, String userMessage) {
+    super(cause, userMessage);
+  }
+}
--- a/kork-exceptions/src/main/java/com/netflix/spinnaker/kork/exceptions/SpinnakerException.java
+++ b/kork-exceptions/src/main/java/com/netflix/spinnaker/kork/exceptions/SpinnakerException.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.exceptions;
+
+import javax.annotation.Nullable;
+
+/** A root-level marker interface for all exceptions to be thrown by Spinnaker code. */
+public class SpinnakerException extends RuntimeException implements HasAdditionalAttributes {
+
+  /**
+   * An optional, end-user friendly message.
+   *
+   * <p>In most cases, an exception message is more suitable for developers and operators, but can
+   * be confusing for an end-user. This field provides a standard method of propagating messaging up
+   * to the edge.
+   */
+  @Nullable private String userMessage;
+
+  /**
+   * Whether or not the exception is explicitly known to be retryable.
+   *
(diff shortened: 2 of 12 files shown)


kork-exceptions/src/main/java/com/netflix/spinnaker/kork/exceptions/SpinnakerException.java
```diff
@@ -0,0 +1,29 @@
...
+
+/** A root-level marker interface for all exceptions to be thrown by Spinnaker code. */
+public interface SpinnakerException {
+
+  /**
+   * Allows an optional user-friendly message, separate from the standard message.
+   */
+  default @Nullable String getUserMessage() {
+    return null;
+  }
```
robzienert: 🤔 not sure if...? `getMessage()` would then be used for developer- & operator-friendly messages.

kork-exceptions/src/main/java/com/netflix/spinnaker/kork/exceptions/IntegrationException.java
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.exceptions;
+
+/** An exception thrown by non-core Spinnaker code (e.g. cloud providers, vendor stages, etc.) */
+public interface IntegrationException extends SpinnakerException {}
```
marchello2000: what's the point of having an interface extend another interface and not add any value add?

kork-exceptions/src/main/java/com/netflix/spinnaker/kork/exceptions/SpinnakerException.java
marchello2000: how would one set the `userMessage`?
Also, maybe worthwhile adding an `isRetryable`?

marchello2000:
I think it makes sense. This would be very useful if we used something to log all our exceptions and then we could have a goal to be, say, `SystemException` clean.

One thing I did in auth header filter, was to have a more sensible stack trace which filters to frames that are in our code base/package (https://github.com/spinnaker/kork/blob/master/kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java#L91)
I think it makes parsing the error message MUCH MUCH easier. but obviously sometimes having the full stack can be useful for system issues. We could have both (where both are logged)

kork-exceptions/src/main/java/com/netflix/spinnaker/kork/exceptions/SpinnakerException.java
robzienert: I'm thinking the `userMessage` would be set on the constructor, similar to how exceptions work today with other fields.

kork-exceptions/src/main/java/com/netflix/spinnaker/kork/exceptions/SpinnakerException.java
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.exceptions;
+
+import javax.annotation.Nullable;
+
+/** A root-level marker interface for all exceptions to be thrown by Spinnaker code. */
+public interface SpinnakerException {
```
dreynaud: when I encounter a `SomethingException` class, I expect it to be an `Exception`. Why not make this directly a subclass of `Exception` or `RuntimeException`? It means you also can't `catch(SpinnakerException e)`

emjburns:
I think I like this 👍

dreynaud:
real strong endorsement there @emjburns

kork-exceptions/src/main/java/com/netflix/spinnaker/kork/exceptions/UserException.java
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.exceptions;
+
+/** An exception thrown when the cause is a user (e.g. bad input). */
+public interface UserException extends SpinnakerException {}
```
dreynaud: I understand the intent but _user_ seems like an oddly charged word. I don't necessarily have a better suggestion off the top of my head... maybe `RequestException`, `InputException` or `ClientException`?

dreynaud:
@robzienert what about something like a `SpinnakerValidationException`? There are a few places where we create exceptions out of thin air that are "expected" under certain conditions. Things like traffic guard and preconditions check

robzienert:
I've added another `ConstraintViolationException` which extends `SpinnakerException`. 👍

kork-exceptions/src/main/java/com/netflix/spinnaker/kork/exceptions/UserException.java
robzienert: `UserlandException`?

dreynaud: make that `LalalandException` and I'm in

ajordens:
I like the idea of trying to classify the types of errors we're seeing. 

> Would a next step be to take the existing exception types in kork-exceptions and have them extend these new types?

Worth noting that the variants in `kork-exceptions` support passing additional data back across the wire, might be something to consider here.

Would we consider try/catching around obvious boundaries and attempt to re-throw the `SpinnakerException` variant (ie. we know an "integration" is running so re-throw everything as `IntegrationException`?)

As for metrics, would we track counts for exception types at the service edges (ie. whatever we're going to serialize back out)?

robzienert:
Good call out on additional data. I think I’ll pull some inspiration from `kork-exceptions` and add some additional metadata, a `retryable` boolean field really sticks out to me to propagate across service boundaries.

> Would we consider try/catching around obvious boundaries and attempt to re-throw

In my mind, yes. Catching something like `RetrofitError` or `Amazon*Exception` and then re-casting into the correct `SpinnakerException` would be great, imo.

> As for metrics, would we track counts for exception types at the service edges (ie. whatever we're going to serialize back out)?

I think so. Having a Spring exception handler that tracks these would be handy; as well as an error handler in RunTaskHandler, for example? Any opinions here?

PR #303, opened 2019-05-27
feat(selector): Adding v2 service selector

jeyrschabu:
- supports flexible selection of a service based on parameters
- allows per service visibility and access to config attributes

### DIFF
--- a/kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/v2/SelectableService.java
+++ b/kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/v2/SelectableService.java
@@ -1,0 +1,375 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.kork.web.selector.v2;
+
+import static java.util.stream.Collectors.toList;
+import static java.util.stream.Collectors.toMap;
+
+import java.util.*;
+import java.util.function.Function;
+
+/**
+ * Provides support for sharding a service based on {@link Parameter}. Example of a service
+ * definition:
+ *
+ * <pre>
+ * bakery:
+ *  baseUrl: http://bakery.com
+ *  altBakeryApiEnabled: false
+ *  timeoutMs: 3000
+ *  baseUrls:
+ *  - baseUrl: http://alt-bakery.us-west-x.com
+ *    priority: 1
+ *    config:
+ *      altBakeryApiEnabled: true
+ *      maxItemsCached: 10
+ *    parameters:
+ *    - name: region
+ *      values:
+ *      - us-west-1
+ *      - us-west-2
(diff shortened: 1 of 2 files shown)


jeyrschabu:
Example service configuration:

```
bakery:
  baseUrl: http://bakery-unstable.test.netflix.net:7001
  roscoApisEnabled: false
  baseUrls:
  - baseUrl: http://rosco-us-west-x.test.netflix.net:7001
    priority: 2
    config:
      roscoApisEnabled: true
    parameters:
    - name: region
      values:
        - us-west-1
        - us-west-2

  - baseUrl: http://rosco-us-west-2.test.netflix.net:7001
    priority: 2
    config:
      roscoApisEnabled: true
    parameters:
    - name: region
      values:
        - us-west-2
```

Regex on values is also supported. This class can also be used to shard services based on authenticated user, cloudProvider, application or execution type.

kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/v2/SelectableService.java
```diff
@@ -0,0 +1,371 @@
...
+ * // configuration {@link BaseUrl#config} attributes for each service definition are accessible like this
+ * if (bakery.config["altBakeryApiEnabled"]) {
+ *     // Do interesting things
+ * }
+ *
+ * assert bakery.baseUrl == "http://alt-bakery.other.com"
+ * assert bakery.config == mapOf("timeoutMs" to 3000, "altBakeryApiEnabled" to true)
+ * }</pre>
+ *
+ * @param <T> the type of service to be selected
```
robzienert: LOOK AT THIS DOCUMENTATION. Jeyrs is a mad man. I love it. 😍😍😍😍😍

kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/v2/SelectableService.java
```diff
@@ -0,0 +1,371 @@
...
+ *        baseUrl: http://bakery.com
+ *        altBakeryApiEnabled: false
+ *        timeoutMs: 3000
+ *        baseUrls:
+ *        - baseUrl: http://alt-bakery.us-west-x.com
+ *          priority: 1
+ *          config:
+ *            altBakeryApiEnabled: true
+ *            maxItemsCached: 10
+ *        parameters:
```
robzienert: This indentation is incorrect. Thanks, YAML.

kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/v2/SelectableService.java
```diff
@@ -0,0 +1,371 @@
...
+ *              - centOS
+ *          - name: user
+ *            values:
+ *              - regx:^[a-f].+@netflix.com
+ *  }</pre>
+ *
+ * Usage:
+ *
+ * <pre>{@code
+ * val config = mapOf("timeoutMs", 3000)
```
robzienert: I'd like to see this config typed... is that something we can reasonably do? It'll be harder to do inspections with a config map...

```
data class BakeryConfig(val timeout: Duration, val altApiEnabled: Boolean)

if(bakery.altApiEnabled) {
  // do things...
}
```

kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/v2/SelectableService.java
jeyrschabu: Thanks YAML indeed

kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/v2/SelectableService.java
jeyrschabu: I'd love that as well. Doing so right now will introduce more complexity around how the global config gets applied to each shard definition's config. Would like to avoid having to resort to using reflection for this at this point. My plan is to begin using the v2 selector for the bakery initially, improve it, then move all our other services.

PR #305, opened 2019-05-31
Stop Retrofit coroutines hanging when exceptions occur

robfletcher:


robfletcher:
See square/okhttp#5151

ajordens:
Any concern with the existing `propagate()` method ... it's doing basically the same thing with the supplied closures.

robfletcher:
Sounds like I should fix that too.

cfieber:
The whole reason I did the try / catch / throw RuntimeException is that having checked exceptions on the method signature makes using this thing pretty garbage-y

could we lombok `@SneakyThrow` Exception on here instead?

cfieber:
and if we are going to do that, I'd rather just throw the underlying exception out of here, and fix the spot where we need to try / catch / throw IOException in the interceptor to work around the kotlin coroutine thing

robfletcher:
@cfieber ultimately we can probably go back to `RuntimeException` once square/okhttp#5151 is fixed.

@ajordens are you talking about this `propagate` method: https://github.com/spinnaker/kork/blob/1bcb1cfad4df0ef18a3be18afe9995067df74127/kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequest.java#L109-L146? I don't see it doing anything like that.

PR #308, opened 2019-06-03
fix(rpc): exception handling on interceptors

cfieber:
Changes AuthenticatedRequest to lombok SneakyThrow instead of catching
and rethrowing RuntimeException to ensure we don't actually wrap and
lose an IO Exception (this is preferrable when used in an interceptor
in okhttp).

Adds defensive catch Exception and rethrow as IOException on OkHttp
interceptors.

closes #305


kork-security/kork-security.gradle
```diff
@@ -16,4 +16,8 @@ dependencies {
   testImplementation "org.spockframework:spock-core"
   testRuntimeOnly "cglib:cglib-nodep"
   testRuntimeOnly "org.objenesis:objenesis"
+
+  compileOnly "org.projectlombok:lombok"
+  annotationProcessor(platform(project(":spinnaker-dependencies")))
```
robfletcher: We should probably just stick this in the root `build.gradle` to avoid having it anywhere Lombok is used.

kork-web/src/main/java/com/netflix/spinnaker/okhttp/OkHttp3MetricsInterceptor.java
```diff
@@ -28,6 +28,12 @@ public OkHttp3MetricsInterceptor(Registry registry) {
 
   @Override
   public Response intercept(Chain chain) throws IOException {
-    return (Response) doIntercept(chain);
+    try {
+      return (Response) doIntercept(chain);
+    } catch (IOException ioe) {
+      throw ioe;
+    } catch (Exception ex) {
+      throw new IOException(ex);
```
robfletcher: Do we want to change this to a `RuntimeException` once the underlying OkHttp issue is fixed?

cfieber: I think we could just remove the try/catch once the underlying issue is resolved, this just makes sure for now we don't accidentally throw something else in our metrics interceptors

kork-security/kork-security.gradle
cfieber: its still only 5 modules out of the whole project so I dunno what the tipping point is but yeah at some point that would make sense

cfieber: I kind of like having to consciously decide that you are going to lombok a module

robfletcher: Yeah, I was just thinking of adding the `platform`. You'd still need to specify the dependency.

cfieber: hah, if I stick this in the root build.gradle I'm going to have to do some sort of shenanigans to exclude the spinnaker-dependencies project

cfieber: think I'll do a gradle/lombok.gradle instead that does all the dependency bits for lombok and make the lombok modules to an `apply from: "gradle/lombok.gradle"`

PR #310, opened 2019-06-03
perf(redis): Optimize payloadSize

ezimanyi:
Instead of builing a long string and getting the payload size of that string, just compute the size of each string in the list and return the sum.  This significantly reduces allocations during caching cycles for the kubernetes provider, which often calls mset with a long list of parameters.


cfieber:
there is a similar implementation for Map<String, String> that could use the same love

ezimanyi:
Good point, #312

PR #314, opened 2019-06-06
feat(headers): add ability to skip auth header checks

marchello2000:
For local development, it is tedious to see all the "anonymous" request warnings.
Add ability to disable the check via config variable, e.g:

```yaml
spring:
  profiles: local

okHttpClient:
  interceptor:
    skipHeaderCheck: true
```


ajordens:
Might be nice to have this nested under `okHttpClient` rather than a new highlevel `interceptor`?

marchello2000:
> Might be nice to have this nested under `okHttpClient` rather than a new highlevel `interceptor`?

yeah, i was thinking the same but then wasn't sure if that would be confusing. will update

PR #316, opened 2019-06-07
fix(compat): adds backwards compatible constructors for interceptors

cfieber:
the fiat api needs the old signature, going to put this in as well as getting fiat-api consumers bumped to a newer fiat-api


cfieber:
I pushed an update to everything using fiat-api already, just going to close this

PR #328, opened 2019-07-01
feat(keel): add basic resource

emjburns:
Adding the basic keel resource structure (no type info, just lovely maps) to kork.

### DIFF
--- a/kork-manageddelivery/src/main/java/com/netflix/spinnaker/kork/manageddelivery/model/Resource.java
+++ b/kork-manageddelivery/src/main/java/com/netflix/spinnaker/kork/manageddelivery/model/Resource.java
@@ -1,0 +1,36 @@
+/*
+ *
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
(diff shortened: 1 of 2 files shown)


kork-manageddelivery/src/main/java/com/netflix/spinnaker/kork/manageddelivery/model/Resource.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.util.Map;
+import lombok.Data;
+
+/**
+ * The container for declaratively managed resources.
+ *
+ * <p>[apiVersion] which plugin and version of that plugin handles this resource [kind] what kind of
+ * resource this is [metadata] information about the resource, including name, uid, and
+ * serviceAccount [spec] the actual resource
```
erikmunson: this paragraph reads like each bracketed (`[]`) bit should be on its own line — was that the intent?

emjburns: Apparently spotless doesn't like my intent and I didn't notice. Arg.

PR #332, opened 2019-07-05
feat(configuration): Add kork-config module for Spring Cloud Config Server integration

scottfrederick:
This PR mostly contains code extracted from clouddriver that will be required by other services for integration with Spring Cloud Config Server. It will also reduce some duplicated Spring Boot default properties.


scottfrederick:
@ezimanyi @ethanfrogers feedback on this move?

PR #334, opened 2019-07-08
chore(dependencies): jedis to 3.x, new dyno-jedis

asher:
redis.clients:jedis has been provided as a transitive dependency of dyno-jedis. This PR updates dyno-jedis for compatibility with dyno-queues which requires upgrading redis.clients:jedis from 2.9.3 to 3.0.1.


asher:
This will require minor code changes to most services utilizing `kork-jedis` due to jedis 2.9.x -> 3.x changes.

asher:
Gate (uses `spring-session-data-redis`) and Front50 (uses `spring-data-redis`) will need to pinned back to Jedis 2.9.x until the Spring Boot 2 modules have Jedis 3.x compatible versions, or the dependencies are removed. Added a `kork-jedis2-test` module so that these services can continue using a kork provided `EmbeddedRedis` while pinned back.

PR #338, opened 2019-07-10
fix(config): fix config server bootstrap location

ethanfrogers:
when deployed by halyard, all config lands in `/opt/spinnaker/config` so we need to set the config server bootstrap location to include `/opt/spinnaker/config` so that it gets picked up.

### DIFF
--- a/kork-config/src/main/java/com/netflix/spinnaker/kork/configserver/ConfigServerBootstrap.java
+++ b/kork-config/src/main/java/com/netflix/spinnaker/kork/configserver/ConfigServerBootstrap.java
@@ -23,7 +23,7 @@ public class ConfigServerBootstrap {
     // of https://github.com/spring-cloud/spring-cloud-commons/issues/466
     defaultSystemProperty(
         "spring.cloud.bootstrap.location",
-        "classpath:/,classpath:/config/,file:./,file:./config/,${user.home}/.spinnaker/");
+        "classpath:/,classpath:/config/,file:./,file:./config/,/opt/spinnaker/config/,${user.home}/.spinnaker/");
     defaultSystemProperty(
         "spring.cloud.bootstrap.name", "spinnakerconfig,${spring.application.name}config");
     defaultSystemProperty("spring.cloud.config.server.bootstrap", "true");


ethanfrogers:
cc @scottfrederick

ezimanyi:
The way we do this for `spring.config.additional-location` is to add have the Debian packaging logic add this as a parameter to the Java process `-Dspring.config.additional-location=/opt/spinnaker/config/` [here](https://github.com/spinnaker/spinnaker-gradle-project/blob/master/spinnaker-dev-plugin/src/main/groovy/com/netflix/spinnaker/gradle/application/SpinnakerApplicationPlugin.groovy#L25).  One advantage of that is that knowledge of install directory is isolated to the packaging code (and you could imagine building another package type that puts this in a different location).  Maybe we want to do the same here?  Also interested in whether @cfieber has thoughts.

ethanfrogers:
Ah, yep! I knew it was probably hiding somewhere that I wasn't aware of. I'm happy to refactor in favor of something more inline with the way we do everything else. LMK.

ethanfrogers:
@ezimanyi @cfieber unfortunately, until [this PR](https://github.com/spring-cloud/spring-cloud-commons/pull/468) is merged we can't do something like `spring.cloud.boostrap.additional-location` which I think is ultimately the way we do this. Maybe this is a hack but what if we drop that system property into the packaging scripts and and append it to `spring.cloud.bootstrap.location` in the code in this PR?

ezimanyi:
Oh, I see---so currently `spring.cloud.bootstrap.location` behaves like boot2 `spring.config.location` and replaces everything currently set (rather than adding to it)?  And there's no `spring.cloud.bootstrap.additional-location`?

Yeah in that case it's probably safer to just add the `/opt/config` path as you have done here unless there's some better way of doing it than I'm missing.

ethanfrogers:
The only other way I can think of doing it is something like this:

```
if (System.getProperty("spring.cloud.bootstrap.additional-location") {
}
```

ethanfrogers:
@ezimanyi what do you think about something like this? This seems a bit safer to me, though probably more obscure. 
```
String defaultBootstrapLocation = "classpath:/,classpath:/config/,file:./,file:./config/";
    if (System.getProperty("spring.cloud.bootstrap.additonal-location") != null) {
      defaultBootstrapLocation += "," + System.getProperty("spring.config.bootstrap.additional-location") + ",";
    }
    defaultBootstrapLocation += ",${user.home}/.spinnaker/";

    defaultSystemProperty(
        "spring.cloud.bootstrap.location",
        defaultBootstrapLocation);
```

ezimanyi:
I'm not sure I see the benefit to the suggestion in the latest comment?  I think that given there's no `spring.cloud.bootstrap.additional-location`, the code currently on the PR is probably the best solution for now so would be fine going with that.

PR #340, opened 2019-07-12
fix(sql): improved testcontainer support

asher:


kork-sql-test/src/main/java/com/netflix/spinnaker/kork/sql/test/SqlTestUtil.java
```diff
@@ -48,14 +49,38 @@ public static TestDatabase initPreviousDatabase() {
...
+    // host, port, and db name are ignored with the jdbcUrl method of TC initialization and
+    // overridden to "test" by the driver.
+    return initDatabase(
+        "jdbc:tc:mysql:5.7.22://somehostname:someport/somedb", SQLDialect.MYSQL_5_7);
+  }
+
+  public static TestDatabase initPreviousTcMysqlDatabase() {
+    MySQLContainer container =
+        new MySQLContainer("mysql:5.7.22")
+            .withDatabaseName("previous")
```
asher: It appears that testcontainers lacks out of the box support for utilizing 2 logical db's within the same container due to how it creates the user grants. Running a current and previous database (such as with orca's `SqlExecutionRepositorySpec`) thus requires running two containers.

bsideup: Anything we can improve in Testcontainers in a generic way? :)

Also, have you tried using the root user to create multiple databases per running instance?

asher: @bsideup I didn't see mention of root user access in docs and only tried it w/o a password, but looks like it defaults to "test"? I'll revisit this next week, would be nice to avoid running a second container instance on the travis-ci vm's :) Thanks for the tip!

I'm in the process of adding a redis cluster implementation to our queueing library (https://github.com/spinnaker/keiko) and am interested in creating a redis cluster testcontainer that runs several shards. I'll let you if I have any feedback after I go thru the setup process.

bsideup: Sorry that it is undocumented 😊 MySQL does support root user access, see the following block of code that describes the "rules" of assigning the password of it:
https://github.com/testcontainers/testcontainers-java/blob/d681aeef6f0d2987b6126dbc7382d6eb83fb7f3b/modules/mysql/src/main/java/org/testcontainers/containers/MySQLContainer.java#L48

For the cluster setup, you may want to use the networks support. Take a look at this test for the inspiration:
https://github.com/testcontainers/testcontainers-java/blob/master/core/src/test/java/org/testcontainers/containers/NetworkTest.java

PR #342, opened 2019-07-16
feat(sql): use a single container for dual db tests

asher:


kork-sql-test/src/main/java/com/netflix/spinnaker/kork/sql/test/SqlTestUtil.java
```diff
@@ -51,25 +55,47 @@ public static TestDatabase initPreviousDatabase() {
   public static TestDatabase initTcMysqlDatabase() {
     // host, port, and db name are ignored with the jdbcUrl method of TC initialization and
     // overridden to "test" by the driver.
-    return initDatabase(
-        "jdbc:tc:mysql:5.7.22://somehostname:someport/somedb", SQLDialect.MYSQL_5_7);
+    return initDatabase(tcJdbcUrl, SQLDialect.MYSQL_5_7);
   }
 
-  public static TestDatabase initPreviousTcMysqlDatabase() {
```
asher: I'm going to re-add this method before merging, so that the resulting kork release remains backwards compatible with current orca-sql tests.

PR #344, opened 2019-07-22
chore(plugins): adding beta tag

link108:
Adding beta tag to signify plugins are a beta feature


kork-plugins/kork-plugins.gradle
```diff
@@ -7,6 +7,8 @@ dependencies {
   api project(":kork-exceptions")
 
   api "org.springframework.boot:spring-boot-starter-web"
+  implementation("com.google.guava:guava:28.0-jre")
+
```
robzienert: Remove the version - we want to use the BOM-provided version.

PR #353, opened 2019-08-09
chore(dependencies): Upgrade Guava

plumpy:
In Clouddriver, we have a transitive dependency on 26.0-android, which
is overriding the version we set here anyway.

I talked to @dreynaud and the issue that prompted
spinnaker/clouddriver#3692 was actually that IntelliJ was compiling
against Guava 27.1, but then using Guava 19.0 on the classpath. Then it
was getting an error about a missing method because it was trying to
call a method added in Guava 20. This error was specific to IntelliJ and
wouldn't have happened if Gradle was used (which will ensure that the
same version is used at compilation and runtime).

So I think this should be a safe thing to do.


plumpy:
cc @cfieber (I don't have permission to ask for a review in kork for some reason)

PR #357, opened 2019-08-14
feat(plugins): Download remote plugins

link108:
This PR is related to spinnaker/spinnaker#4181.

Different environments used to run Spinnaker have varying constraints, in order to ensure plugins can be used in all environments, we added the ability to download plugins via Kork at runtime. By default, downloading of plugin jars will be disabled, but if the Spinnaker administrator trusts the download location, they can enable plugin jar downloading. Spinnaker users (application owners) will not have the ability to add plugins.

To ensure Spinnaker is running as expected, a Spinnaker service will fail to start if it fails to download the plugin JARs.


link108:
fixing tests

PR #360, opened 2019-08-15
fix(dependencies): module cleanup / dependency scoping

cfieber:
This refactors a few things to clean up the dependency graph of kork consumers:

- relocates AwsComponents out of kork-core and into kork-aws
- moves a few things out of api scope in kork core including eureka and some
  spectator implementation details. This most notably results in a bunch of
  older libraries that were present due to the eureka transitive dependency
  being gone in consumer classpaths.
- makes kork-exceptions an api scoped dependency of kork-core and updates
  exception classes in kork-core to extend from appropriate subclasses
  of SpinnakerException

This is a breaking change to kork consumers that were relying on the
transitive dependency graph of the affected modules, but I've preflighted
this against all the Spinnaker repos and fixed the broken imports where
needed. Given the result of that effort, I'm pretty okay with inflicting
this on anyone - in all cases the fix was to either explicitly add a
dependency (best practice anyway, if you are coding against it you
generally shouldn't be getting it transitively) or fixing an accidental
import (you probably didn't mean to use StringUtils from the aws-sdk
when commons-lang3 was also on your classpath).


cfieber:
examples of what this uncovered:

https://github.com/spinnaker/rosco/pull/411
https://github.com/spinnaker/orca/pull/3060
https://github.com/spinnaker/keel/pull/365
https://github.com/spinnaker/igor/pull/486
https://github.com/spinnaker/gate/pull/863
https://github.com/spinnaker/echo/pull/610
https://github.com/spinnaker/front50/pull/561
https://github.com/spinnaker/clouddriver/pull/3949
https://github.com/spinnaker/clouddriver/pull/3899

kork-core/src/main/java/com/netflix/spinnaker/kork/archaius/ArchaiusConfiguration.java
```diff
@@ -108,7 +109,7 @@ private void initPropertyBindings() {
         try {
           return Optional.of(new ResourcePropertySource(name, r));
         } catch (IOException ioe) {
-          throw new RuntimeException("Error loading property source [" + name + "]: " + res, ioe);
+          throw new SystemException("Error loading property source [" + name + "]: " + res, ioe);
```
robzienert: 🙏 thankyou!

PR #364, opened 2019-08-22
fix(config): Make config server auto-config less fragile

scottfrederick:
Under some conditions, `ConfigFileService` would not get auto-configured correctly when Spring Cloud Config Server integration was enabled due to ordering of dependent auto-configuration classes. This change makes the auto-configuration work more reliably.


scottfrederick:
superseded by https://github.com/spinnaker/kork/pull/370

PR #365, opened 2019-08-22
bumped Kotlin, Spring Boot, TestContainers, Retrofit2, JUnit

robfletcher:


robfletcher:
Lots of things Rob, but how many releases do you want?

PR #366, opened 2019-08-22
feat(exceptions): Add standard exception summary model for service boundaries

robzienert:
From https://github.com/spinnaker/clouddriver/pull/3972

### DIFF
--- a/kork-exceptions/src/main/java/com/netflix/spinnaker/kork/exceptions/SpinnakerException.java
+++ b/kork-exceptions/src/main/java/com/netflix/spinnaker/kork/exceptions/SpinnakerException.java
@@ -16,8 +16,10 @@
 package com.netflix.spinnaker.kork.exceptions;
 
 import javax.annotation.Nullable;
+import lombok.Getter;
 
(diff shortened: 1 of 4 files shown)


kork-exceptions/src/main/java/com/netflix/spinnaker/kork/exceptions/SpinnakerException.java
```diff
@@ -16,8 +16,10 @@
 package com.netflix.spinnaker.kork.exceptions;
 
 import javax.annotation.Nullable;
+import lombok.Getter;
 
 /** A root-level marker interface for all exceptions to be thrown by Spinnaker code. */
+@Getter
```
marchello2000: i think we also need `@Setter` otherwise how would someone set the `retryable` flag?
(e..g https://github.com/spinnaker/clouddriver/blob/master/clouddriver-titus/src/main/groovy/com/netflix/spinnaker/clouddriver/titus/exceptions/InsufficientDeploySourceStateException.java#L34)

marchello2000: @robzienert

PR #370, opened 2019-08-29
feat(config): Resolve configfiles using a custom PropertySource

scottfrederick:
The Spring Cloud Config Server integration provided by `kork-config` includes the ability to specify account credentials (e.g. Kubernetes account `kubeconfigfile` and Google account `json-path`) as files that are stored in a Config Server backend like git, using a `configserver:` qualifier. 

Previously code in services that accessed these file references in account properties needed to use a `ConfigFileService` bean provided by `kork-config` to resolve the file references, pulling the specified file from the Config Server backend if necessary. 

This was starting to get unwieldy and intrusive in services, especially in cases where the code using the account properties wasn't implemented as Spring beans that could get a `ConfigFileService` bean injected into it. This PR adds a Spring `PropertySource` wrapper that automatically de-references the `configserver:` file reference when a property is resolved. The result is that services will always get a absolute local file reference. 

`ConfigFileService` remains as a useful encapsulation of methods for dealing with local files.

### DIFF
--- a/kork-config/kork-config.gradle
+++ b/kork-config/kork-config.gradle
@@ -7,6 +7,8 @@ dependencies {
   implementation "commons-io:commons-io"
   implementation "org.apache.commons:commons-lang3"
 
+  api project(':kork-secrets')
+
   api "org.springframework.cloud:spring-cloud-context"
   api "org.springframework.cloud:spring-cloud-config-server"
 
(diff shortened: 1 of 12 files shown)


scottfrederick:
@ethanfrogers @ncknt If you're testing the config server integration in clouddriver using e.g. `kubeconfigFile: configserver:kubeconfig.yml`, this change should be transparent. I've tested with my k8s config, but any additional testing and review would be appreciated.

PR #371, opened 2019-08-29
chore(aws): Bump aws sdk

clareliguori:
Pick up suspend/resume APIs for Application Auto Scaling

Related to spinnaker/spinnaker#4707

### DIFF
--- a/spinnaker-dependencies/spinnaker-dependencies.gradle
+++ b/spinnaker-dependencies/spinnaker-dependencies.gradle
@@ -20,7 +20,7 @@ publishing {
 
 ext {
   versions = [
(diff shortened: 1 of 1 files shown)


emjburns:
FYI we use the netflix/awsobjectmapper and we need to also bump the version there: https://github.com/Netflix/AWSObjectMapper/blob/master/build.gradle

This is not something I would expect you to do/know of since it's kind of a weird thing we use. I can take care of bumping it on Monday!

emjburns:
https://github.com/Netflix/AWSObjectMapper/pull/65

emjburns:
LGTM

PR #372, opened 2019-08-30
feat(secrets): Support SAML metadata file as a secret

KathrynLewis:
Corresponding Halyard PR: https://github.com/spinnaker/halyard/pull/1411

There have been some requests to support the SAML metadata.xml file as a secret. Secrets in the services are decrypted at startup, so this change makes sure the metadata file also gets the "file:" prefix required for SAML files.

@plumpy PTAL?


kork-secrets/src/main/java/com/netflix/spinnaker/kork/secrets/SecretAwarePropertySource.java
```diff
@@ -47,6 +47,14 @@ public Object getProperty(String name) {
     return o;
   }
 
+  private boolean isFile(String name) {
+    return name.endsWith("file") || name.endsWith("path") || name.endsWith("truststore");
+  }
+
+  private boolean isSamlFile(String name) {
+    return name.endsWith("keystore") || name.endsWith("metadataurl");
```
cfieber: was this supposed to be `metadata.xml`?

KathrynLewis: These are the names of the properties in the profile yamls, so in this case the name is "metadataurl" in the saml block of `gate.yml`

KathrynLewis: The idea here is if your `gate.yml` has something like this:
```
saml:
  enabled: true
  issuerId: spinnaker.okta
  metadataUrl: encrypted:s3!....
  keyStore: encrypted:s3!....
  ...
```
it knows it has to decrypt and prepend with "file:".

KathrynLewis:
> I'm not super keen on this - feels like the wrong place / wrong way to handle special cases like this
> 
> can we add a more generic file extension to identify items that should be treated as secret files? (`.secret`?)
> 
> The saml config could then reference `metadata.xml.secret` or something..

kork-secrets/src/main/java/com/netflix/spinnaker/kork/secrets/SecretAwarePropertySource.java
cfieber: so I understand correctly, the `encrypted:s3!...` bit says the actual value for this lives in an s3 bucket, and the SecretAwarePropertySource will fetch it from that bucket (into a local temp file I guess?), so the property value then has to become `file:///path/to/that/file` ?

cfieber: just wondering if there is a way to add that to the encrypted value spec (e.g. `encrypted:s3:file!...` as an indicator that the expectation is to pull remote content to a local file vs into an in-memory value) rather than adding a bunch of special cases to the secret aware property source

KathrynLewis:
We agree this is kind of a goofy way to handle the secret file decryption, but it seemed to work the best in terms of reading in files early in the Spring life cycle, consistently for all services, and minimizing interruption to the normal flow for users not using secrets. 

We could explore a better way to identify files as secret, but the logic here is working with the property name fields in the config files, and changing those would require massive changes across the board to both printing the profile yamls and reading them in at service startup.

cfieber:
After your comment in the thread, I am not proposing changing property names but I'm interested in encoding the fact of whether a secret is a remote file fetch resolved into a local spring Resource (file:) property in the encryption syntax ( https://github.com/spinnaker/kork/pull/372#discussion_r327244536 )

What are your thoughts on that approach?

cfieber:
essentially being able to replace the `isFile` method with `EncryptedSecret.isFile` and `isSamlFile` with `EncryptedSecret.isFileReference` (or something, naming is hard)

KathrynLewis:
I've been looking at what it would take to make that change. It would be a few changes in halyard and kork, but I'm realizing that that sort of goofy logic will be there no matter what. A summary of what would be changed:

If this were new, I think that wouldn't be too hard and would require changes in only a few places to add the tags to the encrypted syntax and the simple `isFile()`, etc, methods.

But this has been a feature for several months now and in order to make it backward compatible, we'd also have to add some changes to Halyard. There, objects are serialized to yaml for the service profiles through `DecryptingObjectMapper` where we'd need to add something like this:
```  
// if this field is a secret file but users are using the old method without the "file:true" tag
if (hasSecretFileAnnotation && !EncryptedSecret.hasFileTag) {
    if (isSamlFile && !EncryptedSecret.hasSamlFileTag) {
        // add saml file tag
    } else {
       // add file tag
    }
}
```

Now the one-off logic for deciding if a property is a file/saml file would just be in Halyard. I can't find a good way of avoiding having the logic live somewhere. 

If done this way, users would have to upgrade Halyard and Spinnaker simultaneously, otherwise their secrets configuration would break, unless we want to keep the existing logic in for a while. Also, if they have any service yamls versioned in source control, any without the updated syntax would fail.

I'll keep trying to think of better solutions, but these are the issues I'm hitting currently.

KathrynLewis:
@cfieber  I'm sorry, I'm not seeing a way around having this type of logic in place somewhere. People have been using this feature for a while now and have their secrets configured the way its documented now. If we add a tag for secret files and remove this logic, it will break spinnaker for anyone storing a file as a secret.

KathrynLewis:
Instead of adding another tag, which can be a bit user-error-prone as they are, we're thinking of changing the syntax to be `encryptedFile:s3!...` and `encrypted:s3!...`.

PR #373, opened 2019-09-03
feat(test): Adding Mimicker data generator lib

robzienert:
While working in clouddriver, I've noticed there's a tremendous amount of static values being used. This can be problematic for a number of reasons, but worst of all, it means input assumptions are being codified into tests, which may lead to brittleness of the code under test.

This PR introduces `Mimic`, a Spinnaker-specific data generator for property testing. It's not very complete at this point, but wanted to get something out there for people to see and interact with, filling out everything as it's needed.

```kotlin

Mimic().run {
  moniker().get().app // flhadksjfha
  aws().region // us-west-2
  aws().availabilityZone // us-west-2b
  // ...
}
```

This WIP is loosely based on [jFairy](https://github.com/Devskiller/jfairy), which I had used at a prior company for doing similar work. I opted to NIHS rather than use jFairy because it comes with a heap of dependencies and would force on us a lot of data producers that we would never use.

From this PR, we can start writing/rewriting tests to utilize Mimic. Down the road, we could use this as a foundation for generative validation testing should we feel inspired to do so.


jonsie:
Should be called `mimicker`.

kork-test/src/main/java/com/netflix/spinnaker/kork/test/mimic/DataContainer.java
```diff
@@ -0,0 +1,98 @@
...
+
+  public DataContainer() {}
+
+  public DataContainer(List<String> resourcePaths) {
+    for (String path : resourcePaths) {
+      load(path);
+    }
+  }
+
+  public DataContainer load(String resourcePath) {
```
jonsie: This is a cool feature, but should probably have a unit test.

kork-test/src/main/java/com/netflix/spinnaker/kork/test/mimic/Mimic.java
```diff
@@ -0,0 +1,58 @@
...
+  public Mimic() {
+    this(new DataContainer());
+  }
+
+  public Mimic(DataContainer dataContainer) {
+    this.dataContainer = dataContainer;
+  }
+
+  public TextProducer text() {
+    return new TextProducer(dataContainer, random());
```
jonsie: `return new TextProducer(dataContainer.load("mimic.yml"), random());`

kork-test/src/main/java/com/netflix/spinnaker/kork/test/mimicker/DataContainer.java
```diff
@@ -0,0 +1,112 @@
...
+  private final Yaml yaml = new Yaml();
+
+  private final Map<String, Object> data = new HashMap<>();
+
+  public DataContainer() {
+    this(Collections.emptyList());
+  }
+
+  public DataContainer(List<String> resourcePaths) {
+    load("mimicker.yml");
```
cfieber: might be nice to make this a method instead `withDefaultResources`, and use that in the no-args constructor but let the explicitly specified constructor not inherit that opinion by default

PR #374, opened 2019-09-04
feat(proto): Adds kork-proto module to support autogenerating protobu…

ttomsu:
…fs. The first one is for the stats module.

This came from the discussion on https://github.com/spinnaker/echo/pull/640

### DIFF
--- a/kork-proto/kork-proto.gradle
+++ b/kork-proto/kork-proto.gradle
@@ -1,0 +1,23 @@
+plugins {
+  id "java-library"
+  id "com.google.protobuf" version "0.8.10"
+}
+
+dependencies {
+  api(platform(project(":spinnaker-dependencies")))
+
+  implementation "com.google.protobuf:protobuf-java"
+}
+
+protobuf {
+  protoc {
+    // TODO: figure out how to reference version in BOM.
+    artifact = "com.google.protobuf:protoc:3.9.1"
+  }
+}
+
+idea {
+  module {
+    generatedSourceDirs += file("${protobuf.generatedFilesBaseDir}/main/java")
+  }
+}
--- a/kork-proto/src/main/proto/stats/log.proto
+++ b/kork-proto/src/main/proto/stats/log.proto
@@ -1,0 +1,85 @@
+syntax = "proto3";
+
+package io.spinnaker.proto.stats;
+
+option go_package = "github.com/spinnaker/kork/kork-proto/proto/stats";
+
+option java_package = "com.netflix.spinnaker.kork.proto.stats";
+option java_multiple_files = true;
+
+message Event {
+  SpinnakerInstance spinnaker_instance = 1;
+}
+
+message SpinnakerInstance {
+  string id = 1;
+  string version = 2;
+  Application application = 3;
+}
+
+message Application {
+  string id = 1;
+  Execution execution = 2;
+}
+
+message Execution {
+  string id = 1;
+  Type type = 2;
+
+  enum Type {
+    UNKNOWN = 0;
+    PIPELINE = 1;
+    ORCHESTRATION = 2;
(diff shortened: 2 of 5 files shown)


kork-proto/kork-proto.gradle
```diff
@@ -0,0 +1,23 @@
...
+}
+
+protobuf {
+  protoc {
+    // TODO: figure out how to reference version in BOM.
+    artifact = "com.google.protobuf:protoc:3.9.1"
+  }
+}
+
+idea {
```
plumpy: FYI, we don't really use these generated idea files anymore since I checked in `.idea` directories to all the repositories. But I can figure out how to add this information to the checked-in files in a follow-up PR

And I suppose I need to update the instructions to tell people not to run `./gradlew idea` anymore.

kork-proto/src/main/proto/stats/log.proto
```diff
@@ -0,0 +1,85 @@
+syntax = "proto3";
+
+package io.spinnaker.proto.stats;
+
+option go_package = "github.com/spinnaker/kork/kork-proto/proto/stats";
+
+option java_package = "com.netflix.spinnaker.kork.proto.stats";
+option java_multiple_files = true;
+
+message Event {
```
ajordens: Assuming we came to a decision that we wanted proto definitions in `kork`? 

Reading the `echo` PR seemed to indicate the definitions would still live in `stats`?

plumpy: Travis and I kinda decided it was okay for this one to live in `kork` since `stats` isn't a Spinnaker microservice, isn't a Java project, doesn't use gradle, etc.

I still think the proto files for each service should live within the service, but I think in this case (and maybe just for now, while the `stats` service is still developing) it would be okay to put it in `kork`. WDYT?

kork-proto/kork-proto.gradle
ttomsu: This is needed to add the generated Java files as a source set, and therefore, to be included in the final build/JAR. It's not relate to `./gradlew idea`, as I also use the `.idea` directory-based approach.

kork-proto/src/main/proto/stats/log.proto
ttomsu: The `stats` service is also not a GRPC service (yet?), so in this case the proto is only establishing the message payload. 

I'm not against amending/appending the capability, though.

ajordens: You answered the question I cared about ... "proto files for each service should live in the service". That's where I'm at right now (although what do I know about this). 

As for this one-off, seems alright and can be always been moved around as other parts evolve.

ttomsu: I'm not against adding it when Cloud Run supports it: https://cloud.google.com/run/docs/issues#grpc_websocket

kork-proto/kork-proto.gradle
cfieber: the idea plugin does interop with intellij project import - I think this change actually works as intended

you can't use any of the idea plugin XML manipulation bits, but anything that touches attributes on the model should get picked up 

see https://docs.gradle.org/current/userguide/idea_plugin.html#sec:idea_configuration

> ... customize the IDEA project and module files that it generates. These take the form of both model properties and lower-level mechanisms that modify the generated files directly. For example, you can add source and resource directories, as well as inject your own fragments of XML. The former type of configuration is honored by IDEA’s import facility, whereas the latter is not.

plumpy: Weird, but okay. Seems like it would be better to modify the underlying model directly, rather than going through the idea plugin to do it...

I should probably actually remove the idea plugin from all the projects that have `.idea` directories, which would break this.

PR #376, opened 2019-09-05
fix(core): Cache service version

robzienert:
There's really no need to re-resolve after we've done it once. Also removing a log message that says a version can't be resolved: It gets noisy.

### DIFF
--- a/kork-core/src/main/java/com/netflix/spinnaker/kork/version/ServiceVersion.java
+++ b/kork-core/src/main/java/com/netflix/spinnaker/kork/version/ServiceVersion.java
@@ -18,22 +18,23 @@ package com.netflix.spinnaker.kork.version;
 import java.util.List;
 import javax.annotation.Nonnull;
 import lombok.extern.slf4j.Slf4j;
-import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.ApplicationContext;
-import org.springframework.context.ApplicationContextAware;
 
 /** Attempts to resolve the version of a service. */
 @Slf4j
-public class ServiceVersion implements ApplicationContextAware {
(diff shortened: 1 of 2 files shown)


kork-core/src/main/java/com/netflix/spinnaker/kork/version/ServiceVersion.java
```diff
@@ -32,6 +32,8 @@
   private final List<VersionResolver> resolvers;
   private ApplicationContext applicationContext;
 
+  private String resolvedServiceVersion;
+
   @Autowired
   public ServiceVersion(List<VersionResolver> resolvers) {
```
cfieber: you could add ApplicationContext here and have that autowired and get rid of the ApplicationContextAware and the null check on applicationContext inside resolve.
could also call resolve on construct in that case as well

kork-core/src/main/java/com/netflix/spinnaker/kork/version/ServiceVersion.java
```diff
@@ -50,22 +52,28 @@ public String resolve() {
...
+          }
+        } catch (Exception e) {
+          log.error("Failed while resolving version: {}", resolverName, e);
         }
-      } catch (Exception e) {
-        log.error("Failed while resolving version: {}", resolverName, e);
+      }
+
+      if (resolvedServiceVersion == null) {
+        return UNKNOWN_VERSION;
```
cfieber: is there a reasonable expectation that if the resolvers don't find a result, that it is worth continuing to check? (should we cache the UNKNOWN_VERSION result?)

robzienert: Fair - unlikely there'd be a situation that we'd end up resolving something previously unknown.

PR #380, opened 2019-09-09
feat(secrets/gcs): Added GcsSecretEngine for decrypting secrets in GCS

german-muzquiz:
New `GcsSecretEngine` for decrypting secrets of the form `encrypted:gcs!b:bucket!f:file[!k:key-in-file]`.
Some of the code is copied from how Halyard downloads Spinnaker BOM's from GCS.


bpowell:
@cfieber mind taking a look at this please?

kork-secrets-gcp/src/main/java/com/netflix/spinnaker/kork/secrets/engines/GoogleCredentials.java
```diff
@@ -0,0 +1,60 @@
...
+import com.google.api.client.http.HttpRequestInitializer;
+import com.google.api.client.http.HttpTransport;
+import com.google.api.client.util.ExponentialBackOff;
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+import java.util.concurrent.TimeUnit;
+
+public class GoogleCredentials {
+
+  private static final long CONNECT_TIMEOUT = TimeUnit.MINUTES.toMillis(2);
```
ezimanyi: Since these timeouts are needed as `int` everywhere they are used, I'd suggest just doing the cast directly here:
`private static final int CONNECT_TIMEOUT = (int) TimeUnit.MINUTES.toMillis(2);`

Also, 2 min seems like a pretty long connection timeout...how did you pick that value?

kork-secrets-gcp/src/main/java/com/netflix/spinnaker/kork/secrets/engines/GcsSecretEngine.java
```diff
@@ -0,0 +1,130 @@
...
+  private Storage applicationDefaultGoogleStorage;
+  private Storage unauthenticatedGoogleStorage;
+
+  public GcsSecretEngine() {
+    applicationDefaultGoogleStorage = createGoogleStorage(true);
+    unauthenticatedGoogleStorage = createGoogleStorage(false);
+  }
+
+  public String identifier() {
+    return GcsSecretEngine.IDENTIFIER;
```
ezimanyi: Super minor, but I'd probably just call this `IDENTIFIER` from within the class rather than use a qualified reference.

kork-secrets-gcp/src/main/java/com/netflix/spinnaker/kork/secrets/engines/GcsSecretEngine.java
```diff
@@ -0,0 +1,130 @@
...
+    String objName = encryptedSecret.getParams().get(STORAGE_FILE_URI);
+
+    ByteArrayOutputStream output = new ByteArrayOutputStream();
+    log.info("Getting contents of object {} from bucket {}", objName, bucket);
+
+    try {
+      applicationDefaultGoogleStorage
+          .objects()
+          .get(bucket, objName)
+          .executeMediaAndDownloadTo(output);
```
ezimanyi: There's no need to download this to an `OutputStream` then convert to an `InputStream`. Instead you can just call `executeMediaAsInputStream()` to directly generate in `InputStream` and return that.

kork-secrets-gcp/src/main/java/com/netflix/spinnaker/kork/secrets/engines/GcsSecretEngine.java
```diff
@@ -0,0 +1,130 @@
...
+              : new GoogleCredential();
+      if (credential.createScopedRequired()) {
+        credential =
+            credential.createScoped(Collections.singleton(StorageScopes.DEVSTORAGE_READ_ONLY));
+      }
+      requestInitializer = GoogleCredentials.setHttpTimeout(credential);
+
+      log.info("Loaded credentials for reading GCS secrets.");
+    } catch (Exception e) {
+      requestInitializer = GoogleCredentials.retryRequestInitializer();
```
ezimanyi: The above block is doing a lot of things, and we're catching a generic `Exception`.  How do we know that the exception is because no application default credential could be loaded?  (In the case where `useApplicationDefaultCreds = false` we're not even trying to load one so the error message can't be correct there.)

Also, why are we setting the `requestInitializer` to `GoogleCredentials.retryRequestInitializer()` only in this exception block?

kork-secrets-gcp/src/main/java/com/netflix/spinnaker/kork/secrets/engines/GcsSecretEngine.java
```diff
@@ -0,0 +1,130 @@
...
+@Component
+@Slf4j
+public class GcsSecretEngine extends AbstractStorageSecretEngine {
+  private static String IDENTIFIER = "gcs";
+
+  private Storage applicationDefaultGoogleStorage;
+  private Storage unauthenticatedGoogleStorage;
+
+  public GcsSecretEngine() {
+    applicationDefaultGoogleStorage = createGoogleStorage(true);
```
ezimanyi: I'm not sure I understand why we need both an authenticated and an unauthenticated `Storage` object here.  Are there cases where we'd expect an authenticated request to fail, but an unauthenticated one to pass?

If this is just to handle the case where we fail on construction of the authenticated `Storage` could we instead just handle that in the constructor, and have a single `Storage` object that's authenticated if default credentials are available and unauthenticated otherwise.  That would also remove the logic that tries fetching every URL twice.

kork-secrets-gcp/src/main/java/com/netflix/spinnaker/kork/secrets/engines/GoogleCredentials.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  public static HttpRequestInitializer setHttpTimeout(
+      final HttpRequestInitializer requestInitializer) {
+    return request -> {
+      requestInitializer.initialize(request);
+      request.setConnectTimeout((int) CONNECT_TIMEOUT);
+      request.setReadTimeout((int) READ_TIMEOUT);
+      request.setUnsuccessfulResponseHandler(
+          new HttpBackOffUnsuccessfulResponseHandler(new ExponentialBackOff()));
```
ezimanyi: Are you sure that we want to retry with exponential backoff on any non-200 response?  What are the backoff parameters created by `ExponentialBackoff`?  I suspect that if a file is missing (or we don't have permission) it will be a better experience to fail fast rather than retry exponentially for a while then fail.

kork-secrets-gcp/src/main/java/com/netflix/spinnaker/kork/secrets/engines/GoogleCredentials.java
```diff
@@ -0,0 +1,60 @@
...
+  public static HttpRequestInitializer retryRequestInitializer() {
+    return request -> {
+      request.setConnectTimeout((int) CONNECT_TIMEOUT);
+      request.setReadTimeout((int) READ_TIMEOUT);
+      request.setUnsuccessfulResponseHandler(
+          new HttpBackOffUnsuccessfulResponseHandler(new ExponentialBackOff()));
+    };
+  }
+
+  public static HttpRequestInitializer setHttpTimeout(
```
ezimanyi: This might be better named as `getRequestInitializer` as it's doing more then setting an HTTP timeout.

kork-secrets-gcp/src/main/java/com/netflix/spinnaker/kork/secrets/engines/GoogleCredentials.java
german-muzquiz: I grabbed this code from how Halyard reads BOM's stored in GCS. I'll decrease the timeout to 20 seconds if that sounds better.

kork-secrets-gcp/src/main/java/com/netflix/spinnaker/kork/secrets/engines/GcsSecretEngine.java
german-muzquiz: I think these are misnamed methods, both `GoogleCredentials.setHttpTimeout` and `GoogleCredentials.retryRequestInitializer` set timeouts and an `ExponentialBackOff` unsuccessful response handler.

kork-secrets-gcp/src/main/java/com/netflix/spinnaker/kork/secrets/engines/GoogleCredentials.java
german-muzquiz: It's not immediately obvious but default behavior of `HttpBackOffUnsuccessfulResponseHandler` is to retry only on 5xx errors. I'll set it explicitly for clarity.

kork-secrets-gcp/src/main/java/com/netflix/spinnaker/kork/secrets/engines/GoogleCredentials.java
ezimanyi: I think 20 seconds sounds more reasonable for a connection timeout, and is more in line with other places in Spinnaker.

kork-secrets-gcp/src/main/java/com/netflix/spinnaker/kork/secrets/engines/GoogleHttpUtils.java
```diff
@@ -0,0 +1,54 @@
...
+import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
+import com.google.api.client.http.HttpBackOffUnsuccessfulResponseHandler;
+import com.google.api.client.http.HttpRequestInitializer;
+import com.google.api.client.http.HttpTransport;
+import com.google.api.client.util.ExponentialBackOff;
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+import java.util.concurrent.TimeUnit;
+
+public class GoogleHttpUtils {
```
ezimanyi: I know this isn't something we often do currently in the Spinnaker codebase, but I think this is a good candidate for being package-private, as presumably no other classes outside `GcsSecretEngine` would need to know about it.

PR #383, opened 2019-09-17
feat(core): Add resilience4j boot2 annotations api

robzienert:
Adds a nice Spring integration & annotations to the global project's dependency list: [Documentation](https://resilience4j.readme.io/docs/getting-started-3)

I was thinking we'd use `{service}-web/config/application.yml` to document all resilience configs within a service. It'd be an interesting exercise to additionally maintain any of Netflix's resilience4j config in OSS, keeping all configs in a `netflixResilience` spring profile.

### DIFF
--- a/kork-sql/src/test/kotlin/com/netflix/spinnaker/kork/sql/SpringStartupTests.kt
+++ b/kork-sql/src/test/kotlin/com/netflix/spinnaker/kork/sql/SpringStartupTests.kt
@@ -36,7 +36,6 @@ import strikt.assertions.isEqualTo
 @RunWith(SpringRunner::class)
 @SpringBootTest(
   classes = [StartupTestApp::class],
-  webEnvironment = SpringBootTest.WebEnvironment.NONE,
   properties = [
     "sql.enabled=true",
     "sql.migration.jdbcUrl=jdbc:h2:mem:test",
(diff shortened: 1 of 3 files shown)


kork-sql/src/test/kotlin/com/netflix/spinnaker/kork/sql/SpringStartupTests.kt
```diff
@@ -36,7 +36,7 @@ import strikt.assertions.isEqualTo
 @RunWith(SpringRunner::class)
 @SpringBootTest(
   classes = [StartupTestApp::class],
-  webEnvironment = SpringBootTest.WebEnvironment.NONE,
+//  webEnvironment = SpringBootTest.WebEnvironment.NONE,
```
cfieber: Can delete if this is unnecessary

spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -116,6 +116,7 @@ dependencies {
     api("de.danielbechler:java-object-diff:0.95")
     api("de.huxhorn.sulky:de.huxhorn.sulky.ulid:8.2.0")
     api("dev.minutest:minutest:1.7.0")
+    api("io.github.resilience4j:resilience4j-annotations:${versions.resilience4j}")
```
cfieber: Be nice when they fix their bom publishing to switch to that

https://github.com/resilience4j/resilience4j/issues/310

robzienert:
@ethanfrogers Yes, Hystrix isn't really used here at Netflix and doesn't get much love. We also have `RetrySupport` in kork here that's... not super flexible. This library has a few more features that we can use, has less dependencies than Hystrix, and seems to be the direction the greater Spring community is headed. Should be nice!

PR #385, opened 2019-09-23
feat(core): Add PublicApi runtime annotation

robzienert:


luispollo:
Good idea! Do we plan to retroactively annotate APIs with this?

robzienert:
@luispollo Yes, as we figure out what we actually want "exposed" as a public API.

PR #388, opened 2019-10-08
feat(secrets): add encryptedFile syntax for secret files

KathrynLewis:
This replaces the less-than-ideal logic around determining if an encrypted secret is a saml file, a regular file or string value. Now, users should use the `encryptedFile:s3!...` syntax for secret files and `encrypted:s3!...` for secret strings. The logic for prepending saml files with "file:" after decrypting moves into gate with this PR: https://github.com/spinnaker/gate/pull/919

This proposes to keep both sets of logic in place for version 1.16/1.17, but mark the old logic as deprecated and remove it fully in version 1.17/1.18?


KathrynLewis:
@cfieber these PRs would replace this one: https://github.com/spinnaker/kork/pull/372

KathrynLewis:
@cfieber do you have some time to take a look at these two PRs?

PR #390, opened 2019-10-16
chore(gcs): Update credentials configuration to latest style and lazy…

duftler:
… load instead of loading in constructor.


duftler:
@german-muzquiz Please review.

PR #391, opened 2019-10-17
feat(config-doc): Generate configuration metadata via spring-boot-configuration-processor

jonsie:
This is a small example of what I have in mind regarding how to generate up-to-date configuration and configuration metadata at build time for Spinnaker services and libraries.

Running `./gradlew kork-aws:build` generates a Spring configuration metadata file.  In this example it looks like:

```
{
  "groups": [
    {
      "name": "bastion",
      "type": "com.netflix.spinnaker.kork.aws.bastion.BastionProperties",
      "sourceType": "com.netflix.spinnaker.kork.aws.bastion.BastionProperties"
    }
  ],
  "properties": [
    {
      "name": "bastion.account-iam-role",
      "type": "java.lang.String",
      "description": "AWS IAM role to assume.",
      "sourceType": "com.netflix.spinnaker.kork.aws.bastion.BastionProperties"
    },
    {
      "name": "bastion.enabled",
      "type": "java.lang.Boolean",
      "sourceType": "com.netflix.spinnaker.kork.aws.bastion.BastionProperties"
    },
    {
      "name": "bastion.host",
      "type": "java.lang.String",
      "sourceType": "com.netflix.spinnaker.kork.aws.bastion.BastionProperties"
    },
    {
      "name": "bastion.port",
      "type": "java.lang.Integer",
      "sourceType": "com.netflix.spinnaker.kork.aws.bastion.BastionProperties"
    },
    {
      "name": "bastion.proxy-cluster",
      "type": "java.lang.String",
      "sourceType": "com.netflix.spinnaker.kork.aws.bastion.BastionProperties"
    },
    {
      "name": "bastion.proxy-region",
      "type": "java.lang.String",
      "sourceType": "com.netflix.spinnaker.kork.aws.bastion.BastionProperties"
    },
    {
      "name": "bastion.user",
      "type": "java.lang.String",
      "sourceType": "com.netflix.spinnaker.kork.aws.bastion.BastionProperties"
    }
  ],
  "hints": []
}
```

If there was a default value set for any of these properties, that would be included in the document in the `defaultValue` key.

Note the `description` for `bastion.account-iam-role` - this is Javadoc.

I think this pattern could be broadly applied to all of Spinnaker (given some refactors) and result in up-to-date configuration examples as well as documentation.  It could possibly even be used to keep Halyard configs up-to-date.

**Caveats:**

- This pattern does not work with Groovy - any classes annotated with `@ConfigurationProperties` would need to be migrated to Java or Kotlin, which brings me to my second point...
- There is a bug with Kotlin's Java annotation processing library `kapt` wherein it does not process default values for class member variables for metadata (see https://github.com/spring-projects/spring-boot/issues/15397 and https://youtrack.jetbrains.com/issue/KT-30164). So, the `defaultValue` key would not exist for Kotlin classes annotated with `@ConfigurationProperties` until someone fixes that bug.

Thoughts?


kork-aws/kork-aws.gradle
```diff
@@ -16,9 +16,15 @@
 apply plugin: "java-library"
 apply plugin: "groovy"
 
+//TODO(csmalley): Move away from sshoogr
```
robzienert: All of your drinks are on me at the Summit if you do this.

robzienert:
For other readers https://www.baeldung.com/spring-boot-configuration-metadata

It's not clear to me (the page linked appears incorrect on this matter) is whether or not this supports validation? I'd love to have that included in this manifest: Do you know if this is something we'd need to bring ourselves?

jonsie:
Regarding `@Value` annotations, it would be my preference that we scrub them from the codebase and use `@ConfigurationProperties` classes.  Unless there's a use for them in some way I'm not thinking about - definitely possible.

Regarding `DynamicConfigService`, I'm not concerned with supporting those configs at this moment.  The nature of `DynamicConfigService` is one such that it can not exist and the application can still function.  I am more interested in covering core configuration and documentation for Spinnaker in a systematic way.  I like your proposal `@Value` annotations in place of `DynamicConfigService`, but I'm not addressing that here.

Regarding `@ConditionalOn*` annotations, I'd have to think about it.  Not sure right now.

Of course tooling would need to be built to make this information useful, but I'm thinking of that as a separate scope of work.  Off the top of my head you could use this to produce example configuration files, documentation for configuration, Halyard configurations for services, and probably more.  This is really just about surfacing the metadata.

kork-aws/kork-aws.gradle
jonsie: 🥂

cfieber:
I'm not sure how I feel about `@Value` - I'd rather not have to have the context refresh to pick up changes. 

I don't mind this pattern at all (autowire DynamicConfigService into the ConfigurationProperties object) and it would give us support for documenting those properties for free:

cfieber:
WRT `@ConditionalOnProperty/Expression` - I suspect there's a few things to consider there:

1. The are probably a smell that something should eventually become a plugin instead
2. We should be able to get documentation on those properties by ensuring they are present on an appropriate `@ConfigurationProperties` object (and ideally we can move away from `ConditionalOnExpression` to `ConditionalOnProperty` wherever possible)

kork-aws/kork-aws.gradle
jonsie: Be careful what you wish for.

😆 🍻

robzienert: No regrets.

robzienert:
RE wiring `DynamicConfigService` into a configuration properties class... interesting. I can get behind this - does the processor pickup any getter/setters or does it operate off of properties? 🤔 

Totally onboard with not doing the tooling as part of this, I just wanted an idea of where you were headed here. I'd be happy to see this merged.

robzienert:
I would greatly appreciate the work you're doing around SSH bastion stuff to be a different PR, though. It has nothing to do with the original PR work.

jonsie:
Yah that's cool, I can do that.  One moment.

jonsie:
I added back the Groovy.

PR #392, opened 2019-10-17
fix(retry): deprecate RetrySupport method with long in favor of Duration

dreynaud:
This should help prevent issues where we mix up maxRetries and retryBackoff


robzienert:
How about removing RetrySupport usage and using resilience4j?

dreynaud:
@robzienert maybe? In the mean time it's probably useful to at least have the deprecation warning on this one if it is a known pitfall

robzienert:
Yes. 👍

PR #397, opened 2019-10-21
fix(*): Move conf annotation processor to kork-core

robzienert:


jonsie:
That should do it.

kork-core/kork-core.gradle
```diff
@@ -5,7 +5,9 @@ apply from: "$rootDir/gradle/lombok.gradle"
 dependencies {
   api(platform(project(":spinnaker-dependencies")))
   implementation project(":kork-exceptions")
+  annotationProcessor(platform(project(":spinnaker-dependencies")))
```
cfieber: I suspect we want this in the top level build.gradle on all subprojects

jonsie: I believe most of the services are already configured this way, for example:

https://github.com/spinnaker/orca/blob/master/build.gradle#L63

jonsie: But maybe you're referring to just kork - that was my initial inclination too.  Not sure if just `kork-core` will cover everything.

cfieber: sorry I actually meant the actual annotationProcessor dependency on spring-boot-configuration-processor could land at the top level on all subprojects. The annotationProcessor platform dependency is probably already there maybe for lombok

PR #398, opened 2019-10-21
feat(expressions): Add support for spel functions provided by plugins

robzienert:
- About the simplest plugin interface I could think of. This just changes the expressions code to be aware of `PluginManager` and makes the `ExpressionFunctionProvider` interface extend `ExtensionPoint`.
- I also moved the `ExpressionFunctionProvider` into the correct source set (`src/main/kotlin` instead of `src/main/java`) and added some docs around the interface.
- spotless shifted a bunch of stuff around, so the changeset is larger than what one would expect.

There will be a subsequent PR in Orca that will wire things up.

Example extension point implementation that would get picked up by Orca:

```java
@Extension
public class HelloExpressionFunctionProvider implements ExpressionFunctionProvider {
  @NotNull
  @Override
  public Functions getFunctions() {
    return new Functions(
      new FunctionDefinition("hello")
    );
  }

  @Nullable
  @Override
  public String getNamespace() {
    return null;
  }

  public static String hello() {
    return "hello";
  }
}
```


kork-expressions/src/main/kotlin/com/netflix/spinnaker/kork/expressions/ExpressionFunctionProvider.kt
```diff
@@ -0,0 +1,104 @@
...
+ *  }
+ * ```
+ *
+ * The above function provider could then be called in a SpEL expression:
+ *
+ * ```
+ * ${netflix.hello(this)} // `this` is an Execution, if running in Orca.
+ * ```
+ */
+interface ExpressionFunctionProvider : ExtensionPoint {
```
ajordens: Thoughts on `@Beta` or `@Alpha` marking extension points when they're first created? 
(I realize this was pre-existing so perhaps not needed in this case)

kork-expressions/src/main/kotlin/com/netflix/spinnaker/kork/expressions/ExpressionFunctionProvider.kt
```diff
@@ -0,0 +1,104 @@
...
+ *    @JvmStatic
+ *    fun hello(execution: Execution): String =
+ *      "Hello, ${execution.id}"
+ *  }
+ * ```
+ *
+ * The above function provider could then be called in a SpEL expression:
+ *
+ * ```
+ * ${netflix.hello(this)} // `this` is an Execution, if running in Orca.
```
christopherthielen: ```// `this` is an Execution, if running in Orca.```
🤯

kork-expressions/src/main/kotlin/com/netflix/spinnaker/kork/expressions/ExpressionFunctionProvider.kt
robzienert: Yes, my intention is to add the `@Beta` tag. I'm happy to introduce our own `@Alpha` / `@Beta` annotations into `kork-core` if we standardize on something like that.

kork-expressions/src/main/kotlin/com/netflix/spinnaker/kork/expressions/ExpressionFunctionProvider.kt
robzienert: Yeah. I'm going to remove that. No sense in documenting some fanciness that is Orca-specific in kork.

PR #399, opened 2019-10-22
refactor(plugins): Embed enabled flag into PluginManager

robzienert:


kork-plugins/src/main/kotlin/com/netflix/spinnaker/config/PluginsAutoConfiguration.kt
```diff
@@ -29,22 +28,21 @@ import java.nio.file.Paths
...
+    SpinnakerPluginManager(properties.enabled, Paths.get(properties.rootPath))
 
   @Bean
   fun extensionsInjector(
     pluginManager: SpinnakerPluginManager,
-    registry: GenericApplicationContext
+    context: GenericApplicationContext
   ): ExtensionsInjector =
-    ExtensionsInjector(pluginManager, registry)
+    ExtensionsInjector(pluginManager, context)
```
jonsie: 🦅

PR #401, opened 2019-10-22
feat(exceptions): chain style accessor for SpinnakerException retryable

cfieber:
Allows for `throw new SpinnakerException("some constructor args").setRetryable(true)`


cfieber:
minor quality of life improvement for non-Kotlin users

PR #402, opened 2019-10-22
feat(plugins): Spring-backed plugin status provider

robzienert:
This will likely need to change in the future to use Front50 as the ultimate source of these values, but this is an incremental improvement over default PF4J.

### DIFF
--- a/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpringPluginStatusProvider.kt
+++ b/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpringPluginStatusProvider.kt
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.plugins
+
+import org.pf4j.PluginStatusProvider
+import org.slf4j.LoggerFactory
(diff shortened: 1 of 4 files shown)


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpringPluginStatusProvider.kt
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+
+  override fun onApplicationEvent(event: ApplicationEnvironmentPreparedEvent) {
+    log.debug("Adding ${this.javaClass.simpleName} as new PropertySource")
+    event.environment.propertySources.addFirst(propertySource)
+  }
+
+  private fun isEnabled(pluginId: String): Boolean =
+    propertySource.getProperty(enabledPropertyName(pluginId))?.toString()?.toBoolean() ?: false
```
cfieber: I feel like this should read from the environment instead of specifically from this property source.

This would allow you to configure enabled plugins in the service yaml/environment/system properties as defaults, and then this just serves as more of a runtime override if something gets disabled later.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpringPluginStatusProvider.kt
```diff
@@ -0,0 +1,66 @@
...
+ */
+class SpringPluginStatusProvider : PluginStatusProvider, ApplicationListener<ApplicationEnvironmentPreparedEvent> {
+
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  private val propertySourceBackingStore: MutableMap<String, Any> = mutableMapOf()
+  private val propertySource = MapPropertySource("plugins", propertySourceBackingStore)
+
+  override fun disablePlugin(pluginId: String) {
+    if (isEnabled(pluginId)) {
```
cfieber: if isEnabled ends up reading from the environment then this may not actually reflect the state of this property source (assuming maybe a fast property source or some other source slides in ahead of this one) so maybe its just worth always updating the MapPropertySource

(same for enablePlugin)

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpringPluginStatusProvider.kt
robzienert: Oh, derp. Yeah. I originally was going through the environment for reads. I don't know what my brain did here to remove that.

PR #404, opened 2019-10-24
fix(config): support configserver files

ncknt:
This enables support for `configserver:` prefixed files transparently for services already using ConfigFileService.

It fixes the issue in Clouddriver of changes not being detected when referencing `kubeconfigFile` with `configserver` from a Cloud config server loaded config.

### DIFF
--- a/kork-config/src/main/java/com/netflix/spinnaker/kork/configserver/ConfigFileService.java
+++ b/kork-config/src/main/java/com/netflix/spinnaker/kork/configserver/ConfigFileService.java
@@ -30,10 +30,17 @@ import org.apache.commons.lang3.StringUtils;
 @Slf4j
 public class ConfigFileService {
   private static final String CLASSPATH_FILE_PREFIX = "classpath:";
+  private final CloudConfigResourceService cloudConfigResourceService;
 
-  public ConfigFileService() {}
+  public ConfigFileService(CloudConfigResourceService cloudConfigResourceService) {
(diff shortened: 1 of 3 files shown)


ncknt:
@scottfrederick This fixes an issue with `getKubeconfigFile` (master and 1.16): https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/security/KubernetesCredentialFactory.java#L45

Referencing kubeconfigfile with `configserver` works the first time but not on `CloudConfigRefreshScheduler`

scottfrederick:
@ncknt This is interesting. `CloudConfigAwarePropertySource.java` should be handling all this resolution of `configserver:` files to local files automatically, but there is probably something else that needs to happen there on refresh. 

This change looks like a reasonable workaround.

ncknt:
Yeah it does not on refresh. Spring config actually builds a map of differences and make this a non `CloudConfigAwarePropertySource` property source. Therefore the kubeconfig value comes on as `configserver:xxxx` instead of the temp file. A more complete fix would reinsert the `CloudConfigAwarePropertySource` not just on Applicationconfigured but also on ApplicationRefreshed.

scottfrederick:
@ncknt Makes sense. Sounds like we should apply this fix now and look at the potential "more complete fix" later.

ethanfrogers:
@spinnakerbot add-label ready-to-merge

PR #405, opened 2019-10-24
feat(annotations): Creating kork-annotations

robzienert:
Removed `PublicApi`. It hasn't been used in the project. With PF4J plugins and having an `ExtensionPoint` marker interface, I felt that we didn't need an annotation duplicating its purpose.

While Guava has a `Beta` annotation, there's desire to have as few external core dependencies for in-process plugins as possible. We've decided to also include the `Alpha` annotation in addition to `Beta`, following the software maturity model definitions used internally at Netflix.

Refactored all of Kork to use the `kork-annotations` version of `Beta` over Guava's.


kork-plugins-api/kork-plugins-api.gradle
```diff
@@ -19,6 +19,8 @@ apply from: "$rootDir/gradle/kotlin.gradle"
 
 dependencies {
```
cfieber: do you know where we are inheriting Guava from? would be good to find/kill that so it isn't still transitively in here especially with everything declared as `api` scope

cfieber: or wait, kork-plugins not kork-plugins-api was the one that was currently using `@Beta` so maybe its transitive / explicit in there, I dunno

robzienert: For `orca-api`, it's being explicitly depended on.

As for `kork-plugins-api`, I don't think it has Guava, but it does have a bunch of other deps. I'll be cleaning those up soon (like making Spring-specific plugins API module, instead of forcing all plugins to have Spring even if they don't need it).

PR #410, opened 2019-11-02
feat(expressions): Add plumbing to support function auto-complete

marchello2000:
We would like to add SpEL function auto-complete ability to `deck` for better useability.
This requires that we can describe functions nicely and enum them from orca

* Add function description to `FunctionDefinition`
* Add ability to get all registered `FunctionProviders`

(also minor java doc fixups to remove compiler warnings)

### DIFF
--- a/kork-expressions/src/main/kotlin/com/netflix/spinnaker/kork/expressions/ExpressionFunctionProvider.kt
+++ b/kork-expressions/src/main/kotlin/com/netflix/spinnaker/kork/expressions/ExpressionFunctionProvider.kt
@@ -83,10 +83,15 @@ interface ExpressionFunctionProvider : ExtensionPoint {
    */
   data class FunctionDefinition(
     val name: String,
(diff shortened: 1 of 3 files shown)


kork-expressions/src/main/kotlin/com/netflix/spinnaker/kork/expressions/ExpressionFunctionProvider.kt
```diff
@@ -83,10 +83,15 @@ interface ExpressionFunctionProvider : ExtensionPoint {
    */
   data class FunctionDefinition(
     val name: String,
+    val description: String,
     val parameters: List<FunctionParameter>
   ) {
+    @Deprecated("Please use the overload with description", replaceWith = ReplaceWith("FunctionDefinition(name, \"\", functionParameters)"))
```
robzienert: 🎉

PR #413, opened 2019-11-04
fix(config): Allow native Config Server repos via customization

scottfrederick:
Spring Cloud Config supports a "native" repository type which allows access to config and plain-text files on the local file system. The `kork-config` module doesn't currently support this repository type, since it needs to use the "native" profile as a way to disable config server integration by default. This fix allows an operator to explicitly configure Spinnaker to use the native repo type as a work-around until Spring Cloud Config provides a better way to disable config server with a property.


clanesf:
@ezimanyi - Is this something we can cherry pick to 1.17 or is that a bigger ask because of kork? We're having to maintain a fork of clouddriver until this is in.

ezimanyi:
I think this will be difficult to cherry-pick to 1.17 because, as you say, it affects kork.  But @ttomsu is the release manager so would get the final say on this.

ttomsu:
I would rather not experiment with Kork cherry picks. 1.18.0 will be cut sometime during the week of January 6.

PR #414, opened 2019-11-04
refactor(plugins): Split plugins-api from plugins-spring-api

robzienert:
Makes `kork-plugins-api` ship with minimal dependencies, moving Spring-y plugin capabilities out to a separate module that plugins can optionally use.


gradle/kotlin-test.gradle
```diff
@@ -0,0 +1,47 @@
+/*
```
jonsie: Good call here.

PR #416, opened 2019-11-07
feat(plugins): Support configurable extensions

robzienert:
PF4J has two concepts of extension: System extensions and Plugin-provided extensions. This PR refactors the current codebase to support configuring both concepts in a common way.

An example config might look something like this:

```yaml
spinnaker:
  plugins:
    robzienert.helloworld:
      enabled: true
      extensions:
        robzienert.helloworld-stage1:
          config:
            defaultRecipient: "Kork Reviewers"
        robzienert.helloworld-stage2:
          config:
            defaultRecipient: "Rob"
  extensions:
    robzienert.something-cool:
      enabled: true
      config:
        myConfig: myValue
```

The extension `$namespace.$id` is defined on the extension and is used to resolve config, the same logic (just with a different pointer) to load extension config:

```kt
@SpinnakerExtension(namespace = "robzienert", id = "something-cool")
class SomethingCool : SomeSpinnakerExtensionPoint, ConfigurableExtension<SomethingCool.CoolConfig> {
  lateinit var config: CoolConfig

  fun setConfiguration(config: CoolConfig) {
    this.config = config
  }
}

class CoolConfig(
  val myConfig: String
)
```

A `ConfigResolver` is used to find configs and the default implementation is `SpringEnvironmentExtensionConfigResolver`. This resolver will use the parent service Spring Environment `PropertySources` for all configs. Sourcing from other places, like Front50, would be a matter of overriding the `ConfigResolver` - perhaps this becomes a resolver chain in the future.

I had to refactor quite a lot of things to make this PR a thing, but I've moved stuff around and I feel the codebase is easier to reason about as a result. One departure from stock PF4J is that all extensions are initialized once as singletons.


kork-plugins-api/src/main/java/com/netflix/spinnaker/kork/plugins/ConfigurableExtension.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.plugins;
+
+import com.netflix.spinnaker.kork.annotations.Alpha;
+
+/** Marks an Extension as being capable of accepting a configuration. */
+@Alpha
+public interface ConfigurableExtension {
```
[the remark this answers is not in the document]
cfieber: I also wonder in that case if there is kotlin reflection business to derive the generic type from the extension rather that requiring it on the annotation? (I don't know the reflection stuff in kotlin super well)

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/config/SpringEnvironmentExtensionConfigResolver.kt
```diff
@@ -0,0 +1,100 @@
...
+  private val environment: ConfigurableEnvironment
+) : ConfigResolver {
+
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  private val mapper = ObjectMapper()
+    .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
+    .findAndRegisterModules()
+
+  override fun <T> resolve(coordinates: ConfigCoordinates, expectedType: Class<T>): T {
```
cfieber: what happens if you try and let the spring binding stuff in the environment do the heavy lifting here?

`environment.getProperty("spinnaker.plugins.${coordinates.pluginNamespace}.${coordinates.pluginId}.extensions.${coordinates.extensionNamespace}.${coordinates.extensionId}.config", expectedType)`

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpringExtensionFactory.kt
```diff
@@ -26,20 +30,24 @@ import org.springframework.core.env.StandardEnvironment
...
 ) : ExtensionFactory {
 
   private val log by lazy { LoggerFactory.getLogger(javaClass) }
 
   override fun <T> create(extensionClass: Class<T>): T {
     val extension = createWithoutSpring(extensionClass)
       ?: throw PluginRuntimeException("Failed to create object of extension class: $extensionClass")
-    val pluginWrapper = pluginManager.whichPlugin(extensionClass) ?: return extension
 
+    val anno = getSpinnakerExtensionAnno(extension)
```
jonsie: I prefer to abbreviate `annotation` as `annot`.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpringExtensionFactory.kt
```diff
@@ -50,16 +58,72 @@ open class SpringExtensionFactory(
...
+  private fun getSpinnakerExtensionAnno(extension: Any): SpinnakerExtension =
+    extension.javaClass.getAnnotation(SpinnakerExtension::class.java)
+      ?: throw IntegrationException("Extensions must be defined using @SpinnakerExtension")
+
+  /**
+   * Load an extension config, provided a set of [ConfigCoordinates].
+   */
+  private fun loadConfig(extension: Any, anno: SpinnakerExtension, coordinates: ConfigCoordinates) {
+    if (anno.config.java == MissingConfiguration::class.java) {
+      return
```
christopherthielen: I had assumed that `MissingConfiguration` was an error state based on the name of the class.

kork-plugins-api/src/main/java/com/netflix/spinnaker/kork/plugins/MissingConfiguration.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.plugins;
+
+/**
+ * A default, empty class used as a marker that a {@link SpinnakerExtension} does not have any
+ * configuration.
+ *
+ * @see SpinnakerExtension
+ */
```
srekapalli: EmptyConfiguration or NoConfiguration ... Name looks like an Exception condition or something.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/ExtensionBeanDefinitionRegistryPostProcessor.kt
```diff
@@ -26,28 +27,51 @@ import org.springframework.beans.factory.support.BeanDefinitionRegistryPostProce
...
+      try {
+        log.debug("Creating extension '{}'", it)
+
+        val extensionClass = javaClass.classLoader.loadClass(it)
+        val bean = pluginManager.extensionFactory.create(extensionClass)
+        val beanName = "${extensionClass.simpleName.decapitalize()}SystemExtension"
+
+        beanFactory.registerSingleton(beanName, bean)
+      } catch (e: ClassNotFoundException) {
+        throw IntegrationException("Could not find system extension class '$it'", e)
```
marchello2000: is it the intent that a failure of one extension fails all extensions/service start?
(also, deliberate that other exceptions aren't caught?  I assume that if the extension throws in the .ctor/somewhere on the init path) it would also fail all other extensions from loading

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpringExtensionFactory.kt
```diff
@@ -50,16 +58,72 @@ open class SpringExtensionFactory(
...
     return extension
   }
 
+  /**
+   * Retrieve the [SpinnakerExtension] annotation from an extension.
+   *
+   * Extensions within Spinnaker MUST use the [SpinnakerExtension] annotation, rather than the default PF4J
+   * [org.pf4j.Extension] annotation.
+   */
+  private fun getSpinnakerExtensionAnno(extension: Any): SpinnakerExtension =
```
marchello2000: What @jonsie said `Anno` -> `Annotation`

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/ExtensionBeanDefinitionRegistryPostProcessor.kt
```diff
@@ -26,28 +27,51 @@ import org.springframework.beans.factory.support.BeanDefinitionRegistryPostProce
...
-    // Do nothing.
+    log.debug("Creating system extensions")
+    pluginManager.getExtensionClassNames(null).forEach {
+      try {
+        log.debug("Creating extension '{}'", it)
+
+        val extensionClass = javaClass.classLoader.loadClass(it)
+        val bean = pluginManager.extensionFactory.create(extensionClass)
+        val beanName = "${extensionClass.simpleName.decapitalize()}SystemExtension"
+
```
srekapalli: Will it be useful to publish an Spring ApplicationEvent here , which allow us to possibly do some reporting/telemetry ?

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/config/SpringEnvironmentExtensionConfigResolver.kt
robzienert: Oh haha. Right.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/ExtensionBeanDefinitionRegistryPostProcessor.kt
robzienert: Great thought!

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/ExtensionBeanDefinitionRegistryPostProcessor.kt
robzienert: I'm going to re-arrange the class so the try-catch is specifically surrounding the class loading instead of the whole logic (as it should be).

robzienert:
Updated from feedback:

1. `s/anno/annot`. Also changed the `Anno` abbreviated method to `Annotation`.
1. Removed `config` from `@SpinnakerExtension` and made `ConfigurableExtension<T>`. The `setConfiguration(T)` method needed to be called through reflection, so `SpringExtensionFactory` got a little more gross, but the end-user developer experience will be much more pleasant now. This change also means that `MissingConfiguration` no longer exists.
1. Added an `ExtensionLoaded` Spring `ApplicationEvent` that is emitted for every extension loaded.
1. Refactored `SpringEnvironmentExtensionConfigResolver` to use the `Environment.getProperty(String, Class<?>)` method, which has greatly simplified this class.

ajordens:
Latest cut is looking pretty clean to me, nice work.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/config/SpringEnvironmentExtensionConfigResolver.kt
robzienert: For readers: Spring's PropertySources do not resolve to a non-primitive type. I wound up having to fold the enumerable property sources into a single `Properties` object, then use Jackson to read that value into a map, and finally use Jackson to cast the config at `pointer` to the desired config type.

PR #417, opened 2019-11-10
bump some dependencies including gradle

robfletcher:


spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -118,7 +118,7 @@ dependencies {
     api("commons-io:commons-io:2.5")
     api("de.danielbechler:java-object-diff:0.95")
     api("de.huxhorn.sulky:de.huxhorn.sulky.ulid:8.2.0")
-    api("dev.minutest:minutest:1.7.0")
+    api("dev.minutest:minutest:1.10.0")
```
robfletcher: Want this because it has proper support for jumping to the correct point in the test in IntelliJ

PR #420, opened 2019-11-16
feat(config): Add and prefer config files in ~/.config/spinnaker

ttomsu:
The [XDG Base Directory spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-0.6.html) suggests putting application config under `~/.config/<app_name>`, as the older `~/.<app_name>/` config dir clutters up the `$HOME` dir. A bunch of other modern applications already do this.

A follow up PR against Halyard would be needed to start putting files in this new directory.

### DIFF
--- a/kork-config/src/main/java/com/netflix/spinnaker/kork/boot/DefaultPropertiesBuilder.java
+++ b/kork-config/src/main/java/com/netflix/spinnaker/kork/boot/DefaultPropertiesBuilder.java
@@ -29,7 +29,7 @@ public class DefaultPropertiesBuilder {
     defaults.put("netflix.account", "${netflix.environment}");
     defaults.put("netflix.stack", "test");
     defaults.put("spring.config.name", "spinnaker,${spring.application.name}");
-    defaults.put("spring.config.additional-location", "${user.home}/.spinnaker/");
+    defaults.put("spring.config.additional-location", "${user.home}/.config/spinnaker,${user.home}/.spinnaker/");
     defaults.put("spring.profiles.active", "${netflix.environment},local");
     // add the Spring Cloud Config "composite" profile to default to a configuration
     // source that won't prevent app startup if custom configuration is not provided


ezimanyi:
This seems fine, but there may be additional changes necessary...it looks the `*-web.gradle` for each service sets `additional-location` as well.  I'm not sure if that overrides the defaults that are set here or if this change is enough.

cfieber:
The `additional-locations` are in the gradle files do override the properties here. They are used when running via gradle.

In the startup scripts they get set to `-Dspring.config.additional-location=/opt/spinnaker/config/` here: https://github.com/spinnaker/spinnaker-gradle-project/blob/master/spinnaker-dev-plugin/src/main/groovy/com/netflix/spinnaker/gradle/application/SpinnakerApplicationPlugin.groovy#L25

ttomsu:
Hmm. So it seems instead of adding this line I should actually remove it, since between gradle for devs and the plugin for actual installations, this is a red herring. Thoughts on removing it?

cfieber:
The third case is running within the IDE (not via gradle), in which case this serves as a reasonable default option.  I know in our group we most typically run that way (but we also override that property to a different location that supports our encrypted secrets).

I think this could be useful, and the equivalent change propagated to all the `-web.gradle` files if we wanted to head towards XDG support (also since this is backwards compatible anyway).

kork-config/src/main/java/com/netflix/spinnaker/kork/boot/DefaultPropertiesBuilder.java
```diff
@@ -29,7 +29,7 @@ public DefaultPropertiesBuilder() {
     defaults.put("netflix.account", "${netflix.environment}");
     defaults.put("netflix.stack", "test");
     defaults.put("spring.config.name", "spinnaker,${spring.application.name}");
-    defaults.put("spring.config.additional-location", "${user.home}/.spinnaker/");
+    defaults.put("spring.config.additional-location", "${user.home}/.config/spinnaker,${user.home}/.spinnaker/");
```
cfieber: this should be `${XDG_CONFIG_HOME}/spinnaker` (maybe with a fallback to `${user.home}/.config/spinnaker`

cfieber: also - if we are going this route would it make sense to do this conditionally by OS (XDG on linux, something like `${user.home}/Library/Preferences/io.spinnaker/` on OSX)?

PR #421, opened 2019-11-20
feat(stats): Add whitelist of cloud providers, as suggested by LF …

ttomsu:
…legal eagles.

@demobox @bdmorgan FYI

@duftler Did I get them all?

### DIFF
--- a/kork-proto/src/main/proto/stats/log.proto
+++ b/kork-proto/src/main/proto/stats/log.proto
@@ -63,8 +63,21 @@ message Stage {
 }
 
 message CloudProvider {
-  string id = 1;
+  ID id = 1;
   string variant = 2;
+
+  enum ID {
+    UNKNOWN = 0;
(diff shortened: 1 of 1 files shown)


ttomsu:
List obtained by looking through classes in Clouddriver that implement `CloudProvider` and grabbing their ID. Skipped `alicloud` for now since they're still working on it. Skipped `titus` because it's Netflix-only AFAIK.

duftler:
huaweicloud also seems to be in progress: https://github.com/spinnaker/rosco/pull/466

The list looks right to me, best as I can tell from doing the same thing you did.

demobox:
I'm wondering how we could easily be alerted if this list is no longer up-to-date?

ttomsu:
Perhaps a Stackdriver log-based metric if an excessive number of UNKNOWNS (in any enum field) starts occurring.

PR #423, opened 2019-11-21
fix(plugins): fix SpinnakerPluginManager initialization

claymccoy:
Plugins didn't actually load. AbstractPluginManager conflates instantiation and initialization. The SpinnakerPluginManager pluginStatusProvider was delegating to null (even with the proxy in place) and load plugins, enabling and disabling failed with NPEs. The most straight forward way to fix it was to circumvent the create* methods and just initialize the protected members direct that are passing into SpinnakerPluginManager. There are tests that show the problem and fail without the changes.

### DIFF
--- a/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginManager.kt
+++ b/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginManager.kt
@@ -23,22 +23,14 @@ import com.netflix.spinnaker.kork.plugins.loaders.SpinnakerDevelopmentPluginLoad
 import com.netflix.spinnaker.kork.plugins.loaders.SpinnakerJarPluginLoader
 import org.pf4j.CompoundPluginLoader
 import org.pf4j.DefaultPluginManager
-import org.pf4j.ExtensionFactory
-import org.pf4j.PluginDescriptorFinder
 import org.pf4j.PluginLoader
+import org.pf4j.PluginDescriptorFinder
 import org.pf4j.PluginStatusProvider
+import org.pf4j.PluginWrapper
 import java.nio.file.Path
 
 /**
  * The primary entry-point to the plugins system from a provider-side (services, libs, CLIs, and so-on).
- *
- * WARNING: Due to how [org.pf4j.AbstractPluginManager] is written, we have to jump through hoops to get injected
- * code to initialize correctly. Unfortunately, PF4J attempts to initialize everything on object creation, so we
- * don't have access to [SpinnakerPluginManager] properties in the `create*` methods. To work around this, the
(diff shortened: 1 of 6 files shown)


cfieber:
Thanks - looking at this. Wondering if there is a middle ground that doesn't invoke and throw away the results of the create methods (both the things we are calling back for are pretty tiny interfaces)

claymccoy:
Maybe, what do you have in mind?
I was trying to do the least invasive thing to make it work. I went a little further to make the members private and remove the complaints in the comment. I don't really have an opinion on how it should be done as long as it actually loads plugins.
There are some changes that could be done in PF4J to decouple the instantiation and initialization and make extending it like this easier. But I'm just working with what we have.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginManager.kt
```diff
@@ -23,36 +23,29 @@ import com.netflix.spinnaker.kork.plugins.loaders.SpinnakerDevelopmentPluginLoad
...
-  internal val statusProvider: PluginStatusProvider,
-  internal val configResolver: ConfigResolver,
+  private val statusProvider: PluginStatusProvider,
+  private val configResolver: ConfigResolver,
   pluginsRoot: Path
 ) : DefaultPluginManager(pluginsRoot) {
 
-  override fun createExtensionFactory(): ExtensionFactory {
-    return SpringExtensionFactory(this)
+  init {
```
cfieber: instead of the `init` block we could do:

```kotlin
  private val springExtensionFactory: ExtensionFactory = SpringExtensionFactory(this, configResolver)

  override fun createExtensionFactory(): ExtensionFactory = ExtensionFactoryDelegate()

  override fun createPluginStatusProvider(): PluginStatusProvider = PluginStatusProviderDelegate()

  private inner class ExtensionFactoryDelegate() : ExtensionFactory {
    override fun <T : Any?> create(extensionClass: Class<T>?): T = springExtensionFactory.create(extensionClass)
  }

  private inner class PluginStatusProviderDelegate() : PluginStatusProvider {
    override fun disablePlugin(pluginId: String?) = statusProvider.disablePlugin(pluginId)

    override fun isPluginDisabled(pluginId: String?): Boolean = statusProvider.isPluginDisabled(pluginId)

    override fun enablePlugin(pluginId: String?) = statusProvider.enablePlugin(pluginId)
  }
```

its definitely uglier looking, but it plays nice with the way the class expects to initialize and if someone tries to extend things in the future they don't need to know the magic of how the class initializes

what do you think?

claymccoy: Sure, that seems to work fine. I'm not a fan of the create* methods, but it is what is currently there so we should be consistent.

PR #424, opened 2019-11-27
feat(sql): Support for multiple SQL data sources

srekapalli:
- Conditionally add secondary liquibase target if configured.
- Conditionally add secondary JOOQ (DSLContext) to target a different dialect than the primary source. If two connection pools exist and they both are of same dialect we will not create secondary JOOQ bean. (eg: application having 2 pools against the same datasource will not need a secondary JOOQ)
- Configuration is backward compatible for the most part. Applications will need to wire up the JOOQ and pool as applicable on the app side of things.
- Tests exercise taking in the new sql config and verifies the bean existence as expected.

### DIFF
--- a/kork-sql/src/main/kotlin/com/netflix/spinnaker/kork/sql/config/DefaultSqlConfiguration.kt
+++ b/kork-sql/src/main/kotlin/com/netflix/spinnaker/kork/sql/config/DefaultSqlConfiguration.kt
@@ -74,7 +74,7 @@ class DefaultSqlConfiguration {
     SpringLiquibaseProxy(properties.migration)
 
   @Bean
-  @ConditionalOnProperty("sql.secondaryMigration.jdbcUrl")
+  @ConditionalOnProperty("sql.secondary-migration.jdbc-url")
   fun secondaryLiquibase(properties: SqlProperties): SpringLiquibase =
     SpringLiquibaseProxy(properties.secondaryMigration!!)
 
--- a/kork-sql/src/main/kotlin/com/netflix/spinnaker/kork/sql/config/DefaultSqlConfiguration.kt
+++ b/kork-sql/src/main/kotlin/com/netflix/spinnaker/kork/sql/config/DefaultSqlConfiguration.kt
@@ -76,7 +76,7 @@ class DefaultSqlConfiguration {
   @Bean
   @ConditionalOnProperty("sql.secondary-migration.jdbc-url")
   fun secondaryLiquibase(properties: SqlProperties): SpringLiquibase =
-    SpringLiquibaseProxy(properties.secondaryMigration!!)
+    SpringLiquibaseProxy(properties.secondaryMigration)
 
   @DependsOn("liquibase")
   @Bean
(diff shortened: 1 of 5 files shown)


kork-sql/src/main/kotlin/com/netflix/spinnaker/kork/sql/config/DefaultSqlConfiguration.kt
```diff
@@ -72,6 +73,11 @@ class DefaultSqlConfiguration {
   fun liquibase(properties: SqlProperties): SpringLiquibase =
     SpringLiquibaseProxy(properties.migration)
 
+  @Bean
+  @ConditionalOnProperty("sql.secondaryMigration.jdbcUrl")
```
cfieber: needs to be kebab-case for boot2:
```suggestion
  @ConditionalOnProperty("sql.secondary-migration.jdbc-url")
```

kork-sql/src/main/kotlin/com/netflix/spinnaker/kork/sql/config/SecondaryPoolDialectCondition.kt
```diff
@@ -0,0 +1,47 @@
...
+
+import org.springframework.boot.autoconfigure.condition.ConditionOutcome
+import org.springframework.boot.autoconfigure.condition.SpringBootCondition
+import org.springframework.context.annotation.ConditionContext
+import org.springframework.core.type.AnnotatedTypeMetadata
+import org.springframework.boot.context.properties.source.ConfigurationPropertyName
+import org.springframework.boot.context.properties.bind.Bindable
+import org.springframework.boot.context.properties.bind.Binder
+
+class SecondaryPoolDialectCondition : SpringBootCondition() {
```
cfieber: a unit test on this would be handy if its possible to set up all the conditional evaluation stuff in a sane way

I don't think SqlConfigurationTests was exercising the case of two pools with the same dialect

kork-sql/src/main/kotlin/com/netflix/spinnaker/kork/sql/config/SqlProperties.kt
```diff
@@ -20,6 +20,7 @@ import org.springframework.boot.context.properties.ConfigurationProperties
 @ConfigurationProperties("sql")
 data class SqlProperties(
   var migration: SqlMigrationProperties = SqlMigrationProperties(),
+  var secondaryMigration: SqlMigrationProperties? = null,
```
jonsie: Any reason not to have `var secondaryMigration: SqlMigrationProperties = SqlMigrationProperties()` ? You could then remove the `!!` operator in `DefaultSqlConfiguration` -`SpringLiquibaseProxy(properties.secondaryMigration!!)`

kork-sql/src/main/kotlin/com/netflix/spinnaker/kork/sql/config/SecondaryPoolDialectCondition.kt
srekapalli: Added scenario to test single dialect two conn.pools to SqlConfigurationTests suite itself. Setting up the config props to unit test conditional evaluation stuff will be tricky and hard but with this integration test it will be easy to wire up props more close to how it will be wired up in the real world.

kork-sql/src/main/kotlin/com/netflix/spinnaker/kork/sql/config/SqlProperties.kt
srekapalli: Good point.  Will do that.

PR #426, opened 2019-11-27
fix(plugins): plugin enable config did not use namespace

claymccoy:
My example plugin (https://github.com/claymccoy/pf4jStagePlugin) has a namespace. In order for it to load in orca the config must have the namespace in the config path, but it can't be in the enabled path.
```
spinnaker:
  plugins:
    RandomWaitPlugin:
      enabled: true
    Armory.RandomWaitPlugin:
      extensions:
        armory.randomWaitStage:
          enabled: true
          config:
            defaultMaxWaitTime: 60
```
This change unifies the path to be \<namespace\>.\<plugin name\>.
Since PF4J uses the pluginId, I've had to override our SpinnakerPluginDescriptor to use \<namespace\>.\<plugin name\> as the pluginId. There are still places that require just the plugin name without the namespace, so I've added an attribute pluginName to distinguish that.
I think that plugin namespace should probably be required, but I've made it optional for now. I noticed that https://github.com/robzienert/spinnaker-plugin-helloworld/blob/master/helloworld-orca/helloworld-orca.gradle does not have a plugin namespace.


robzienert:
My repo should have a plugin namespace. Oops.

PR #431, opened 2019-12-03
fix(selector/v2): fix regex parsing issue

asher:


kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/v2/SelectableService.java
```diff
@@ -322,7 +322,7 @@ public boolean equals(Object o) {
 
         for (Object v : values) {
           if (v instanceof String && ((String) v).startsWith("regex:")) {
-            final String regex = ((String) v).split(":")[1];
+            final String regex = ((String) v).substring(((String) v).indexOf(":") + 1);
```
asher: Given a config of:
```
regex:(?:this.*|that.*)
```
The prior logic assigned `(?` to regex, this change ensures that it will be set to `(?:this.*|that.*)`

PR #433, opened 2019-12-05
fix(plugins): allow unsafe plugins to load

claymccoy:
Unsafe plugins were putting jars and classes in their own ClassLoader, and then loading Classes from the parent delegate (a different class loader). This obviously failed.
The easiest fix for this was to make UnsafePluginClassLoader addURL append to the same `parent` class loader. Since addURL is protected this was done using reflection, and there is an assumption that `parent` is a URLClassLoader. Not ideal at all, but it is "unsafe" mode right?
I looked at making UnsafePluginClassLoader being more careful about who it delegates to, but PF4J makes that difficult. There may be better ways to handle this at a higher level like overriding AbstractPluginManager.loadPlugin or making changes to PF4J so that using the parent class loader exclusively is an option (basically making unsafe mode an option in PF4J core).
Some other issues are that unsafe plugin extensions are seen as system extensions. This is tricky because AbstractPluginManager.whichPlugin looks plugins up by class loader. That means that unsafe plugin extensions must be configured like system extensions. It also means that ExtensionBeanDefinitionRegistryPostProcessor needs to ignore registering unsafe plugin extensions, since it already loads them as system extensions and it causes a Spring exception when loading them twice.
This is probably not the long term solution, but that would mean more extensive changes (perhaps to PF4J core). This at least allows unsafe plugins to load.

### DIFF
--- a/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/ExtensionBeanDefinitionRegistryPostProcessor.kt
+++ b/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/ExtensionBeanDefinitionRegistryPostProcessor.kt
@@ -64,6 +64,7 @@ class ExtensionBeanDefinitionRegistryPostProcessor(
 
     log.debug("Creating plugin extensions")
     pluginManager.startedPlugins.forEach { plugin ->
+      if (plugin.isUnsafe()) return@forEach
       log.debug("Creating extensions for plugin '{}'", plugin.pluginId)
       pluginManager.getExtensionClassNames(plugin.pluginId).forEach {
         log.debug("Creating extension '{}' for plugin '{}'", it, plugin.pluginId)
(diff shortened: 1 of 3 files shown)


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/ExtensionBeanDefinitionRegistryPostProcessor.kt
```diff
@@ -64,6 +64,7 @@ class ExtensionBeanDefinitionRegistryPostProcessor(
 
     log.debug("Creating plugin extensions")
     pluginManager.startedPlugins.forEach { plugin ->
+      if (plugin.isUnsafe()) return@forEach
```
cfieber: is this here because the block above (`pluginManager.getExtensionClassNames(null)`) catches all the unsafe extensions?

claymccoy: Yes

claymccoy:
> It'd be worthwhile adding an issue into PF4J to see if they have appetite for an "unsafe" concept.

Absolutely, I'm hoping we can talk about that a bit in the platform sig today. Seems like we could improve PF4J in several ways.

PR #435, opened 2019-12-09
fix(secret): fixed SecretAwarePropertySource.getProperty so that when…

mayitbeegh:
… the property value is 'encryptedFile:...' it's decrypted as a file

### DIFF
--- a/kork-secrets/src/main/java/com/netflix/spinnaker/kork/secrets/SecretAwarePropertySource.java
+++ b/kork-secrets/src/main/java/com/netflix/spinnaker/kork/secrets/SecretAwarePropertySource.java
@@ -32,16 +32,17 @@ public class SecretAwarePropertySource extends EnumerablePropertySource<Enumerab
   public Object getProperty(String name) {
     Object o = source.getProperty(name);
     if (o instanceof String && EncryptedSecret.isEncryptedSecret((String) o)) {
+      String propertyValue = (String) o;
       if (secretManager == null) {
         throw new SecretException("No secret manager to decrypt value of " + name);
       }
       String lName = name.toLowerCase();
       if (isSamlFile(lName)) {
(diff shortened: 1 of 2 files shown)


kork-secrets/src/main/java/com/netflix/spinnaker/kork/secrets/SecretAwarePropertySource.java
```diff
@@ -32,16 +32,17 @@ public SecretAwarePropertySource(EnumerablePropertySource source, SecretManager
...
     Object o = source.getProperty(name);
     if (o instanceof String && EncryptedSecret.isEncryptedSecret((String) o)) {
+      String propertyValue = (String) o;
       if (secretManager == null) {
         throw new SecretException("No secret manager to decrypt value of " + name);
       }
       String lName = name.toLowerCase();
       if (isSamlFile(lName)) {
-        return "file:" + secretManager.decryptAsFile((String) o).toString();
-      } else if (isFile(lName) || EncryptedSecret.isEncryptedFile(lName)) {
```
mayitbeegh: `EncryptedSecret.isEncryptedFile(lName)` -> `EncryptedSecret.isEncryptedFile(propertyValue)`
It was trying match property name (such as 'server.ssl.key-store') instead of its value (encryptedFile:secrets-manager!r:us-west-2!s:keystore-local)

fieldju:
This PR fixes a bug that is preventing us from storing our SSL Certs and downloading them at run time.

You might notice that we are using a new impl `secrets-manager`, a PR for that is incoming soon (tm). We made an IMPl for [AWS Secrets Manager](https://aws.amazon.com/secrets-manager/)

ncknt:
Thanks for the fix @fieldju. Looking forward to AWS secret manager support!

PR #437, opened 2019-12-09
refactor(plugins): makes SpinnakerPluginDescriptor a subclass of DefaultPluginDescriptor.

cfieber:
This was largely to implement equals/hashCode but also plays slightly nicer with the
expected extension mechanism of PF4J.

There is a slightly worse immutability story here, but the base PluginDescriptor is
already a fully nullable POJO so I don't think it is too bad to have our new
attributes as var as well.

Also adds some PluginLoading tests (the original driver for the desire to have an
equality test on SpinnakerPluginDescriptor..). These are implemented as a generic
TCK as I have a follow-up PR that introduces a new plugin-type that exercises the
same set of tests for that type.

### DIFF
--- a/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginDescriptor.kt
+++ b/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginDescriptor.kt
@@ -15,7 +15,9 @@
  */
 package com.netflix.spinnaker.kork.plugins
 
+import org.pf4j.DefaultPluginDescriptor
 import org.pf4j.PluginDescriptor
+import java.util.Objects
 
 /**
  * Decorates the default [PluginDescriptor] with additional Spinnaker-specific metadata.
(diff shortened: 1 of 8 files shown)


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginDescriptor.kt
```diff
@@ -24,16 +26,54 @@ import org.pf4j.PluginDescriptor
...
+      super.getPluginId()
     } else {
-      return "$namespace.${baseDescriptor.pluginId}"
+      "$namespace.${super.getPluginId()}"
     }
+
+  val pluginName: String
+    get() = super.getPluginId()
+
+  // TODO(cf): rework once upstream equals/hashCode change is released:
```
cfieber: I've upstreamed equals/hashCode in pf4j/pf4j#352 but they are not in a release yet

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginDescriptor.kt
```diff
@@ -24,16 +26,54 @@ import org.pf4j.PluginDescriptor
...
+  provider,
+  license) {
+
+  override fun getPluginId(): String =
     if (namespace == "undefined") {
-      return baseDescriptor.pluginId
+      super.getPluginId()
     } else {
-      return "$namespace.${baseDescriptor.pluginId}"
+      "$namespace.${super.getPluginId()}"
```
robzienert: Since the namespace is being returned in the plugin id, do we want to get rid of the custom attribute in the manifest?

robzienert: Also, I think `.` as the delimiter could bust the current config resolver... `/` and `.` may be no-go's.

claymccoy: Yeah, namespace here is going to be redundant right? And I was wondering if '.' would introduce problems somewhere.

claymccoy: This definitely conflicts with #439.

robzienert: I managed to get `$namespace.$pluginId` to work in the other PR. If this gets merged before #439, that's cool. Either ordering works for me.

cfieber:
finally got back to this with post vacation mush-brain, and rebased over #439

PR #442, opened 2019-12-12
fix(dependencies): Spring Boot 2.2 upgrade issues

pdelagrave:
* Forgot to update `com.fasterxml.jackson:jackson-bom` to 2.10.1 in the boot-2.2 PR, leaving both 2.9.9 and 2.10.1 in the classpath of many services and causing orca to fail.

* Boot 2.2 brings Liquibase 3.8.2 which JAR contains a misplaced `META-INF/services/com.fasterxml.jackson.databind.Module` file.

  Jackson's `ObjectMapper.findModules()` will scan the classpath for such files in order to dynamically loads all Jackson modules found. The liquibase file contains `com.fasterxml.jackson.module.jaxb.JaxbAnnotationModule` but this class isn't provided by the liquibase JAR nor necessarily elsewhere in the classpath.

  Kork tests weren't impacted because they don't execute with liquibase loaded on the classpath.

  Orca for example is loading both kork-plugins and liquibase, causing `SpringEnvironmentExtensionConfigResolver` to fail initializing because its ObjectMapper can't find the module "published" by the liquibase jar.

  Only specifying the necessary Jackson modules fixes the issue without the code having to be modified again once liquibase's JAR is fixed.
https://liquibase.jira.com/browse/CORE-3522


pdelagrave:
Tested locally to be compatible with every services.

pdelagrave:
@robzienert @cfieber

PR #445, opened 2019-12-16
feat(plugins): Emit application events on plugin download

robzienert:


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/PluginUpdateService.kt
```diff
@@ -70,8 +77,14 @@ class PluginUpdateService(
       val updated = updateManager.updatePlugin(plugin.id, lastRelease.version)
       if (updated) {
         log.debug("Updated plugin '{}'", plugin.id)
+        applicationEventPublisher.publishEvent(
+          PluginDownloaded(this, UPDATE, SUCCEEDED, plugin.id, lastRelease.version)
+        )
       } else {
         log.error("Failed updating plugin '{}'", plugin.id)
+        applicationEventPublisher.publishEvent(
```
srekapalli: 👍

srekapalli: For this,  listener may be in the parent context will communicate to Front50 about downloading this. Something like that , or will that be too early because rest of the app didn't initialize yet.?

robzienert: I don't think we'd bother communicating to Front50 (or any other service, for that matter) on these application events. I don't have an explicit use for these events yet, but I could see power users compiling in their own (ServiceLoader-loaded) event listeners that could do some custom reporting or error handling.

PR #446, opened 2019-12-16
feat(plugins): allow config to be super class on extension

claymccoy:
This is done in order to use the following pf4j feature with a configurable extension.
"Support any interface as an ExtensionPoint" https://github.com/pf4j/pf4j/pull/350

This is nice because I can extend something that is not an extension point, and even do it in safe mode. The problem is that now my extension can only have one interface. So I can't use the ConfigurableExtension directly. I can only have one interface unless pf4j is modified to allow more.
Using my one super class slot for configuration is still not ideal, but it works. It might be nice to allow some other mechanism for config other than class hierarchy or modify pf4j to handle config natively.


claymccoy:
This is purposefully built on kork 6.22.2 because I am testing this inside of Orca, which doesn't easily work with any newer version of kork.

PR #448, opened 2019-12-18
feat(plugins): Add ExtensionInvocationProxy and InvocationAspect with metric invocation implementation

jonsie:
Closes https://github.com/spinnaker/spinnaker/issues/5240.

Todo:
- [x] Write better javadocs
- [x] Write and fix tests
- [x] Add a LogInvocationAspect
- [x] Support extension name as part of metric namespace / tag
- [x] Cache metric IDs (similar to `InstrumentedProxy`)


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/ExtensionBeanDefinitionRegistryPostProcessor.kt
```diff
@@ -66,11 +69,14 @@ class ExtensionBeanDefinitionRegistryPostProcessor(
...
     }
 
-    log.debug("Creating plugin extensions")
     pluginManager.startedPlugins.forEach { plugin ->
       if (plugin.isUnsafe()) return@forEach
       log.debug("Creating extensions for plugin '{}'", plugin.pluginId)
-      pluginManager.getExtensionClassNames(plugin.pluginId).forEach {
+
+      val pluginExtensions = pluginManager.getExtensionClassNames(plugin.pluginId)
+      if (pluginExtensions.isNullOrEmpty()) log.warn("No extensions found for plugin '{}'", plugin.pluginId)
```
jonsie: Minor warn log I added here to let us know when a plugin does not have extensions.

kork-plugins/kork-plugins.gradle
```diff
@@ -6,6 +6,8 @@ apply from: "$rootDir/gradle/lombok.gradle"
 dependencies {
   api(platform(project(":spinnaker-dependencies")))
   api project(":kork-exceptions")
+  api "com.netflix.spectator:spectator-api"
```
robzienert: Should be `implementation`

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/ExtensionBeanDefinitionRegistryPostProcessor.kt
robzienert: All of this logic should be going in the `SpinnakerExtensionFactory`, rather than in this class.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/proxy/ExtensionInvocationProxy.kt
```diff
@@ -0,0 +1,88 @@
...
+      invocationAspects: List<InvocationAspect<InvocationHolder>>,
+      descriptor: SpinnakerPluginDescriptor
+    ): Any {
+      return Proxy.newProxyInstance(
+        target.javaClass.classLoader,
+        target.javaClass.interfaces,
+        ExtensionInvocationProxy(target, invocationAspects, descriptor)
+      )
+    }
+  }
```
robzienert: Companion objects should come list in a class definition.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/proxy/ExtensionInvocationProxy.kt
```diff
@@ -0,0 +1,88 @@
...
+        it.supports(holder.javaClass)
+      }?.error(e, holder)
+    }
+  }
+
+  private fun MutableSet<InvocationHolder>.after(success: Boolean) {
+    this.forEach { holder ->
+      invocationAspects.find {
+        it.supports(holder.javaClass)
+      }?.after(success, holder)
```
robzienert: I think these methods should allow for the case where multiple aspects can support a particular `InvocationHolder`.

```
invocationAspects.findAll { it.supports(holder.javaClass) }.forEach { it.after(success, holder) }
```

etc.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/proxy/aspects/holders.kt
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.kork.plugins.proxy.aspects
+
+import com.netflix.spectator.api.Id
+
+interface InvocationHolder
```
robzienert: Holder of what? I don't mind longer type names if it makes reasoning about the code easier. Perhaps something like `MethodInvocationState`?

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/ExtensionBeanDefinitionRegistryPostProcessor.kt
jonsie: Actually, the extension factory is after a list of extension classes has been found - I'll need to override `getExtensionClassNames` in `SpinnakerPluginManager` - that'll be cleaner.

jonsie: Through conversation determined this comment was actually meant for:

```val bean = ExtensionInvocationProxy.proxy(
          pluginManager.extensionFactory.create(extensionClass),
          invocationAspects as List<InvocationAspect<InvocationHolder>>,
          plugin.descriptor as SpinnakerPluginDescriptor)```

Addressing that.

File: kork-plugins/src/main/java/com/netflix/spinnaker/config/PluginsAutoConfiguration.java [hunk 6]
```diff
@@ -106,13 +113,30 @@ public static PluginUpdateService pluginUpdateManagerAgent(
     return new PluginUpdateService(updateManager, pluginManager, applicationEventPublisher);
   }
 
+  @Bean
+  @ConditionalOnMissingBean(Registry.class)
+  public static Registry registry() {
+    return new DefaultRegistry(Clock.SYSTEM);
+  }
```
robzienert: Oh, heh. Yeah, this isn't going to work. We'll need to do some more Spring-y things. Basically the issue here is that we're creating a `Registry` class _way_ early in the application lifecycle and adding it to the bean registry. Despite the `@ConditionalOnMissingBean` annotation, this will forever be the `Registry` that is wired up into applications.

We'll need to change the `Registry` dependency in `metricInvocationAspect` to be `Provided<Registry>`. This interface is normally used for working around circular bean dependencies, but in this case it'll let us use the configured `Registry` later. I don't think we'll lose any metrics as a result of this, but we'll need to protect against it anyway and log out a `warn` or something whenever we `get()` the provided `Registry` when it throws. This would happen in method invocations of plugins that are running before the Registry has been initialized. We can have a fallback for holding onto internal metrics and dumping once the dependency is available as a separate PR.

robzienert: This would probably be easier in person - we can chat tomorrow if this doesn't make sense.

jonsie: That all makes sense to me.  I wonder if `ObjectProvider` will work well for us here.

kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/ExtensionBeanDefinitionRegistryPostProcessorTest.kt
```diff
@@ -102,9 +103,10 @@ class ExtensionBeanDefinitionRegistryPostProcessorTest : JUnit5Minutests {
     val pluginWrapper: PluginWrapper = mockk(relaxed = true)
     val extensionFactory: ExtensionFactory = mockk(relaxed = true)
     val applicationEventPublisher: ApplicationEventPublisher = mockk(relaxed = true)
+    val invocationAspects: List<InvocationAspect<*>> = mockk(relaxed = true)
     val pluginDescriptor: SpinnakerPluginDescriptor = mockk(relaxed = true)
 
-    val subject = ExtensionBeanDefinitionRegistryPostProcessor(pluginManager, updateService, applicationEventPublisher)
+    val subject = ExtensionBeanDefinitionRegistryPostProcessor(pluginManager, updateService, applicationEventPublisher, invocationAspects)
```
robzienert: nit: Wrap so it's sub-120 char.

kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/proxy/aspects/MetricInvocationAspectTest.kt
```diff
@@ -0,0 +1,124 @@
...
+      val state = subject.createState(target, proxy, method, args, spinnakerPluginDescriptor)
+      val logState = LogInvocationState("foo", "bar")
+      expectThat(subject.supports(state.javaClass)).isTrue()
+      expectThat(subject.supports(logState.javaClass)).isFalse()
+    }
+  }
+
+  private inner class Fixture {
+    val timing: String = "timing"
+    val invocations: String = "invocations"
```
robzienert: What is gained by these?

kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/proxy/aspects/utils.kt
```diff
@@ -0,0 +1,48 @@
...
+
+internal class SomeExtension {
+  fun helloWorld(): String {
+    return "Hello Public World!"
+  }
+
+  private fun privateHelloWorld(): String {
+    return "Hello Private World!"
+  }
+}
```
robzienert: Even for test code, I think documenting DSL extension methods is pretty important.

kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/proxy/aspects/MetricInvocationAspectTest.kt
jonsie: Reuse.  If you look through the test we make a number of assertions with these variables, ie:
```
"$pluginId.helloWorld.$timing"
```

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/proxy/ExtensionInvocationProxy.kt
```diff
@@ -0,0 +1,96 @@
...
+    var success = false
+    try {
+      result = method.invoke(target, *(args ?: arrayOfNulls<Any>(0)))
+      success = true
+    } catch (e: InvocationTargetException) {
+      invocationStates.error(e)
+      throw e.cause ?: RuntimeException("Caught invocation target exception without cause.", e)
+    } finally {
+      invocationStates.after(success)
+    }
```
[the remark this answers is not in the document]
jonsie: I think I can dig this.  Trying it out.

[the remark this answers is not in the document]
jonsie: I think if I do this I'll remove the default implementation of `error`.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/proxy/aspects/MetricInvocationAspect.kt
```diff
@@ -0,0 +1,159 @@
...
+ * Two metrics will be recorded for any extension: timing and invocations, with an additional tag
+ * for "success", having either the value "success" or "failure".
+ *
+ */
+class MetricInvocationAspect(
+  private val registryProvider: ObjectProvider<Registry>
+) : InvocationAspect<MetricInvocationState> {
+
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+  private val methodMetricIds: ConcurrentHashMap<Method, MetricIds> = ConcurrentHashMap()
```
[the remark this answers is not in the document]
jonsie: I toyed with this idea, but then decided since we're only instrumenting public methods on extension points, the list of metric IDs will probably (ahem) not grow too large.  I'll investigate this more though.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/proxy/aspects/method.kt
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.kork.plugins.proxy.aspects
+
+import java.lang.reflect.Method
+import java.lang.reflect.Modifier
+
+internal fun Method.isAllowed(): Boolean {
```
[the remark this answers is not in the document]
jonsie: Fair point.  I'll add some docs, and I'm also going to change where this is called such that only methods are are `allowed` are eligible for `InvocationAspect` instrumentation.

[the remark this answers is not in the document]
jonsie: Actually, scratch that - I'm going to let the `InvocationAspect` determine if the method is allowed for instrumentation.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/proxy/aspects/MetricInvocationAspect.kt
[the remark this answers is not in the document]
jonsie: Went with a Caffeine cache.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/proxy/aspects/MetricInvocationAspect.kt
```diff
@@ -0,0 +1,172 @@
...
+ */
+class MetricInvocationAspect(
+  private val registryProvider: ObjectProvider<Registry>
+) : InvocationAspect<MetricInvocationState> {
+
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  private val methodMetricIds: Cache<Method, MetricIds> = Caffeine.newBuilder()
+    .maximumSize(1000)
+    .expireAfterWrite(1, TimeUnit.HOURS)
```
jonsie: These seem like sane defaults to me.

PR #449, opened 2019-12-18
feat(plugins): New actuator endpoint to expose spinnaker plugins info…

srekapalli:
- New actuator endpoint **(/plugins and /plugins/{pluginId})**
- Provides plugin descriptor information.

### DIFF
--- a/kork-plugins/kork-plugins.gradle
+++ b/kork-plugins/kork-plugins.gradle
@@ -12,6 +12,7 @@ dependencies {
   implementation "org.pf4j:pf4j-update"
 
   implementation "org.springframework.boot:spring-boot-starter-web"
+  implementation "org.springframework.boot:spring-boot-starter-actuator"
   implementation("com.google.guava:guava")
 
   implementation("com.fasterxml.jackson.dataformat:jackson-dataformat-yaml")
@@ -25,4 +26,8 @@ dependencies {
 
   testRuntimeOnly "cglib:cglib-nodep"
   testRuntimeOnly "org.objenesis:objenesis"
+}
+
+compileJava {
+  options.compilerArgs << '-parameters'
 }
--- a/kork-plugins/src/main/java/com/netflix/spinnaker/config/PluginsAutoConfiguration.java
+++ b/kork-plugins/src/main/java/com/netflix/spinnaker/config/PluginsAutoConfiguration.java
@@ -22,6 +22,7 @@ import com.netflix.spinnaker.kork.plugins.SpinnakerPluginManager;
 import com.netflix.spinnaker.kork.plugins.SpringPluginStatusProvider;
 import com.netflix.spinnaker.kork.plugins.config.ConfigResolver;
 import com.netflix.spinnaker.kork.plugins.config.SpringEnvironmentExtensionConfigResolver;
+import com.netflix.spinnaker.kork.plugins.spring.actuator.SpinnakerPluginEndPoint;
 import com.netflix.spinnaker.kork.plugins.update.PluginUpdateService;
 import com.netflix.spinnaker.kork.plugins.update.SpinnakerUpdateManager;
 import java.net.MalformedURLException;
@@ -110,5 +111,11 @@ public class PluginsAutoConfiguration {
       ApplicationEventPublisher applicationEventPublisher) {
     return new ExtensionBeanDefinitionRegistryPostProcessor(
         pluginManager, updateManagerService, applicationEventPublisher);
+  }
+
+  @Bean
+  public static SpinnakerPluginEndPoint spinnakerPluginEndPoint(
+      SpinnakerPluginManager pluginManager) {
+    return new SpinnakerPluginEndPoint(pluginManager);
   }
 }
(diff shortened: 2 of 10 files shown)


kork-plugins/src/main/java/com/netflix/spinnaker/kork/plugins/spring/actuator/SpinnakerPluginEndPoint.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.stream.Collectors;
+import org.pf4j.PluginDescriptor;
+import org.pf4j.PluginWrapper;
+import org.springframework.boot.actuate.endpoint.annotation.Endpoint;
+import org.springframework.boot.actuate.endpoint.annotation.ReadOperation;
+import org.springframework.boot.actuate.endpoint.annotation.Selector;
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@Endpoint(id = "spinnakerplugins")
```
jonsie: I don't think the endpoint should be `spinnakerplugins` - all the plugins here will be for Spinnaker.  Maybe `/actuator/plugins` will do?

srekapalli: Yh. plugins might be too generic and possibility of actuator given endpoint with that name is high so I went with that.

ajordens: +1 to `/plugins` personally w/ the risk of something built-in coming and squashing it later (that seems unlikely to me).

kork-plugins/src/main/java/com/netflix/spinnaker/config/PluginsAutoConfiguration.java
```diff
@@ -111,4 +112,10 @@ public static ExtensionBeanDefinitionRegistryPostProcessor pluginBeanPostProcess
     return new ExtensionBeanDefinitionRegistryPostProcessor(
         pluginManager, updateManagerService, applicationEventPublisher);
   }
+
+  @Bean
+  public static SpinnakerPluginEndPoint spinnakerPluginEndPoint(
```
ajordens: I believe `EndPoint` should be `Endpoint`.

kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/spring/actuator/SpinnakerPluginEndPointTest.kt
```diff
@@ -0,0 +1,72 @@
...
+import strikt.assertions.isEqualTo
+
+class SpinnakerPluginEndPointTest : JUnit5Minutests {
+
+  fun tests() = rootContext<Fixture> {
+    fixture {
+      Fixture()
+    }
+
+    test("end point should return list of plugins") {
```
ajordens: s/end point/endpoint

kork-plugins/src/main/java/com/netflix/spinnaker/kork/plugins/spring/actuator/SpinnakerPluginEndPoint.java
robzienert: Yes, and also leave out `"/actuator"` from the endpoint URL, since Spring config handles where actuator endpoints are rooted on.

kork-plugins/src/main/java/com/netflix/spinnaker/kork/plugins/spring/actuator/SpinnakerPluginEndPoint.java
```diff
@@ -0,0 +1,61 @@
...
+  public PluginDescriptor pluginById(@Selector String pluginId) {
+    PluginWrapper pluginWrapper = pluginManager.getPlugin(pluginId);
+    if (pluginWrapper == null) {
+      throw new PluginNotFoundException("Plugin not found: " + pluginId);
+    }
+    return pluginWrapper.getDescriptor();
+  }
+
+  @ResponseStatus(value = HttpStatus.NOT_FOUND)
+  public static class PluginNotFoundException extends RuntimeException {
```
robzienert: We have a `NotFoundException` in kork that should probably be used instead that is handled automatically.

srekapalli: ah!!  Will remove this...

kork-plugins/kork-plugins.gradle
```diff
@@ -26,3 +27,7 @@ dependencies {
   testRuntimeOnly "cglib:cglib-nodep"
   testRuntimeOnly "org.objenesis:objenesis"
 }
+
+compileJava {
+  options.compilerArgs << '-parameters'
```
jonsie: Is this necessary?

srekapalli: Yh. I ran into issues because our spinnaker apps don't inherit from springboot starter parent which takes care of this complier flag setting apparently.
https://github.com/spring-projects/spring-boot/issues/11010

jonsie: Ah, thanks for the explanation 👍

PR #452, opened 2019-12-18
test(plugins): adding a failing test re: pluginsConfigurationProperties

link108:
PluginsConfigurationProperties is not respecting properties, I've added a test that shows this happening. 

I put some breakpoints in PluginsAutoConfiguration.java and was able to see the properties I expect to be there are present, but its not clear why PluginsConfigurationProperties is not being created correctly: https://share.getcloudapp.com/4gux9pWx

I also checked out the bean definition names, the bean name for PluginsConfigurationProperties is prefixed with `spinnaker.extensibility-`, which seems odd: https://share.getcloudapp.com/eDu9BJL4


robzienert:
Build is failing.

PR #453, opened 2019-12-19
fix(plugins): Correctly read remote repositories from config

robzienert:
I forgot that Spring configs are not available when we're setting up the plugin framework, so remote repository configuration did not work. I've refactored `SpringEnvironmentConfigResolver` a little bit so it's not just for extension configuration, but can be used broadly for loading whatever kind of configs we need as POJOs, instead of resorting to only `Environment.getProperty`, etc.

I made a couple other changes while testing this stuff out, I think some of it related to the Boot 2.2 upgrade:

- Added `@ImportAutoConfiguration` excludes for some resilience4j stuff. It was causing Orca to crash while starting up. I'd like to have resilience4j all autowired, but it's an issue for another day.
- Moved plugin updates / downloads before PluginManager's call to `loadPlugins`. I was finding that if a plugin already exists and it's downloaded again, it'll crash the application. The current solution still causes an `ERROR` log, but doesn't crash anything. The error is just that the plugin is already loaded, but this doesn't break anything. Will fix separately.

### DIFF
--- a/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/config/ConfigResolver.kt
+++ b/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/config/ConfigResolver.kt
@@ -15,6 +15,7 @@
  */
 package com.netflix.spinnaker.kork.plugins.config
 
+import com.fasterxml.jackson.core.type.TypeReference
 import com.netflix.spinnaker.kork.annotations.Alpha
 
 /**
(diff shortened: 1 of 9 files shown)


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/config/ConfigResolver.kt
```diff
@@ -29,23 +30,53 @@ interface ConfigResolver {
...
+   * instead of [Map], so that a default can be returned should no config exist at the desired [coordinates].
+   */
+  fun <T> resolve(coordinates: ConfigCoordinates, expectedType: TypeReference<T>): T
 }
 
-sealed class ConfigCoordinates(
-  val extensionId: String
-)
+sealed class ConfigCoordinates {
+  abstract fun toPointer(): String
```
robzienert: Wee bit of refactoring here. Now the pointers live on these coordinates, so there should be less reason to modify the actual `ConfigResolver` in the future, even as we add new config for the framework.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/config/SpringEnvironmentConfigResolver.kt
```diff
@@ -57,36 +59,43 @@ class SpringEnvironmentExtensionConfigResolver(
...
+    resolveInternal(coordinates, {
+      // Yo, it's fine. Totally fine.
+      @Suppress("UNCHECKED_CAST")
+      val type = ((expectedType.type as ParameterizedType).rawType as Class<T>)
+      if (type.isInterface) {
+        // TODO(rz): Maybe this should be supported, but there's only one use at this point, and we can just call in
+        //  with HashMap instead of Map.
+        throw SystemConfigException("Expected type must be a concrete class, interface given")
+      }
+      type.newInstance() as T
```
robzienert: Getting from a `TypeReference` object to something we can actually instantiate is a little dodgy, but this is what was recommended by StackOverflow - which, as we know - is infallible.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/SpinnakerUpdateManager.kt
```diff
@@ -15,14 +15,84 @@
...
 import org.pf4j.update.UpdateManager
 import org.pf4j.update.UpdateRepository
+import org.slf4j.LoggerFactory
 
 /**
  * TODO(rz): Update [hasPluginUpdate] such that it understands the latest plugin is not always the one desired
  */
 class SpinnakerUpdateManager(
-  pluginManager: PluginManager,
+  private val pluginManager: PluginManager,
```
robzienert: All of these changes is because PF4J still has not made a release with my fixes into `pf4j-update`, which was causing a hassle during testing. This just allows PF4J to replace existing plugins if they already exist in the plugins directory... which seems like the obvious behavior if we're trying to update plugins.

jonsie: Do you mind adding a link to that change in the comments here?

robzienert: https://github.com/pf4j/pf4j-update/pull/46

PR #455, opened 2019-12-20
chore(dependencies): Upgrade Spring Cloud to Hoxton.SR1

scottfrederick:
The Spring Cloud Hoxton.SR1 release train includes a new version of Spring Cloud Config that provides additional capabilities for Spinnaker. 

* Additional Vault authentication options such as AWS EC2, AWS IAM, Google Cloud Compute, Google Cloud IAM, and Kubernetes. This requires a new dependency on Spring Cloud Vault. 

* AWS s3 as an additional Config Server backend option, including the ability to serve credentials files via the [`configserver:` syntax](https://www.spinnaker.io/setup/configuration/#configuration-files). This requires a new dependency on Spring Cloud AWS. 

* Redis as an additional Config Server backend option. 

Corresponding documentation updates in https://github.com/spinnaker/spinnaker.github.io/pull/1653

### DIFF
--- a/kork-config/kork-config.gradle
+++ b/kork-config/kork-config.gradle
@@ -11,6 +11,8 @@ dependencies {
 
   api "org.springframework.cloud:spring-cloud-context"
   api "org.springframework.cloud:spring-cloud-config-server"
+  api "org.springframework.cloud:spring-cloud-starter-aws"
+  api "org.springframework.vault:spring-vault-core"
 
   testImplementation "org.springframework.boot:spring-boot-starter-test"
   testImplementation "org.assertj:assertj-core"
--- a/kork-config/src/main/java/com/netflix/spinnaker/kork/configserver/autoconfig/SpringCloudAwsConfiguration.java
+++ b/kork-config/src/main/java/com/netflix/spinnaker/kork/configserver/autoconfig/SpringCloudAwsConfiguration.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2019 Pivotal, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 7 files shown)


kork-config/kork-config.gradle
```diff
@@ -11,6 +11,8 @@ dependencies {
 
   api "org.springframework.cloud:spring-cloud-context"
   api "org.springframework.cloud:spring-cloud-config-server"
+  api "org.springframework.cloud:spring-cloud-starter-aws"
+  api "org.springframework.vault:spring-vault-core"
```
robzienert: I'm not particularly keen on adding a specific cloud provider to a core module. We do this elsewhere in kork and other services, but I don't believe it's a good pattern to follow, since not everyone is actually running on AWS, or using Vault.

The plugins work will help address this, so maybe we can just let this slide in with a TODO to remove the deps later on? I'd prefer to see a user story of "adding support for [AWS, Vault, GCP, etc]" to be a matter of adding a plugin, rather than it coming along for the ride for all installations.

kork-config/src/main/java/com/netflix/spinnaker/kork/configserver/autoconfig/SpringCloudAwsConfiguration.java
```diff
@@ -0,0 +1,79 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.env.ConfigurableEnvironment;
+import org.springframework.core.env.MapPropertySource;
+import org.springframework.core.env.PropertySource;
+
+@Configuration
+@AutoConfigureBefore({
+  ContextRegionProviderAutoConfiguration.class,
+  ContextStackAutoConfiguration.class
+})
```
robzienert: Whoa. I didn't know about this annotation. Very handy!

cfieber:
@ethanfrogers / @scottfrederick I would love to know how close this puts us to deprecating / removing kork-secrets with the vault implementations that exist in configserver

robzienert:
@cfieber The next Platform SIG is going to have a topic about `kork-secrets` and SCCS. I'm hoping for a go-forward plan from this.

robzienert:
Ping.

scottfrederick:
This PR was in limbo while the Boot 2.2 upgrade was rolled back for 1.18 and re-applied after the release. I've re-based it on master, hopefully it will be ready to go.

robzienert:
By "will" do you mean if this build passes we're good to merge, or are there other things that should be checked prior to clicking the button? Happy to green-light whenever you're ready to.

scottfrederick:
This should be ready to merge when the build is green.

PR #457, opened 2019-12-27
feat(archaius): Allow users to disable Archaius

ncknt:
The polling in Archaius is causing issues in some environments that do not use it: unnecessary reloads and when secrets - any, not just `kork-secrets` - are referenced they are being reloaded which can be costly and cause temporary files to keep being created.

I'm not sure if this is the best way to disable Archaius but this adds a simple `archaius.enabled` flag to turn it off.

cc: @cfieber

### DIFF
--- a/kork-core/src/main/java/com/netflix/spinnaker/kork/archaius/ArchaiusConfiguration.java
+++ b/kork-core/src/main/java/com/netflix/spinnaker/kork/archaius/ArchaiusConfiguration.java
@@ -38,7 +38,7 @@ import org.springframework.core.io.Resource;
 import org.springframework.core.io.support.ResourcePropertySource;
 
 @Configuration
-@ConditionalOnExpression("${archaius.enabled:true}")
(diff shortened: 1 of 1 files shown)


ethanfrogers:
@cfieber @robzienert can either of you TAL at this?

kork-core/src/main/java/com/netflix/spinnaker/kork/archaius/ArchaiusConfiguration.java
```diff
@@ -37,6 +38,7 @@
 import org.springframework.core.io.support.ResourcePropertySource;
 
 @Configuration
+@ConditionalOnExpression("${archaius.enabled:true}")
```
cfieber:
```suggestion
@ConditionalOnProperty(value = "archaius.enabled", matchIfMissing = true)
```

I'd prefer an explicit property reference to an expression here

robzienert: Save a Developer; Avoid SpEL.

PR #459, opened 2020-01-06
chore(dependencies): Reverting Spring Boot 2.2 upgrade

pdelagrave:


cfieber:
I'm just getting my brain back into work mode but is there a description somewhere with the problems that the 2.2 upgrade has caused?

pdelagrave:
@cfieber Two issues that have been discussed separately on the #dev Slack channel, but it's good to summarize here:

1. Gate was made unusable by an important bug in the new spring-web version, the fix is delicate and we won't have time to make this work before the next Spinnaker version. https://spinnakerteam.slack.com/archives/C0DPVDMQE/p1576877302011700
  @ezimanyi has context on this too.

2. The Orca boot 2.2 patch was already reverted as it caused an issue with pipeline templates. @louisjimenez encountered the issue and reverted initially.

The plan was to fix #2 without reverting anything but #1 and the holidays got in the way. Now to keep things working and not have lots of mixed spring version everywhere and complicated gradle transitive dependencies version selection rules it's just safer to revert everything.

robzienert:
So are we reverting all services, too, then?

pdelagrave:
Yes, the PRs were submitted. Most are failing the CI tests because the unreleased expected versions of kork, keiko, fiat, orca and clouddriver were already updated in their `gradle.properties`.

PR #460, opened 2020-01-07
chore(dependencies): jackson 2.10.2

sullis:


cfieber:
this will be addressed via #482

PR #462, opened 2020-01-13
feat(plugins): Add ability to download plugins with an external process

robzienert:
⚠️ Do not merge. Just putting this out there, I need to do another pass on it and clean up the code. ⚠️ 

Prior to this PR, plugins can only be downloaded from unauthenticated HTTP
endpoints. This functionality adds the capability of specifying new
file downloader strategies, with an initial alternative strategy for
running an external process to download a plugin artifact.

We will be writing a new default FileDownloader that can be configured for
custom certificates, as well as basic auth, and probably other integrations
for artifact repositories that need something extra.


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/UpdateRepositoryFactory.kt
```diff
@@ -0,0 +1,80 @@
...
+import org.pf4j.update.FileDownloader
+import org.pf4j.update.SimpleFileDownloader
+import org.pf4j.update.UpdateRepository
+import org.pf4j.update.verifier.CompoundVerifier
+import java.net.URL
+
+/**
+ * Factory for [UpdateRepository].
+ */
+class UpdateRepositoryFactory(
```
jonsie: I think this factory will need to support more than a `ConfigurableUpdateRepository` - for example, there will be a concrete `Front50UpdateRepository` implementation that we will want to instantiate.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/downloader/internal/DefaultProcessRunner.kt
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Default [ProcessFileDownloader.ProcessRunner] implementation.
+ */
+internal class DefaultProcessRunner : ProcessFileDownloader.ProcessRunner {
+
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  override fun completeOrTimeout(processBuilder: ProcessBuilder): String {
+    val process = processBuilder.start()
+    val finished = process.waitFor(1, TimeUnit.MINUTES)
```
jonsie: Will this wait for 1 minute regardless of the proc status?

robzienert: That's right.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/UpdateRepositoryFactory.kt
robzienert: I agree - but since we only have one `UpdateRepository` implementation today, this is all I can really build for.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/downloader/internal/DefaultProcessRunner.kt
robzienert: I've changed this to not have a timeout and to wait for the process to finish.

PR #464, opened 2020-01-14
fix(artifacts): Revert making constructors private

ezimanyi:
This reverts commit 66d3391b2c071ea932d972343e6fa32a9452bb91.

When bumping the services, I realized that deserialization with Jackson is using the all-args constructor implicitly.  We need to configure the Artifact and ExpectedArtifact classes to use the builder for deserialization.  I will make this change and re-apply later today, but I don't want to leave things broken all day so reverting for now.


emjburns:
😬 yikes! Good catch

PR #466, opened 2020-01-14
feat(plugins): plugin reference plugin type for local development

cfieber:
This introduces a new plugin type that is registered if pf4j is
set to development mode to enable a local plugin development
experience.

A plugin-ref is a JSON file with a pointer to the actual location
of a plugin and a list of the classes and libraries to include
when building the PluginClasspath.

This allows a runtime experience similar to dropping a fully
packaged plugin into the host application, without requiring
the packaging and deployment step - aside from a one time
generation and copy/link of the plugin-ref file.

This changeset includes all the plumbing to make plugin-ref
a supported type of plugin, but there is still work to do
in tooling and documentation to make the development
experience reasonable.

With this changeset and the following set of steps, you can
work with the host application and the plugin project open
in IntelliJ, launch the host application, and debug into
the plugin project. To do this you need to:

1. create a plugin-ref file, which is a JSON document:
```
{
  "pluginPath": "/path/to/plugin",
  "classesDirs": [
    "paths",
    "with",
    "code",
    "e.g. PLUGIN_PROJECT/build/classes/java/main"],
  "libsDirs": [
    "paths",
    "with",
    "jars"]
}
```

2. copy or link the plugin-ref file into the host
applications plugins directory

3. configure the host-application with the plugin enabled:
```
spinnaker.extensibility:
  plugins:
    <my plugin id>:
      enabled: true
```

4. run the host application with `pf4j.mode=development`

Given that set of steps you can then open the host
application project in IntelliJ. You can then add the
plugin project as a new project into the same workspace (
on the gradle tab, the `+` button).
Then create a Run Configuration for the host application,
but edit the Before Launch to do a Build Project instead
of just Build.

When you have alllll these steps complete, if you run the
host application you should see it pick up your plugin-ref,
and add your plugin. Additionally you can run in debug
mode and you will be able to set breakpoints in your plugin
and debug through it.

### DIFF
--- a/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginManager.kt
+++ b/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginManager.kt
@@ -19,14 +19,18 @@ import com.netflix.spinnaker.kork.annotations.Beta
 import com.netflix.spinnaker.kork.plugins.bundle.PluginBundleExtractor
 import com.netflix.spinnaker.kork.plugins.config.ConfigResolver
 import com.netflix.spinnaker.kork.plugins.finders.SpinnakerPluginDescriptorFinder
+import com.netflix.spinnaker.kork.plugins.loaders.PluginRefPluginLoader
 import com.netflix.spinnaker.kork.plugins.loaders.SpinnakerDefaultPluginLoader
 import com.netflix.spinnaker.kork.plugins.loaders.SpinnakerDevelopmentPluginLoader
 import com.netflix.spinnaker.kork.plugins.loaders.SpinnakerJarPluginLoader
+import com.netflix.spinnaker.kork.plugins.repository.PluginRefPluginRepository
 import org.pf4j.CompoundPluginLoader
+import org.pf4j.CompoundPluginRepository
 import org.pf4j.DefaultPluginManager
 import org.pf4j.ExtensionFactory
 import org.pf4j.PluginDescriptorFinder
 import org.pf4j.PluginLoader
+import org.pf4j.PluginRepository
 import org.pf4j.PluginStatusProvider
 import org.pf4j.PluginWrapper
 import java.nio.file.Path
@@ -67,12 +71,13 @@ open class SpinnakerPluginManager(
 
   override fun createPluginLoader(): PluginLoader =
     CompoundPluginLoader()
+      .add(PluginRefPluginLoader(this), this::isDevelopment)
       .add(SpinnakerDevelopmentPluginLoader(this), this::isDevelopment)
       .add(SpinnakerDefaultPluginLoader(this))
       .add(SpinnakerJarPluginLoader(this))
 
   override fun createPluginDescriptorFinder(): PluginDescriptorFinder =
-    SpinnakerPluginDescriptorFinder()
+    SpinnakerPluginDescriptorFinder(this.getRuntimeMode())
 
   override fun loadPluginFromPath(pluginPath: Path): PluginWrapper? {
     val extractedPath = bundleExtractor.extractService(pluginPath, serviceName)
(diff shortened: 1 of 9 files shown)


cfieber:
I expect to build tooling in the plugin for spinnaker/spinnaker-gradle-project#88 to support automatic generation of `.plugin-ref` files, pushing them to a known location, and emitting a yaml configuration document that enables the specific plugin configuration within the host service.

As to the IDE setup, this might just be in the form of documentation.

kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/loaders/SpinnakerPluginLoaderTests.kt
```diff
@@ -1,106 +0,0 @@
-/*
```
cfieber: this feels like a rebase error, will investigate

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginManager.kt
```diff
@@ -64,14 +68,19 @@ open class SpinnakerPluginManager(
...
   override fun createPluginLoader(): PluginLoader =
     CompoundPluginLoader()
+      .add(PluginRefPluginLoader(this), this::isDevelopment)
       .add(SpinnakerDevelopmentPluginLoader(this), this::isDevelopment)
       .add(SpinnakerDefaultPluginLoader(this))
       .add(SpinnakerJarPluginLoader(this))
 
   override fun createPluginDescriptorFinder(): PluginDescriptorFinder =
-    SpinnakerPluginDescriptorFinder()
+    SpinnakerPluginDescriptorFinder(this)
```
jonsie: Maybe instead we pass in the `RuntimeMode` like so `SpinnakerPluginDescriptorFinder(runtimeMode)`.

Then in `SpinnakerPluginDescriptorFinder` we check that runtime mode equals `RuntimeMode.DEVELOPMENT`.  This way we're not passing around the `PluginManager`, we have the option to key off other runtime modes in `SpinnakerPluginDescriptorFinder`, and we still get to use the `DefaultPluginManager` implementation of `getRuntimeMode`.

jonsie:
This looks great and it makes me excited.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/loaders/PluginRefPluginLoader.kt
```diff
@@ -0,0 +1,44 @@
...
+import com.netflix.spinnaker.kork.plugins.SpinnakerPluginDescriptor
+import com.netflix.spinnaker.kork.plugins.pluginref.PluginRef
+import org.pf4j.BasePluginLoader
+import org.pf4j.PluginClasspath
+import org.pf4j.PluginDescriptor
+import org.pf4j.PluginLoader
+import org.pf4j.PluginManager
+import java.nio.file.Path
+
+open class PluginRefPluginLoader(val pluginManager: PluginManager) : PluginLoader {
```
robzienert: Why `open`?

robzienert: Would also like to see class-level docs on this, similar to the description you have in the commit.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginManager.kt
cfieber: that makes a lot more sense, will look at that

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/loaders/PluginRefPluginLoader.kt
cfieber: why open indeed?! (fixed)

also javadoc'd up the swath of things I had to extend to make `PluginRef` happen.

PR #468, opened 2020-01-15
feat(plugins): Support for extracting plugin bundles

robzienert:
Plugin bundles are an aggregate of one or more PF4J plugin zip artifacts. It provides an experience for plugin developers to produce a single plugin artifact that is applied to one or more services. A bundle would look something like this:

```
my-cool-plugin.zip
  orca.zip
  deck.zip
  ...
```

The `PluginBundleExtractor` provides the ability to extract a downloaded plugin bundle and source the correct internal plugin artifact for use by a service. It is also backwards compatible with a normal PF4J plugin artifact, in case a developer does not need bundle capabilities.

### DIFF
--- a/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/bundle/PluginBundleExtractor.kt
+++ b/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/bundle/PluginBundleExtractor.kt
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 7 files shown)


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/bundle/PluginBundleExtractor.kt
```diff
@@ -0,0 +1,71 @@
...
+ * Individual service plugins will be as what PF4J would normally expect.
+ */
+class PluginBundleExtractor {
+
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  /**
+   * Extract the bundle. This does not unpack any of the underlying plugin zips.
+   */
+  fun extractBundle(bundlePath: Path): Path {
```
jonsie: This could probably be private?  Or, maybe we could just remove this function and have one `extract` function.

robzienert: This method is going to be used (I think?) in the Gate integration, which is why it's public.

PR #471, opened 2020-01-16
fix(plugins): create plugins directory if not present

link108:
SpinnakerUpdateManager attempts to put plugins into a root directory, this causes the service to crash if the directory does not exist. This PR ensures the directory is present


robzienert:
Just verifying - when you say "put plugins into a root directory", you don't mean root filesystem directory, do you?

link108:
Ah no, its the pluginsRoot directory as defined here: https://github.com/spinnaker/kork/pull/471/files#diff-8b37015b7646e288eade5da2c5407e3aR52

I'll leave the wording so your comment makes sense, it is *not* the root directory 👍

PR #475, opened 2020-01-17
feat(plugins): Added privileged Spring plugin

claymccoy:
This allows plugins to wire themselves into the application’s Spring context by explicitly registering BeanDefinitions.
See https://github.com/claymccoy/springExamplePlugin for a working example and list of capabilities.
Note that unsafe mode in your plugin is not required for this.

I tried this the way that was intended by sharing the app's Spring context and it didn't load my plugin beans correctly.
https://github.com/spinnaker/kork/pull/454
I did some research about Spring context initialization and lifecycle and tried a lot of different things. The full application context doesn't seem to be fully ready at this point. So I tried using postProcessBeanFactory to create and register the plugin beans, but that didn't work quite right either. The recommended approach is using postProcessBeanDefinitionRegistry to register bean definitions and this seems to work great, though it forces you to be a bit more explicit. The example plugin linked above has docs and code demonstrating how this works.

### DIFF
--- a/kork-plugins-spring-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/spring/PrivilegedSpringPlugin.java
+++ b/kork-plugins-spring-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/spring/PrivilegedSpringPlugin.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2020 Armory, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.plugins.api.spring;
+
+import com.netflix.spinnaker.kork.annotations.Alpha;
+import org.pf4j.Plugin;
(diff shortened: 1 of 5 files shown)


robzienert:
FWIW, I think this can replace the `SpringPlugin` that exists today - would recommend deleting it, I think, when this goes in. Thoughts?

claymccoy:
Tests are not a problem. I just wanted to make sure this approach would work for everyone.
I was thinking SpringPlugins was more concerned with a plugin that had an isolated Spring Context and didn't want to share. I don't know if we have a use for that though.

cfieber:
very cool! agree this could replace the existing SpringPlugin

cfieber:
hmm now that you mention it I think SpringPlugin does still provide useful convenience for a nested isolated application context.

I wonder if there is some world where a hybrid of this change and SpringPlugin exists where you can build a nested application context, and selectively pluck beandefinitions out of it that should be registered into the host application's context

robzienert:
Yeah, they do provide different functionality. It might be useful to see if we could merge the two if we see the existing `SpringPlugin` functionality being useful.

claymccoy:
The nice thing about this approach is that the plugin still has an isolated classloader, but everything lives in the same SpringContext. I think that is ideal for internal plugins. It lets us hook into the existing Spinnaker services like we would in our extension projects. I think if we wanted to isolate our Spring context as well then we would be considering an external plugin or separate microservice.
I pointed out the difference in SpringPlugin and PrivilegedSpringPlugin mainly to understand when the isolated Spring context would be useful.  I don't currently see a use case for the isolated context, and I'm hesitant to overcomplicate PrivilegedSpringPlugin.

claymccoy:
@robzienert I've added tests that were in line with the existing tests for ExtensionBeanDefinitionRegistryPostProcessor and SpringPlugin. I think that is sufficient. I don't really want to test the Spring framework itself here.
In a production plugin I would have tests for the services, but this is just a contrived example that has no real logic worth testing.

PR #478, opened 2020-01-20
refactor(secrets/s3): s3 custom endpoint

jorgebee65:
In order to use another storage system like for example Minio, this change allows to set a custom endpoint url to the s3 secret configuration.

```
secrets:
  s3:
    enabled: true
    endpoint-url: <custom-endpoint>
    path-style-access-enabled: true
```


kork-secrets-aws/src/main/java/com/netflix/spinnaker/kork/secrets/engines/S3Configuration.java
```diff
@@ -0,0 +1,44 @@
...
+@Slf4j
+public class S3Configuration {
+
+  private String endpointUrl;
+  private boolean pathStyleAccessEnabled;
+
+  @Autowired
+  public S3Configuration(
+      @Value("${secrets.s3.endpoint-url}") String endpointUrl,
+      @Value("${secrets.s3.path-style-access-enabled}") boolean pathStyleAccessEnabled) {
```
robzienert: This class is odd: It's acting like a `@ConfigurationProperties`, but it isn't one.

We prefer to avoid the use of `@Value` annotations in favor of `@ConfigurationProperties` classes that get injected where they need to go. Would recommend refactoring to fall in line with this pattern.

robzienert: I'm also worried (having not looked at the whole module) that you have a `@ConditionalOnProperty` annotation here, but then use it directly in `S3SecretEngine`'s constructor, but have it typed as a required object. Depending on how the rest of the module is wired up, this could cause a runtime exception during startup.

kork-secrets-aws/src/main/java/com/netflix/spinnaker/kork/secrets/engines/S3Configuration.java
```diff
@@ -0,0 +1,44 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.stereotype.Component;
+
+@Getter
+@Component
+@Configuration
+@ConditionalOnProperty("secrets.s3.enabled")
+@Slf4j
```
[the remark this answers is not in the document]
jorgebee65: removed

kork-secrets-aws/src/main/java/com/netflix/spinnaker/kork/secrets/engines/S3Configuration.java
jorgebee65: S3Configuration changed according to the pattern. 
S3SecretEngine: change the initialization from the constructor to an Autowired (not required) attribute.

jorgebee65:
@robzienert changes applied, please review and let me know what you think.

kork-secrets-aws/kork-secrets-aws.gradle
```diff
@@ -6,4 +6,8 @@ dependencies {
   api project(':kork-secrets')
 
   implementation "com.amazonaws:aws-java-sdk-s3"
+
+  compileOnly 'org.projectlombok:lombok:+'
+
+  annotationProcessor 'org.projectlombok:lombok:+'
```
robzienert: Just do `org.projectlombok:lombok` so that `kork-bom` takes care of the lombok version.

kork-secrets-aws/src/main/java/com/netflix/spinnaker/kork/secrets/engines/S3Configuration.java
robzienert: Do not use property injection. If the property is not required, you can still use constructor injection, but you have to wrap the optional properties in `Optional`:

```java
class Foo {
  private final Optional<MyDep> myDep;

  Foo(Optional<MyDep> myDep) {
    this.myDep = myDep;
  }
}
```

jorgebee65: changed following the instructions.

kork-secrets-aws/kork-secrets-aws.gradle
jorgebee65: I just remove these lines and add: `"$rootDir/gradle/lombok.gradle"`

jorgebee65:
Hi @robzienert, changes applied according to your comments.

jorgebee65:
Hi @robzienert changes applied, could you please take a look?

PR #479, opened 2020-01-20
feat(namedthreadfactory): Add NamedThreadFactory

marchello2000:
We now use a NamedThreadFactory in orca and clouddriver, hoist it up in kork so that it's available to all without reinventing the wheel


marchello2000:
I will remove it from orca and clouddriver when this is merged/published

PR #480, opened 2020-01-20
fix(plugins): VersionRequirementsParser should throw InvalidPluginVersionRequirementException when SUPPORTS_PATTERN does not match

jonsie:
Supports https://github.com/spinnaker/front50/pull/690

### DIFF
--- a/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/VersionRequirementsParser.kt
+++ b/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/VersionRequirementsParser.kt
@@ -16,7 +16,6 @@
 package com.netflix.spinnaker.kork.plugins
 
 import com.netflix.spinnaker.kork.exceptions.UserException
-import java.lang.IllegalStateException
 import java.util.regex.Pattern
 
 /**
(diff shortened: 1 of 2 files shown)


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/VersionRequirementsParser.kt
```diff
@@ -43,10 +42,8 @@ object VersionRequirementsParser {
   fun parse(version: String): VersionRequirements {
     return SUPPORTS_PATTERN.matcher(version)
       .also {
-        try {
-          it.matches()
-        } catch (e: IllegalStateException) {
-          throw InvalidPluginVersionRequirementException(version, e)
+        if (!it.matches()) {
+          throw InvalidPluginVersionRequirementException(version)
```
jonsie: This way we throw `InvalidPluginVersionRequirementException` when the pattern does match `SUPPORTS_PATTERN`.

robzienert: Ha. Oops.

PR #483, opened 2020-01-21
fix(plugins): Fixes for SpinnakerPluginInfo, front50 update repository, and other minor issues

jonsie:


kork-plugins/src/main/java/com/netflix/spinnaker/config/Front50UpdateRepositoryConfiguration.java
```diff
@@ -60,6 +61,21 @@ public static UpdateRepository pluginFront50UpdateRepository(
...
+                    new BeanCreationException(
+                        "Unable to bind front50.base-url property to "
+                            + URL.class.getSimpleName()));
+
+    PluginRepositoryProperties front50RepositoryProps =
+        pluginRepositoriesConfig.get(PluginsConfigurationProperties.FRONT5O_REPOSITORY);
+
+    URL front50Url =
+        defaultFront50Url != null ? defaultFront50Url : front50RepositoryProps.getUrl();
+
```
jonsie: This basically ensures we have a decent configuration process such that if `front50.baseUrl` is already defined in the config the user does not need to add the url to `spinnaker.extensibility.repositories.front50.url`.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/SpinnakerPluginInfo.kt
```diff
@@ -16,12 +16,20 @@
...
+
+  @Suppress("UNCHECKED_CAST")
+  fun getReleases(): List<SpinnakerPluginRelease> {
+    return releases as List<SpinnakerPluginRelease>
+  }
+
+  @JsonSetter("releases")
+  fun setReleases(spinnakerReleases: List<SpinnakerPluginRelease>) {
+    releases = spinnakerReleases
+  }
```
jonsie: There are no getters or setters on the parent `PluginInfo` public fields, so I had to do this to support `SpinnakerPluginRelease`.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/SpinnakerUpdateManager.kt
```diff
@@ -85,21 +73,24 @@ class SpinnakerUpdateManager(
...
+  private fun Path.write(downloaded: Path): Boolean {
+    val file = this.resolve(downloaded.fileName)
+    File(this.toString()).mkdirs()
+    try {
+      return Files.move(downloaded, file, StandardCopyOption.REPLACE_EXISTING)
+        .contains(downloaded.fileName)
+    } catch (e: IOException) {
+      throw PluginRuntimeException(e, "Failed to write file '{}' to plugins folder", file)
+    }
+  }
```
jonsie: These changes ensure we only call `loadPlugins` and `startPlugins` once from `ExtensionBeanDefinitionRegistryPostProcessor`.  This is more efficient and I think easier to understand.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/front50/Front50Service.kt
```diff
@@ -29,8 +30,8 @@ import retrofit2.http.Query
 interface Front50Service {
 
   @GET("/pluginInfo/{id}")
-  fun getById(@Path("id") id: String): SpinnakerPluginInfo
+  fun getById(@Path("id") id: String): Call<SpinnakerPluginInfo>
 
   @GET("/pluginInfo")
-  fun list(@Query("service") service: String): Collection<SpinnakerPluginInfo>
+  fun list(@Query("service") service: String): Call<Collection<SpinnakerPluginInfo>>
```
jonsie: There may be a way around this, but I think it's the standard pattern with Retrofit2 to wrap the result in a `Call`.  Without this, we threw a Retrofit exception.

robzienert: This kind of leakiness is why I don't like Retrofit.

kork-plugins/src/main/java/com/netflix/spinnaker/config/Front50UpdateRepositoryConfiguration.java
```diff
@@ -73,21 +87,18 @@ public static UpdateRepository pluginFront50UpdateRepository(
...
+            .baseUrl(front50Url)
             .client(okHttpClient)
             .build()
             .create(Front50Service.class);
 
     return new Front50UpdateRepository(
         PluginsConfigurationProperties.FRONT5O_REPOSITORY,
-        applicationContext.getApplicationName(),
-        front50RepositoryProps.getUrl(),
+        Objects.requireNonNull(environment.getProperty("spring.application.name")),
```
jonsie: `applicationContext.getApplicationName()` was returning null.  It didn't seem to be reading `spring.application.name`, this approach works though.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/PluginUpdateService.kt
```diff
@@ -33,23 +39,16 @@ import java.nio.file.Path
  *
  * All plugins that are returned by the repository will be downloaded and installed.
  *
- * While it's possible that this could be used after the application has started up, there is no guarantee that the
- * extensions a plugin exposes will be correctly refreshed within the Spring context at this point. Therefore, it is
- * strongly advised that if a plugin update needs to occur, the service should be redeployed. If live updates are
- * desired in the future, we'll need to proxy extensions so that the real implementing object is delegated to, allowing
- * Spring injection to continue working.
```
jonsie: Loading and starting plugins is no longer possible outside the context of `ExtensionBeanDefinitionRegistryPostProcessor`, so this comment no longer applies.  I think it's clearer to have just one path for loading and running plugins, until we want to explicitly tackle the problem of reloading plugins at runtime and refreshing the Spring context.

PR #484, opened 2020-01-22
feat(annotations): Add annotations to default nullity

ezimanyi:
There is a ParametersAreNonnullByDefault annotation (and corresponding nullable) in javax.annotation that has gained some use in Spinnaker's microservices.  There is not, however, a corresponding annotation that allows the default for fields or methods to be set.

Add annotations to specify that all fields within the scope are Nonnull (or Nullable), and to specify that all methods within the scope reutrn Nonnull (or Nullable).

Finally, add a NonnullByDefault annotation to specify that all methods, fields, and parameters in the scope should be Nonnull unless explicitly annotated (and a corresponding NullableByDefault).

The primary use case I envison for the Nullable annotations is to reverse the effect of an annotation at a wider scope---perhaps an entire package is annotated NonnullByDefault by there is a particular
class whose fields should all be nullable.


ezimanyi:
I wavered back and forth around whether to add all combinations of the `@...NullableByDefault` annotation; I don't think we want to encourage this to be applied to wide swaths of our codebase, but I think that having these available to overrride a wider-scoped default might encourage adding the `@...NonnullByDefault` annotations in more places.

kork-annotations/src/main/java/com/netflix/spinnaker/kork/annotations/FieldsAreNonnullByDefault.java
```diff
@@ -0,0 +1,32 @@
...
+package com.netflix.spinnaker.kork.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import javax.annotation.Nonnull;
+import javax.annotation.meta.TypeQualifierDefault;
+
+/**
+ * Defaults all fields within annotation's scope to @Nonnull by default, unless a more
```
plumpy: "by default" is redundant... (×6)

ezimanyi: Good call, fixed.

robzienert:
This is neat. Anywhere that it's used yet?

ezimanyi:
Yes, a few places! I've started trying to add this to any class that I significantly touch (particularly if it already has a lot of the `@Nonnull` annotations individually added).  Examples:
* [ArtifactResolver](https://github.com/spinnaker/orca/blob/02cd0b800f19f5bcafaf6db830e152864a0789e6/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactResolver.java#L36)
* [ArtifactUtils](https://github.com/spinnaker/orca/blob/02cd0b800f19f5bcafaf6db830e152864a0789e6/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactUtils.java#L55)

It's even on a whole package here:
* [com.netflix.spinnaker.igor.gcb](https://github.com/spinnaker/igor/blob/e32f6a750917ee5f17cca6b7cfa64ec68eb152ac/igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/package-info.java#L17)

I'd love to see us try to make this the default for new classes/packages to force us to be explicit.

PR #487, opened 2020-01-24
refactor(plugins): Plugin version and service requirement selected via SpinnakerUpdateManager

jonsie:
This does a few things for us:
- Update repositories should return all plugin info objects, that includes the front50 repo.  Cache the world.  This keeps the process of determining which plugin to install unified, which leads me to...
- `SpinnakerUpdateManager.getLastPluginRelease` will be used to select the plugin version for installation.  There will be a follow up PR to Gate for this (for use in the `DeckPluginCache`). We will eventually refactor this, but at least it's all in one place.

### DIFF
--- a/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/PluginUpdateService.kt
+++ b/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/PluginUpdateService.kt
@@ -37,17 +37,16 @@ import java.nio.file.StandardCopyOption
  * The current behavior is that the latest plugin release (according to semver) will always be selected as the
  * most desirable plugin.
  *
- * All plugins that are returned by the repository will be downloaded and installed.
+ * Plugins that require the Spinnaker service will be downloaded and installed.
  *
  * Plugins will not be loaded or started from this service.  All plugin loading and starting occurs
  * via [com.netflix.spinnaker.kork.plugins.ExtensionBeanDefinitionRegistryPostProcessor].
  *
- * TODO(rz): Look to a local config or front50 (depending on spinnaker setup) to determine if there are specific
- *  plugin versions to use. We may not want the most recent release of a plugin.
  */
 class PluginUpdateService(
   internal val updateManager: SpinnakerUpdateManager,
   internal val pluginManager: SpinnakerPluginManager,
+  private val spinnakerServiceName: String,
   private val applicationEventPublisher: ApplicationEventPublisher
 ) {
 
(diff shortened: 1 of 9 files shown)


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/PluginUpdateService.kt
```diff
@@ -58,13 +57,6 @@ class PluginUpdateService(
     installNewPlugins()
   }
 
-  /**
-   * This is used in Gate to download Deck plugin artifacts - do not remove and must remain public.
-   */
-  fun download(pluginId: String, version: String): Path {
```
jonsie: Minor -- Going to switch Gate over to using the `SpinnakerUpdateManager` `downloadPluginRelease` since it has a dependency on `UpdateManager` already, and then we can remove `PluginUpdateService` as a dependency in `DeckPluginCache`.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/PluginUpdateService.kt
```diff
@@ -74,10 +66,12 @@ class PluginUpdateService(
     val updates = updateManager.updates
     log.debug("Found {} plugin updates", updates.size)
 
-    updates.forEach { plugin ->
+    updates.forEach updates@{ plugin ->
```
robzienert: What does this marker do? I've never seen that before.

robzienert: (Seen it on return statements, but not this)

robzienert: Ohh... you name scopes and return from that in Kotlin. Yeah, nice.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/front50/Front50Service.kt
```diff
@@ -33,5 +33,8 @@ interface Front50Service {
   fun getById(@Path("id") id: String): Call<SpinnakerPluginInfo>
 
   @GET("/pluginInfo")
-  fun list(@Query("service") service: String): Call<Collection<SpinnakerPluginInfo>>
+  fun listByService(@Query("service") service: String): Call<Collection<SpinnakerPluginInfo>>
```
robzienert: We don't use this endpoint anymore, do we? Why not just remove it?

jonsie: Will do.

PR #489, opened 2020-01-27
chore(api): make AuthenticatedRequest::getSpinnakerApplication public

dreynaud:
Seems like this was made private by mistake and I intend to use it in gate.


marchello2000:
👍

PR #490, opened 2020-01-27
feat(auth): expand AuthenticatedRequest API

dreynaud:
Quality of life change that adds nice getters and setters in AuthenticatedRequest. With this change, the MDC updates and lookups are encapsulated in AuthenticatedRequest and don't leak out in AuthenticatedRequestFilter or any other caller.

Also factor out AuthenticatedRequest::clear() to reduce some duplication.

### DIFF
--- a/kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequest.java
+++ b/kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequest.java
@@ -67,6 +67,11 @@ public class AuthenticatedRequest {
     public static String makeCustomHeader(String header) {
       return XSpinnakerPrefix + header.toUpperCase();
     }
+
+    @Override
+    public String toString() {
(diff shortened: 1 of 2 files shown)


kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequest.java
```diff
@@ -217,23 +196,74 @@ private static void setOrRemoveMdc(String key, String value) {
...
+  public static void setExecutionId(String value) {
+    set(Header.EXECUTION_ID, value);
+  }
+
+  public static void setApplication(String value) {
+    set(Header.APPLICATION, value);
+  }
+
+  public static void set(Header header, String value) {
+    set(header.toString(), value);
```
marchello2000: will `.toString()` result in same as `.getHeader()`

dreynaud: nope, good catch

PR #492, opened 2020-01-28
feat(selector): make SelectableService generic

dreynaud:


kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/SelectableService.java
```diff
@@ -19,14 +19,14 @@
 import java.util.List;
 import org.springframework.util.Assert;
 
-public class SelectableService {
-  private final List<ServiceSelector> serviceSelectors;
+public class SelectableService<T> {
```
marchello2000: yes! I was slightly unhappy with it being an object :)

PR #493, opened 2020-01-29
feat(context): provide a RequestContext abstraction

dreynaud:
This PR lifts Header out of AuthenticatedRequest and moves it into kork-core so that it can be accessed from both kork-web and kork-security.

In kork-web, we introduce a RequestContext interface with a concrete implementation that pulls values from AuthenticatedRequest. Consumers that need to access values from a RequestContext just need to get an autowired RequestContextProvider.

This should make code that depends on these values to be more easily testable and make it easier to go from one type of context to another (as I am currently doing in gate).

### DIFF
--- a/kork-core/src/main/java/com/netflix/spinnaker/kork/common/Header.java
+++ b/kork-core/src/main/java/com/netflix/spinnaker/kork/common/Header.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.kork.common;
+
+/**
+ * Known X-SPINNAKER headers, but any X-SPINNAKER-* key in the MDC will be automatically propagated
+ * to the HTTP headers.
+ *
+ * <p>Use makeCustomerHeader() to add customer headers
+ */
+public enum Header {
+  USER("X-SPINNAKER-USER", true),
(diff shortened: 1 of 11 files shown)


kork-core/src/main/java/com/netflix/spinnaker/kork/dynamicconfig/SpringDynamicConfigService.java
```diff
@@ -87,6 +87,6 @@ public void setEnvironment(Environment environment) {
   }
 
   private static String flagPropertyName(String flagName) {
-    return format("%s.enabled", flagName);
+    return flagName.endsWith(".enabled") ? flagName : format("%s.enabled", flagName);
```
dreynaud: I forgot that I snuck this in there, I have been tripped by the unconditional appending of `".enabled"` which I did not expect. Seems reasonable enough?

kork-core/src/main/java/com/netflix/spinnaker/kork/common/Header.java
```diff
@@ -0,0 +1,62 @@
...
+
+package com.netflix.spinnaker.kork.common;
+
+/**
+ * Known X-SPINNAKER headers, but any X-SPINNAKER-* key in the MDC will be automatically propagated
+ * to the HTTP headers.
+ *
+ * <p>Use makeCustomerHeader() to add customer headers
+ */
+public enum Header {
```
dreynaud: no changes in this class, it's a pure move from `AuthenticatedRequest`

kork-web/src/main/java/com/netflix/spinnaker/kork/web/context/RequestContext.java
```diff
@@ -0,0 +1,95 @@
...
+  default Optional<String> getExecutionType() {
+    return get(Header.EXECUTION_TYPE);
+  }
+
+  default Optional<String> get(Header header) {
+    return get(header.getHeader());
+  }
+
+  // setters
+  default void setAccounts(String accounts) {
```
marchello2000: nit: 
```suggestion
  default void setAccounts(String value) {
```

kork-web/src/main/java/com/netflix/spinnaker/kork/web/context/RequestContext.java
```diff
@@ -0,0 +1,95 @@
...
+    set(Header.ACCOUNTS, accounts);
+  }
+
+  default void setUser(String user) {
```
marchello2000: nit:
```suggestion
  default void setUser(String value) {
```
to make consistent with others..? 🤷‍♂

PR #495, opened 2020-02-01
fix(plugins): check classes/META-INF/ for MANIFEST.MF before other lo…

link108:
…cations
Different plugin manifest files are found when running locally vs running in a container. This is due to the PF4J function, [FileUtils.findFile](https://github.com/pf4j/pf4j/blob/master/pf4j/src/main/java/org/pf4j/util/FileUtils.java#L243), returning different files. This PR checks to see if there is a manifest file in `/classes/META-INF/` before falling back to default functionality.

### DIFF
--- a/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/finders/SpinnakerManifestPluginDescriptorFinder.kt
+++ b/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/finders/SpinnakerManifestPluginDescriptorFinder.kt
@@ -20,6 +20,8 @@ import com.netflix.spinnaker.kork.plugins.validate
 import org.pf4j.DefaultPluginDescriptor
 import org.pf4j.ManifestPluginDescriptorFinder
 import org.pf4j.PluginDescriptor
+import java.io.File
+import java.nio.file.Path
 import java.util.jar.Manifest
 
 /**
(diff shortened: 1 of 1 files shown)


cfieber:
what specifically is the difference when running local vs in-container?

what is the file structure or plugin file type you are using locally and what does it look like in your container file system?

link108:
@cfieber I see the same file structure when running locally and in the container:
```

**Comment by link108** [comment]
@cfieber I've been testing using a PF4J extension-point based plugin, https://github.com/spinnaker-plugin-examples/pf4jStagePlugin, for testing.

I see the same file structure running locally and in the container (plugins root directory is different) eg:
```
tree orca-web/plugins/Armory.RandomWaitPlugin-1.0.0

orca-web/plugins/Armory.RandomWaitPlugin-1.0.0
├── META-INF
│   └── MANIFEST.MF
├── classes
│   ├── META-INF
│   │   ├── MANIFEST.MF
│   │   ├── extensions.idx
│   │   └── pf4jStagePlugin.kotlin_module
│   └── io
│       └── armory
│           └── plugin
│               └── stage
│                   └── wait
│                       └── random
│                           ├── Context.class
│                           ├── ...

└── lib
    ├── annotations-13.0.jar
    ├── ...

```

I discovered the cause of this issue by modifying Kork to print out the manifest path in `SpinnakerManifestPluginDescriptorFinder` and saw it return `/classes/META-INF/MANIFEST.MF` when running locally, but returns `/META-INF/MANIFEST.MF` when running on the container. 

PF4J's [getManifestPath()](https://github.com/pf4j/pf4j/blob/master/pf4j/src/main/java/org/pf4j/ManifestPluginDescriptorFinder.java#L90) function uses another PF4J function, [FileUtils.findFile()](https://github.com/pf4j/pf4j/blob/master/pf4j/src/main/java/org/pf4j/util/FileUtils.java#L243) for finding a file, which in turn uses [File.listFiles()](https://docs.oracle.com/javase/7/docs/api/java/io/File.html#listFiles())  to get a list of files. That call, `File.listFiles()` does not return files in a specific order, which I think is causing the issue

cfieber:
why is there a top level `META-INF/MANIFEST.MF` in there?

cfieber:
ahh, I think the issue is this:

https://github.com/spinnaker-plugin-examples/pf4jStagePlugin/blob/master/build.gradle#L62

that should be a Zip task not a Jar task, in which case the archive won't contain a top level MANIFEST

link108:
was able to confirm that is the issue, good catch 👍

PR #497, opened 2020-02-04
fix(plugins): Check multiple default property locations for front50 base-url

jonsie:
Kind of annoying this is a problem but I'd rather not spread unnecessary configuration around.


kork-plugins/src/main/java/com/netflix/spinnaker/config/Front50UpdateRepositoryConfiguration.java
```diff
@@ -58,14 +58,15 @@ public static UpdateRepository pluginFront50UpdateRepository(
...
             .bind("front50.base-url", Bindable.of(URL.class))
-            .orElseThrow(
-                () ->
-                    new BeanCreationException(
-                        "Unable to bind front50.base-url property to "
-                            + URL.class.getSimpleName()));
+            .orElse(
+                Binder.get(environment)
+                    .bind("services.front50.base-url", Bindable.of(URL.class))
+                    .get());
```
jonsie: If not found, this'll throw a `NoSuchElementException` with the message `No value bound` which I think is legit.

PR #501, opened 2020-02-06
fix(swagger): Exclude MetaClass from swagger model generation

ezimanyi:
All groovy classes have an implicit metaClass field that handles metaprogramming. In cases where a groovy class is used in an API the springfox library to generate Swagger docs is including the metaClass in the model for the class. In addition to being confusing (as a producer/consumer of the API would never set that field), it also leads to terrible performance in some cases.

The performance issues come from the fact that it then tries to build a model of MetaClass by reflecting its fields, etc. until it has built models of most of the Groovy classes.

In particular, before this change, front50 took 140 seconds to start; with this change it now takes 13 seconds.


kork-swagger/kork-swagger.gradle
```diff
@@ -6,4 +6,5 @@ dependencies {
   api "org.springframework.boot:spring-boot-autoconfigure"
   api "io.springfox:springfox-swagger2"
   api "io.springfox:springfox-swagger-ui"
+  api "org.codehaus.groovy:groovy-all"
```
ezimanyi: I don't love needing to pull in groovy as a dependency just to exclude a groovy class from model building...open to suggestions for a better way of doing this.  (I think it needs to be `api` as if this isn't on the classpath of the consuming service at runtime I think it will fail if it looks at the list of excluded types and finds an undefined class.)

One other thought I had was to define a bean of type `List<Class> excludedSwaggerClasses()` here that's empty, and leave it to services to override that bean with anything they want to exclude...given that we'd probably want this in almost all services (as they almost all have groovy) that seemed more verbose, but definitely open to thoughts on that.

robzienert: I'm fairly certain we can change this to `implementation` - at least at that point we're not forcing groovy as a dependency on everything else that uses this module.

ezimanyi: Good call, let me quickly test that and will update the PR if it works.

robzienert: If this doesn't work, I'd be happy to volunteer being the grunt that converts gate's controllers to Java.

ezimanyi: I just tested and it does indeed work using an `implementation` dependency though unfortunately (which I probably already knew at some point) that does mean we're still pulling Groovy into the runtime classpath of any consuming services.  It's not on the compile classpath (which is the difference from `api`) so at least these services won't be able to directly call groovy code.

ezimanyi: I noticed this in `front50` in particular as 90% of the startup time was this swagger generation; looks like `clouddriver` also uses it, probably others too.

Let me know if you think this is reasonable as `implementation` even though it will put `groovy` on the runtime classpath of java-only dependencies.

robzienert: I think that's reasonable. We can improve the swagger experience now and then once Groovy is removed from our controllers, we'll be able to remove this dep. I've updated the TOC board to include this task: https://github.com/orgs/spinnaker/projects/6

This specific refactor was something I was targeting as next for me to take from the TOC project board anyway.

PR #503, opened 2020-02-06
feat(telemetry): add deployment method and version to kork-proto

dogonthehorizon:
We would like to be able to track how users are installing Spinnaker and what version of those tools they're using. One example would be tracking which versions of halyard people are using to install Spinnaker in their environments.

We have some related work to do in a PR to echo to start propagating this information.


ttomsu:
Have you cleared this extra data with LF legal?

Chris-Tan:
Hey Travis! This was approved by LF on 2020-01-09.

Screen Shot 2020-02-07 at 1 08 04 PM

Screen Shot 2020-02-07 at 1 04 24 PM

PR #506, opened 2020-02-06
feat(plugins): Constructor dependency injection for extensions

robzienert:
This refactor changes how system dependencies are injected from a service into an extension (and likely, eventually, a plugin). Instead of using method-based injection, this PR refactors the `SpringExtensionFactory` (which is a misnomer) to not actually do the initialization of an extension, but passes it off to a new `ExtensionInitializer` interface, for which there is only one implementation at the moment: `ConstructorExtensionInitializer`.

An `ExtensionInitializer` works in tandem with `DependencyProvider`s, defined within `kork-plugins`, to inject dependencies into an extension. For now, this is only for configuration, but an upcoming PR would be to also inject a `HttpClientProvider`, which provides a generic (and dependency-less) HTTP client for plugins to use that has its logic internalized to the service itself.

Most of this PR is shuffling code around, rather than actually adding new code, but here we are anyway.


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpringExtensionFactory.kt
```diff
@@ -16,149 +16,49 @@
...
+    val extension = injector.createInstance(extensionClass, pluginWrapper)
 
-    loadConfig(extension, PluginConfigCoordinates(
-      coordinates.id,
-      annot.id
-    ))
+    if (pluginWrapper != null) {
+      // If the plugin is a SpringPlugin, we'll create an [ApplicationContext] for it and autowire the extension
+      val plugin = pluginWrapper.plugin
+      if (plugin is SpringPlugin) {
```
jonsie: Wait - you're removing SpringPlugin in #508

kork-plugins-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/Injectable.java
```diff
@@ -0,0 +1,42 @@
...
+ *   public MyExtension(HttpClientProvider httpClientProvider) {
+ *     this.httpClientProvider = httpClientProvider;
+ *   }
+ * }
+ * }</pre>
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.CLASS)
+@Documented
+public @interface Injectable {}
```
jonsie: I'm not actually seeing this used anywhere.  What are the plans for this?

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpringExtensionFactory.kt
robzienert: Yes, but it still existed when I was writing this. :)

kork-plugins-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/Injectable.java
robzienert: You're right, it's not used in this PR. It would be used for SDKs, such as `HttpClient` and so-on. It's a marker annotation that informs plugin developers what is available for injection, and could be used to auto-generate docs on what is available. It's possible that we could tie it back into the `DependencyProvider`s, but I'm not sure if there's value at this point.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpringExtensionFactory.kt
```diff
@@ -16,149 +16,49 @@
...
 import org.springframework.core.env.StandardEnvironment
-import java.lang.reflect.InvocationTargetException
 
 /**
  * TODO(rz): Support creation of unsafe plugins
  */
 class SpringExtensionFactory(
   private val pluginManager: SpinnakerPluginManager,
-  private val configResolver: ConfigResolver
+  private val injector: ExtensionInitializer
```
jonsie: Can we call this variable `extensionInitializer` or `initializer` instead?

kork-plugins/src/main/java/com/netflix/spinnaker/config/PluginsAutoConfiguration.java
```diff
@@ -86,10 +92,10 @@ public static ConfigResolver springEnvironmentConfigResolver(
   public static SpinnakerPluginManager pluginManager(
       PluginStatusProvider pluginStatusProvider,
       ApplicationContext applicationContext,
-      ConfigResolver configResolver) {
+      ExtensionInitializer extensionExtensionInitializer) {
     return new SpinnakerPluginManager(
         pluginStatusProvider,
-        configResolver,
+        extensionExtensionInitializer,
```
jonsie: `extensionExtensionInitializer` -- probably a typo.

kork-plugins/src/main/java/com/netflix/spinnaker/config/PluginsAutoConfiguration.java
```diff
@@ -74,6 +74,12 @@ public static ConfigResolver springEnvironmentConfigResolver(
     return new SpringEnvironmentConfigResolver(environment);
   }
 
+  @Bean
+  ExtensionInitializer extensionsInjector(ConfigResolver configResolver) {
```
jonsie: It wont be long before we build up another `DependencyProvider`, so let's make the `DependencyProvider` objects separate beans and then add them to this `extensionInjector` factory so we keep the arguments clean (i.e. -- `ConfigResolver` is a dependency for `ConfigDependencyProvider` and not `ExtensionInitializer`).

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/inject/ConstructorExtensionInitializer.kt
```diff
@@ -0,0 +1,81 @@
...
+package com.netflix.spinnaker.kork.plugins.inject
+
+import com.netflix.spinnaker.kork.exceptions.IntegrationException
+import org.pf4j.PluginWrapper
+import org.slf4j.LoggerFactory
+import java.lang.reflect.Constructor
+import java.lang.reflect.InvocationTargetException
+
+/**
+ * Initializes
```
jonsie: Unfinished document.

robzienert: Uh, nope. That's all you need to know. 🙄

jonsie:
Thinking more about this -- I wonder if the this constructor pattern is the best way to inject dependencies for extensions, my concern (stated above) is that it's not clear and prone to user error.  Perhaps we want to be explicit about what dependencies exist and provide some mechanism get them (like a `SpinnakerPluginContext` object or something)?  Just a thought.

srekapalli:
Also, to add to what @jonsie mentioned,  we might end with quite a number of provider impls and constructor injection will be bit difficult to maintain. Instead our Plugin SDK will supply the rich context object and from a plugin dev perspective that is the only thing they inject and that will have access methods to the hosting spinnaker app aspects ( such as service clients, loggers, metric clients , authentication contexts, execution contexts ).

robzienert:
A singular object to get various SDK services sounds like a better solution. I assume this would just be injected via constructor all of the time and people can just use what they want. That said, we'll still need to have `DependencyProvider`, so that we can encapsulate the creation of these various SDK services, but at least we won't need to maintain a special constructor injector.

I'll update the PR so that this change is reflected. Will ping the PR again once I have updated with the new strategy.

robzienert:
Ah, I forgot one wrinkle:

I've moved extension configuration into constructor injection. I still want this, because it'll simplify the class hierarchy of extension (no need for a `ConfigurableExtension<T>` interface). That'll mean that plugins will have two options for constructors:

```kotlin
class MyExtension(val services: PluginServices)

class MyExtension(val services: PluginServices, val config: MyExtensionConfig)
class MyExtensionConfig
```

If the constructors are anything but those two options, then 🔥 .

robzienert:
Shifted things around:

- Only two things can be injected now: An `SdkServices` object and a configuration object.
- An `SdkFactory` is needed for every "SDK service" that is offered via `SdkServices`. We still need to have factories for whatever we're providing into plugins.
- There's no more `DependencyProvider`. Most of its purpose was moved into `SdkFactory`, but configuration is treated as its own unique thing now.
- There's no more extensible story for extension initialization. Renamed `s/SpringExtensionFactory/SpinnakerExtensionFactory`, I moved all of the constructor injection stuff back into that class.

Overall, I think this PR is a lot simpler now and is improved by having a single frontdoor dependency into whatever future SDKs we provide.

PR #507, opened 2020-02-07
feat(plugins): HTTP client plugin API

robzienert:
The first actual API for `kork-plugins-api` and one that will be necessary pretty quick here as we at Netflix start getting closer to moving our internal extensions into plugins. This HTTP client is designed to allow plugins to make HTTP requests without directly wiring up & configuring their own HTTP clients, per-se.

An extension could have a constructor dependency on `HttpClientProvider` and with #506, it would be injected and the implementation of the `HttpClient`s themselves would be encapsulated within the service internals, allowing us to iterate the client without forcing transitive dependencies on plugin developers.

**ACTION**: Review the API. I went with _simple_ being the primary guiding light. I explicitly decided not to follow in a pattern like Retrofit despite its nice API because it's not very simple and we'd be stuck with that pattern for a long time. We can always make the API more complex, but it'll be hard to make it simpler as time goes on. Does the API do the trick? Does it make sense?

If we like this, I'll add class docs, tests and make it work and so-on.

### DIFF
--- a/kork-plugins-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/httpclient/Response.java
+++ b/kork-plugins-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/httpclient/Response.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.plugins.api.httpclient;
+
+import com.netflix.spinnaker.kork.annotations.Beta;
+import java.io.InputStream;
+import java.util.Map;
+import java.util.Optional;
+import javax.annotation.Nonnull;
+
+/** HTTP Response wrapper. */
+@Beta
+public interface Response {
+
+  /** Get the response body. */
+  InputStream getBody();
+
+  /** Get the response body, casting it into the expected type. */
+  <T> T getBody(@Nonnull Class<T> expectedType);
+
+  /** Get the error exception, if present. */
+  @Nonnull
+  Optional<Exception> getException();
+
+  /** Get the response code. */
+  int getStatusCode();
+
+  /** Get the response headers. */
+  @Nonnull
+  Map<String, String> getHeaders();
+
+  /**
+   * Returns whether or not the request had an error (4xx/5xx response code or underlying
+   * IOException)
+   */
+  default boolean isError() {
+    return getException().isPresent();
+  }
+}
--- a/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/sdk/httpclient/HttpClientSdkFactory.kt
+++ b/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/sdk/httpclient/HttpClientSdkFactory.kt
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.plugins.sdk.httpclient
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.netflix.spinnaker.config.OkHttp3ClientConfiguration
+import com.netflix.spinnaker.kork.plugins.sdk.IdResolver
+import com.netflix.spinnaker.kork.plugins.sdk.SdkFactory
+import com.netflix.spinnaker.kork.plugins.sdk.httpclient.internal.CompositeOkHttpClientFactory
+import org.pf4j.PluginWrapper
+import org.springframework.core.env.Environment
+
+/**
+ * Creates [Ok3HttpClientRegistry]s for a given extension.
+ */
+class HttpClientSdkFactory(
+  private val okHttpClientFactory: CompositeOkHttpClientFactory,
+  private val environment: Environment,
+  private val objectMapper: ObjectMapper,
+  private val okHttp3ClientConfiguration: OkHttp3ClientConfiguration
+) : SdkFactory {
+  override fun create(extensionClass: Class<*>, pluginWrapper: PluginWrapper?): Any {
+    return Ok3HttpClientRegistry(
+      IdResolver.pluginOrExtensionId(extensionClass, pluginWrapper),
+      environment,
+      objectMapper,
+      okHttpClientFactory,
+      okHttp3ClientConfiguration
+    )
+  }
+}
--- a/kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttpClientComponents.groovy
+++ b/kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttpClientComponents.groovy
@@ -26,8 +26,10 @@ import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
 
+import javax.inject.Provider
+
 @Configuration
-@EnableConfigurationProperties(OkHttpClientConfigurationProperties)
+@EnableConfigurationProperties([OkHttpClientConfigurationProperties, OkHttpMetricsInterceptorProperties])
 class OkHttpClientComponents {
   @Bean
   SpinnakerRequestInterceptor spinnakerRequestInterceptor(OkHttpClientConfigurationProperties okHttpClientConfigurationProperties) {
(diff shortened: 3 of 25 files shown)


kork-plugins-api/kork-plugins-api.gradle
```diff
@@ -15,10 +15,12 @@
  */
 
 apply plugin: "java-library"
+apply from: "$rootDir/gradle/kotlin-test.gradle"
 
 dependencies {
   api(platform(project(":spinnaker-dependencies")))
 
+  api "javax.annotation:javax.annotation-api"
```
robzienert: I'm adding a new dependency for the `kork-plugins-api`. I think having `@Nonnull` and so-on are pretty valuable, and this dependency changes once in never.

kork-plugins-api/src/test/kotlin/com/netflix/spinnaker/kork/plugins/httpclient/HttpClientIntegrationTest.kt
```diff
@@ -0,0 +1,40 @@
...
+import io.mockk.mockk
+
+class HttpClientIntegrationTest : JUnit5Minutests {
+
+  fun tests() = rootContext {
+    test("wip - dsl") {
+      val clientProvider: HttpClientProvider = mockk(relaxed = true)
+
+      val front50 = clientProvider.getInternal("front50")
+
```
srekapalli: Prob. We can get the response Object straight back from the HTTP method and throw a error/exception back for unsuccessful calls. ?

kork-plugins-api/src/test/kotlin/com/netflix/spinnaker/kork/plugins/httpclient/HttpClientIntegrationTest.kt
```diff
@@ -0,0 +1,40 @@
...
+import dev.minutest.junit.JUnit5Minutests
+import dev.minutest.rootContext
+import io.mockk.mockk
+
+class HttpClientIntegrationTest : JUnit5Minutests {
+
+  fun tests() = rootContext {
+    test("wip - dsl") {
+      val clientProvider: HttpClientProvider = mockk(relaxed = true)
+
```
srekapalli: How about `getInternalServiceClient` instead of `getInternal` ?

kork-plugins-api/src/test/kotlin/com/netflix/spinnaker/kork/plugins/httpclient/HttpClientIntegrationTest.kt
ajordens: I suspect there’s value in by-default making it easy to access response headers, status codes, etc. which I believe this is supporting. 

(Versus just the object)

kork-plugins-api/src/test/kotlin/com/netflix/spinnaker/kork/plugins/httpclient/HttpClientIntegrationTest.kt
```diff
@@ -0,0 +1,40 @@
...
+      val front50 = clientProvider.getInternal("front50")
+
+      val response = front50.get(Request("getGate", "/v2/applications/gate"))
+
+      val app = response.getPayload(Application::class.java)
```
ajordens: getBody(clazz) might be more idiomatic.

kork-plugins-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/httpclient/Response.java
```diff
@@ -0,0 +1,43 @@
...
+  Optional<Exception> getException();
+
+  /** Get the response code. */
+  int getStatusCode();
+
+  /** Get the response headers. */
+  Map<String, String> getHeaders();
+
+  /** Returns whether or not the request had an error. */
+  default boolean isError() {
```
ajordens: Had an error response body or just returned a 4xx/5xx?

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/httpclient/Ok3HttpClientProvider.kt
```diff
@@ -0,0 +1,94 @@
...
+      Ok3HttpClient("$pluginId.$name", baseUrl, okClient, objectMapper)
+    }
+  }
+
+  override fun get(name: String): HttpClient {
+    return clients["$pluginId.$name"] ?: throw IntegrationException("No client configured for '$name'")
+  }
+
+  override fun getInternal(name: String): HttpClient {
+    val baseUrl = applicationContext.environment.getProperty("${name.toLowerCase()}.baseUrl")
```
ajordens: Some of the existing services have this nested within services.BLAH.baseUrl. 

We going to try and make that consistent or ...?

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/httpclient/Ok3HttpClientProvider.kt
```diff
@@ -0,0 +1,94 @@
...
+
+              trustStore = config.security.trustStorePath.toFile()
+              trustStoreType = config.security.trustStoreType
+              trustStorePassword = config.security.trustStorePassword
+
+              tlsVersions = config.security.tlsVersions
+              cipherSuites = config.security.cipherSuites
+            }
+          },
+          applicationContext.getBean(OkHttp3MetricsInterceptor::class.java)
```
ajordens: Cool beans, having the normal set of metrics will be handy. 

Do we think there would be any value in an additional tag to indicate the originating pluginId?

ajordens:
No big concerns about the direction — this will fill a gap and let us port some of existing extensions which is a big 👍

ajordens:
Thoughts on how we would (internally) support calls to metatron-protected resources? We don’t really want to expose the underlying certificates. We may possibly want to proxy through `hook`?

kork-plugins-api/src/test/kotlin/com/netflix/spinnaker/kork/plugins/httpclient/HttpClientIntegrationTest.kt
robzienert: Yes, the extra information around a response is important (and in the case of a 204 response, the only stuff to action on).

kork-plugins-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/httpclient/Response.java
robzienert: Just 4xx/5xx. I don't want to build any smarts beyond the bare minimum on the SDK, since that would make it more coupled to our current idea of what an error response is (which varies service-to-service today).

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/httpclient/Ok3HttpClientProvider.kt
robzienert: Oh yeah, I forgot about that inconsistency. I'll update the code to take that into account, but yes, there's strong desire - from me at least - to standardize / simplify our configs.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/httpclient/Ok3HttpClientProvider.kt
robzienert: Oh, yes, that's a good call. I'll work on getting that squeezed in, too!

robzienert:
Re: Metatron 🤔 -- What `hook` are you referring to? I was originally thinking people would just use `configure` to setup Metatron endpoints, likely with a custom `MetatronHttpConfig` that we would publish (new `netflix.spinnaker-plugins-api` artifact) that would have all of the goodies set for them, but still allow modification of timeouts, etc. WDYT?

kork-plugins-api/src/test/kotlin/com/netflix/spinnaker/kork/plugins/httpclient/HttpClientIntegrationTest.kt
jonsie: +1

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/sdk/httpclient/HttpClientSdkFactory.kt
```diff
@@ -0,0 +1,44 @@
...
+ * Creates [Ohc3HttpClientRegistry]s for a given extension.
+ */
+class HttpClientSdkFactory(
+  private val okHttpClientFactory: CompositeOkHttpClientFactory,
+  private val environment: Environment,
+  private val objectMapper: ObjectMapper,
+  private val okHttp3ClientConfiguration: OkHttp3ClientConfiguration
+) : SdkFactory {
+  override fun create(extensionClass: Class<*>, pluginWrapper: PluginWrapper?): Any {
+    return Ohc3HttpClientRegistry(
```
jonsie: Oh my goodness c-3PO!

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/sdk/httpclient/Ohc3Response.kt
```diff
@@ -0,0 +1,49 @@
...
+package com.netflix.spinnaker.kork.plugins.sdk.httpclient
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.netflix.spinnaker.kork.plugins.api.httpclient.Response
+import okhttp3.ResponseBody
+import java.io.InputStream
+import java.lang.Exception
+import java.util.Optional
+
+class Ohc3Response(
```
jonsie: Going deep with the rename, I like it.

robzienert: It was getting confusing having so many various `OkHttpClient` various classes. It's not great, but it was the least bad.

jonsie: What about just calling it `Ok3HttpClient`? Instead of `Ohc3HttpClient`.

kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttpClientComponents.groovy
```diff
@@ -36,17 +38,17 @@ class OkHttpClientComponents {
 
   @Bean
   OkHttpMetricsInterceptor okHttpMetricsInterceptor(
-    Registry registry,
-    @Value('${ok-http-client.interceptor.skip-header-check:false}') boolean skipHeaderCheck) {
+    Provider<Registry> registry,
```
jonsie: I wonder if we should use `ObjectProvider` instead of `Provider`?  I have run into situations wherein `Provider` was not on the classpath (haven't actually tracked down exactly where this gets pulled in, but last I checked it was not part of `kork-plugins`).  `ObjectProvider` is the modern Spring way, as far as I can tell, and I've used it before for this purpose - https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/ObjectProvider.html

robzienert: `Provider` is part of `javax.inject` - it's definitely on the classpath. I don't really see the value in using `ObjectProvider`. It's certainly fancier, but I don't need any of the methods that it uses.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/sdk/httpclient/Ohc3Response.kt
robzienert: Ah, good suggestion - changed.

ajordens:
> Re: Metatron 🤔 -- What `hook` are you referring to? I was originally thinking people would just use `configure` to setup Metatron endpoints, likely with a custom `MetatronHttpConfig` that we would publish (new `netflix.spinnaker-plugins-api` artifact) that would have all of the goodies set for them, but still allow modification of timeouts, etc. WDYT?

We've got an ability within webhooks that overrides the scheme and encode the metatron application name.

So long as we're able to eliminate a lot of the boilerplate, 👍.

kork-plugins-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/httpclient/Response.java
ajordens: At some point I thought I remembered this looking at a payload vs just status codes ... just 4xx/5xx makes sense to me tho!

robzienert:
Ah, overriding the scheme is a good idea. I'll let this linger a little longer and looked to use that strategy instead - I think it may simplify a couple spots.

PR #516, opened 2020-02-14
feat(pubsub): initial port of kork-pubsub

dreynaud:
This is mostly a port of echo's pubsub support, with some additional SNS support coming from clouddriver.

There are minor differences with some classes from echo, but overall the idea is that one can configure a subscription (topic + queue) through Spring these classes will:
- parse the list of subscriptions from the configuration
- make sure the topics and queues for each subscription exist
- start a polling thread to read messages from each queue
- hand over messages to the handlers registered by the application


kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/config/AmazonPubsubProperties.java
```diff
@@ -0,0 +1,60 @@
...
+    @NotEmpty private String queueARN;
+
+    private List<String> accountIds = Collections.emptyList();
+
+    int visibilityTimeout = 30;
+    int sqsMessageRetentionPeriodSeconds = 120;
+    int waitTimeSeconds = 5;
+    int maxNumberOfMessages = 1;
+
+    // TODO: why the constructors?
```
emjburns: There might be something weird with having an args constructor and not having a no args constructor and serialization... I don't remember.

kork-pubsub-aws/kork-pubsub-aws.gradle
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+apply plugin: 'java-library'
+apply plugin: "groovy"
```
marchello2000: do you need groovy here? (is it just for tests?)

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/SNSPublisher.java
```diff
@@ -0,0 +1,120 @@
...
+      return Optional.of(publishResponse);
+    } catch (Exception e) {
+      log.error("failed to publish message {} to topic {}", message, topicARN, e);
+      getErrorCounter(e).increment();
+      return Optional.empty();
+    }
+  }
+
+  private Counter getSuccessCounter() {
+    return registry.counter("pubsub.amazon.published", "name", getName(), "topic", getTopicName());
```
marchello2000: why not init it once and store as instance var?

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/SNSPublisher.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  private Counter getErrorCounter(Exception e) {
+    return registry.counter(
```
marchello2000: personally, i prefer something like this:
https://github.com/spinnaker/orca/pull/3430/files#diff-9c5eb62982a1c7e4a37c3f0b83973543R96
and 
https://github.com/spinnaker/orca/pull/3430/files#diff-541df92cd9036d3155fe9bf54405fac3R23
but... don't feel strongly but slight less work on every invocation

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/config/AmazonPubsubProperties.java
marchello2000: i believe spring will call the no args .ctor and then set props, but not 100% sure

kork-core/src/main/java/com/netflix/spinnaker/kork/eureka/EurekaActivated.java
```diff
@@ -0,0 +1,59 @@
...
+
+import com.netflix.appinfo.InstanceInfo;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.context.ApplicationListener;
+
+/**
+ * A component that starts doing something when the instance is up in discovery and stops doing that
+ * thing when it goes down.
```
robzienert: We should probably stay consistent in OSS on calling Eureka, Eureka rather than also calling it Discovery.

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/SQSSubscriberProvider.java
```diff
@@ -0,0 +1,118 @@
...
+    this.properties = properties;
+    this.pubsubSubscribers = pubsubSubscribers;
+    this.pubsubMessageHandlerFactory = pubsubMessageHandlerFactory;
+    this.registry = registry;
+  }
+
+  @PostConstruct
+  public void start() {
+    if (properties == null) {
+      return;
```
robzienert: I feel like this should throw an exception, or at minimum log an error. If you've got the distance of setting up configuration to the point that `@ConditionalOnExpression("${pubsub.enabled:false} && ${pubsub.amazon.enabled:false}")` evaluates true, I would think the application is misconfigured and should not silently fail to do what it's supposedly enabled to do.

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/api/AmazonPubsubMessageHandler.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.kork.pubsub.aws.api;
+
+import com.amazonaws.services.sqs.model.Message;
+
+public interface AmazonPubsubMessageHandler {
+  void handleMessage(Message message);
```
robzienert: Class and method docs on interfaces.

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/api/AmazonPubsubMessageHandlerFactory.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.kork.pubsub.aws.api;
+
+import com.netflix.spinnaker.kork.pubsub.aws.config.AmazonPubsubProperties;
+
+public interface AmazonPubsubMessageHandlerFactory {
+  AmazonPubsubMessageHandler create(AmazonPubsubProperties.AmazonPubsubSubscription subscription);
```
robzienert: Docs on interfaces.

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/config/AmazonPubsubConfig.java
```diff
@@ -0,0 +1,61 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+
+@Configuration
+@ConditionalOnExpression("${pubsub.enabled:false} && ${pubsub.amazon.enabled:false}")
+@EnableConfigurationProperties(AmazonPubsubProperties.class)
+@Import(BastionConfig.class)
+public class AmazonPubsubConfig {
+  @Valid @Autowired private AmazonPubsubProperties amazonPubsubProperties;
```
robzienert: Now that I see this, `SQSSubscriberProvider` has a redundant `@ConditionalOnExpression` annotation, and the conditional handling that its `properties` member is null will always evaluate to false.

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/config/AmazonPubsubProperties.java
robzienert: Do we serialize these configuration properties anywhere? Spring does not need constructors for its properties.

kork-pubsub/src/main/java/com/netflix/spinnaker/kork/pubsub/model/PubsubPublisher.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.kork.pubsub.model;
+
+import java.util.Collections;
+import java.util.Map;
+
+public interface PubsubPublisher {
```
robzienert: Docs.

kork-pubsub/src/main/java/com/netflix/spinnaker/kork/pubsub/model/PubsubSystem.java
```diff
@@ -0,0 +1,34 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.kork.pubsub.model;
+
+public enum PubsubSystem {
```
robzienert: nit: I'm not particularly stoked on using enums for things like this. It means that if anyone comes along and wants to add an extension for their own pubsub implementation, they need to make a change to core Spinnaker code before it's even possible to do so.

kork-core/src/main/java/com/netflix/spinnaker/kork/eureka/EurekaActivated.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+public interface EurekaActivated extends ApplicationListener<RemoteStatusChangedEvent> {
+
+  Logger log = LoggerFactory.getLogger(EurekaActivated.class);
+
+  AtomicBoolean enabled = new AtomicBoolean();
```
cfieber: this field is static since declared on the interface - I feel like if that is the intention then there should be a single concrete EurekaActivated object in the application context that other components could have injected rather than having things extend this interface

kork-pubsub-aws/kork-pubsub-aws.gradle
cfieber: I like to do `tasks.compileGroovy.enabled = false` on modules where I don't want groovy in the main source tree

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/MessageAttribute.java
```diff
@@ -0,0 +1,37 @@
...
+@Data
+public class MessageAttribute {
+
+  @JsonProperty("Type")
+  private String attributeType;
+
+  @JsonProperty("Value")
+  private String attributeValue;
+
+  public MessageAttribute() {}
```
cfieber: I think you can control constructors via some lombokery (`@NoArgsConstructor` and `@AllArgsConstructor` some of which you maybe get for free from `@Data`)

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/NotificationMessage.java
```diff
@@ -0,0 +1,61 @@
...
+
+  public NotificationMessage() {}
+
+  public NotificationMessage(
+      String type,
+      String messageId,
+      String topicArn,
+      String subject,
+      String message,
+      Map messageAttributes) {
```
cfieber: would type this if you can't get the constructors for free

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/PubSubUtils.java
```diff
@@ -14,26 +14,44 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /** Utils for working with AWS SNS and SQS across services */
 public class PubSubUtils {
   private static final Logger log = LoggerFactory.getLogger(PubSubUtils.class);
 
+  private static String getQueueUrl(AmazonSQS amazonSQS, ARN queueARN) {
+    String queueUrl;
+    try {
```
cfieber: maybe worth some retrys around this bit

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/PubSubUtils.java
```diff
@@ -63,4 +81,37 @@ public static Policy buildSQSPolicy(ARN queue, ARN topic) {
...
     return new Policy("allow-sns-send", Collections.singletonList(snsStatement));
   }
+
+  /**
+   * Ensure that the topic exists and has a policy granting the specified accounts permission to
+   * publish messages to it
+   */
+  public static String ensureTopicExists(
+      AmazonSNS amazonSNS, ARN topicARN, AmazonPubsubSubscription subscription) {
+    String createdTopicARN = amazonSNS.createTopic(topicARN.getName()).getTopicArn();
```
cfieber: maybe worth some retries on this too - I suspect topic creation is idempotentish so probably safe to retry.

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/SNSPublisher.java
```diff
@@ -0,0 +1,120 @@
...
+
+  public Optional<PublishResult> publishMessage(String message) {
+    if (!isEnabled.get()) {
+      log.warn("Publishing is disabled for topic {}, dropping message {}", topicARN, message);
+      return Optional.empty();
+    }
+
+    try {
+      PublishRequest publishRequest = new PublishRequest(topicARN.getArn(), message);
+      PublishResult publishResponse = amazonSNS.publish(publishRequest);
```
cfieber: any thing we could do to safely retry in here?

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/SNSPublisherProvider.java
```diff
@@ -0,0 +1,91 @@
...
+import javax.annotation.PostConstruct;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.stereotype.Component;
+
+/** Creates one SNSPublisher per subscription */
+@Component
+@ConditionalOnExpression("${pubsub.enabled:false} && ${pubsub.amazon.enabled:false}")
```
cfieber: can replace this with `@ConditionalOnProperty({"pubsub.enabled", "pubsub.amazon.enabled"})`

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/SQSSubscriber.java
```diff
@@ -0,0 +1,170 @@
...
+  private void initializeQueue() {
+    this.queueUrl =
+        PubSubUtils.ensureQueueExists(
+            amazonSQS, queueARN, topicARN, subscription.getSqsMessageRetentionPeriodSeconds());
+    PubSubUtils.subscribeToTopic(amazonSNS, topicARN, queueARN);
+  }
+
+  private void listenForMessages() {
+    while (isEnabled.get()) {
+      ReceiveMessageResult receiveMessageResult =
```
cfieber: might need error handling here not to break out of the while loop on an AmazonException

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/SQSSubscriber.java
```diff
@@ -0,0 +1,170 @@
...
+        continue;
+      }
+
+      receiveMessageResult.getMessages().forEach(this::handleMessage);
+    }
+  }
+
+  private void handleMessage(Message message) {
+    try {
+      messageHandler.handleMessage(message);
```
cfieber: is it the messageHandler's job to mark the message as handled successfully?
would it make sense to put that responsibility in here, and then maybe on error we could know that we could return the message to the queue with a failure status and eventually it can get DLQ'd (otherwise it won't redrive until its visiblity timeout expires - maybe a longer backoff than we want)

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/SQSSubscriber.java
cfieber: other thing here is that SQS can deliver the same message twice (in a short-ish span of time) - do we want to do anything here to try to prevent processing the same message twice in a cluster (shared datasource in which to try and dibs the specific message for example)

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/SQSSubscriberProvider.java
```diff
@@ -0,0 +1,118 @@
...
+import javax.annotation.PostConstruct;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.stereotype.Component;
+
+/** * Starts the individual SQS workers (one for each subscription) */
+@Component
+@ConditionalOnExpression("${pubsub.enabled:false} && ${pubsub.amazon.enabled:false}")
```
cfieber: same on the collapsing this to a `@ConditionalOnProperty`

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/config/AmazonPubsubProperties.java
dreynaud: I think @marchello2000 is right, I tried to remove the no args constructor and then Spring got angry at me

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/SNSPublisher.java
dreynaud: why not indeed

kork-pubsub-aws/kork-pubsub-aws.gradle
dreynaud: yes, otherwise gradle says `':kork-pubsub-aws:test'` has no source files

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/SNSPublisher.java
dreynaud: noted, thanks

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/SQSSubscriber.java
emjburns: There's another bit of the echo pubsub bits that do this [here](https://github.com/spinnaker/echo/blob/master/echo-pubsub-core/src/main/java/com/netflix/spinnaker/echo/pubsub/PubsubMessageHandler.java)

kork-core/src/main/java/com/netflix/spinnaker/kork/eureka/EurekaActivated.java
dreynaud: Trying to remember what this is... It's a copy from `DiscoveryActivated` in echo, I renamed it `EurekaActivated` for OSS consistency but forgot the javadoc. Good catch!

kork-pubsub/src/main/java/com/netflix/spinnaker/kork/pubsub/model/PubsubSystem.java
dreynaud: ok, so straight up String then?

dreynaud: (for context, this is straight from echo)

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/PubSubUtils.java
dreynaud: done! thanks

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/PubSubUtils.java
dreynaud: badaboom, badadone

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/SNSPublisherProvider.java
dreynaud: TIL I learned! 🙏

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/SQSSubscriber.java
dreynaud: I was thinking it's kind of up to the app to handles duplicates 🤷‍♂ Or at least document it for now, and see later if we want to have duplicate support in kork

dreynaud: re: exception handling, there is a `catch (Exception e)` around the call to `listenForMessages()` which should work

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/SQSSubscriber.java
dreynaud: good point, I haven't dealt with message acknowledgement yet

dreynaud:
> Are you planning on pulling all the pubsub things to kork (inluding the google things)?

@emjburns yes I think it would make sense, then echo could "just" adopt the kork-pubsub lib and all the services would benefit from the shared improvements

dreynaud:
Thanks everyone for the great feedback! I addressed most of the low-hanging fruits in 3dbc731, still a couple more things on my todo list:

- add some DynamicConfig support to disable message processing at runtime
- port some of the tests in echo that exist for these classes
- document what's expected of applications in terms of message duplicates and acknowledgement
- make EurekaActivated an injectable bean
- a high-level README.md doc

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/config/AmazonPubsubProperties.java
ttomsu: @marchello2000 's explanation matches my understanding of what Spring does under the hood.

ttomsu:
@dreynaud Will you be the person to port it over? We don't have the bandwidth at the moment for this.

Also - can I get a little more context here? Why is it necessary to pull PubSub code into Kork?

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/SQSSubscriber.java
emjburns: @dreynaud I think duplicate notifications should be part of the kork support. It feels weird that the duplicate support is already part of the echo support, but you wouldn't port it over. It seems like a really helpful feature that would be tricky to implement again and again.

dreynaud:
> @dreynaud Will you be the person to port it over? We don't have the bandwidth at the moment for this.
> 
> Also - can I get a little more context here? Why is it necessary to pull PubSub code into Kork?

Hi @ttomsu yes, I'll deal with the migration. 

The idea here is that I intend to implement a feature based on pubsub in orca, and after a little bit of research I found some support in clouddriver and really good support in echo. It seems worthwhile making this usable as a library so that all services that rely on some form of pubsub have some consistent level of support/config/metrics, well defined interfaces, etc.

kork-pubsub/src/main/java/com/netflix/spinnaker/kork/pubsub/model/PubsubSystem.java
robzienert: Ya, I think String would do just fine.

kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/SQSSubscriber.java
robzienert: I'd agree with @emjburns that duplicates should be handled by kork. Perhaps this could be done via an interface & default implementation that services can override if they need to, but I don't suspect it would be overridden often.

PR #517, opened 2020-02-17
Revert "chore(dependencies): Upgrade Spring Cloud to Hoxton.SR1 (#455)"

marchello2000:
This reverts commit 51e0197b8e7c0db841619b5b1c533bce2ce207ca.

With this change, `echo` and `clouddriver` (the two services that pick up `kork-config` module) fail to start with the following error:

```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'regionProvider' 
  defined in class path resource [com/netflix/spinnaker/kork/configserver/autoconfig/SpringCloudAwsConfiguration.class]: Bean instantiation via factory method failed; 
  nested exception is org.springframework.beans.BeanInstantiationException: 
  Failed to instantiate [org.springframework.cloud.aws.core.region.StaticRegionProvider]: Factory method 'environmentPropertiesAwsS3RegionProvider' threw exception;
    nested exception is java.lang.IllegalArgumentException: The region 'null' is not a valid region!
```

I tried to fix it by adding `ConditionalOnProperty` on `SpringCloudAwsConfiguration` but then it just failed with a bunch of other errors.
I don't think it's OK for `kork` to create beans that are a) on by default and b) not turn-offable.


scottfrederick:
I'm not sure why this is happening, given that there's [code specifically intended to work around this ](https://github.com/spinnaker/kork/commit/51e0197b8e7c0db841619b5b1c533bce2ce207ca#diff-1d435bbcd151ce51762e0ebb4aeebe2fR45-R59) awkwardness in the Spring Cloud AWS configuration. 

I'll see if I can find some time to test again. If anyone else is already diving in, I'm happy to discuss.

PR #518, opened 2020-02-18
feat(plugins): Add `@ExtensionConfiguration` annotation, removing `@SpinnakerExtension` and `ConfigurableExtension` interface

jonsie:
*Breaking change:*

Building off work @robzienert did to add constructor injection, this simplifies the story around configuring extensions.  We now annotate configuration classes with `@ExtensionConfiguration`, provide an annotation value that lines up with our config, and things should Just Work.  

The constructor injection will fail if the constructor contains anything other than an `@ExtensionConfiguration` class or a `PluginsSdkImpl`.  I also added the (currently unused) concept of a plugin configuration, if we at some point want a high level config for a plugin.


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerExtensionFactory.kt
```diff
@@ -68,10 +63,22 @@ class SpinnakerExtensionFactory(
...
+            configFactory.createExtensionConfig(
+              paramType,
+              paramType.getAnnotation(ExtensionConfiguration::class.java).value,
+              pluginWrapper?.descriptor?.pluginId
+            )
+          }
+          else -> {
+            throw IntegrationException("'${extensionClass.simpleName}' extension has unsupported " +
+              "constructor argument class '${paramType.simpleName}'.  Expected argument classes " +
+              "should be annotated with @ExpectedConfiguration or implement PluginSdks.")
```
jonsie: While writing this message I thought - should we have configuration classes implement a config interface?  With that mechanism we would be able to load config classes without requiring the `@ExtensionConfiguration` annotation.  The benefit being one would not have to think about the YAML config structure when writing a config class if the defaults member values suffice. Ya dig?

kork-plugins-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/ExtensionConfiguration.java
```diff
@@ -22,20 +22,23 @@
...
-@Extension
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.TYPE)
 @Documented
-public @interface SpinnakerExtension {
-  /** The unique canonical id of the extension. */
-  String id();
+public @interface ExtensionConfiguration {
+  @Nonnull
+  String value();
```
robzienert: Let's add some docs on what the the value actually is supposed to be. There's currently no indicator that this value is supposed to be the configuration key (path?)

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerExtensionFactory.kt
robzienert: Personally, I prefer annotations over implementing interfaces / extending classes for use cases like this.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/config/ConfigCoordinates.kt
```diff
@@ -24,15 +24,31 @@ sealed class ConfigCoordinates {
 /**
```
robzienert: Side thought: We should probably do more to normalize these config coordinates (`toLowerCase()` and maybe even converting from `camelCase` to `snake-case` automatically for people. This would obviously be a different PR, just putting a thought out there.

jonsie: Good idea.

PR #520, opened 2020-02-18
chore(dependencies): remove CVE-2019-10173, bump xstream dependency

asgard-r:
- remove CVE-2019-10173, bump xstream dependency


asgard-r:
upgrade eureka version with new version of xstream on:
https://github.com/spinnaker/kork/pull/524

PR #521, opened 2020-02-19
* Add additional use cases to SecretsManagerSecretEngine

mayitbeegh:
* Add unit tests


mayitbeegh:
Context:
SecretsManager allows 2 types of encrypted data: string and binary. String usually takes the form of key-value pairs.

Before this PR, the engine is able to:
- return decrypted value when the encrypted data is a key-value string and a key is provided
- return decrypted value when the encrypted data is a binary and a key is NOT provided

With this PR, the engine is able to:
- return decrypted value regardless of the data type of the encrypted data if config starts with `encryptedFile`
- return decrypted value when the encrypted data is a key-value string and a key is provided
- return decrypted value when the encrypted data is a plaintext string and a key is NOT provided

kork-secrets-aws/kork-secrets-aws.gradle
```diff
@@ -8,4 +8,7 @@ dependencies {
 
   implementation "com.amazonaws:aws-java-sdk-s3"
   implementation "com.amazonaws:aws-java-sdk-secretsmanager"
+
+  testImplementation "org.spockframework:spock-core"
```
fieldju: nit: your not using this as far as I can tell.

kork-secrets/src/main/java/com/netflix/spinnaker/kork/secrets/EncryptedSecret.java
```diff
@@ -67,6 +69,7 @@ public static EncryptedSecret parse(String secretConfig) {
   }
 
   protected void update(String secretConfig) {
+    encryptedFile = isEncryptedFile(secretConfig);
```
fieldju: What is this for?

mayitbeegh: a.isFile equivalent, for determine if a config begins with `encryptedFile`

kork-secrets-aws/kork-secrets-aws.gradle
mayitbeegh: It's only for pulling in junit. I find other modules using spock so it looks more aligned.

PR #524, opened 2020-02-20
chore(dependencies): resolve CVEs issues

asgard-r:
Upgrade eureka version.
Fix CVE-2019-10173 caused by xstream dependency from following services:
- clouddriver
- echo
- fiat
- front50
- gate
- halyard
- igor
- kayenta
- orca
- rosco


dwilliams-armory:
@ezimanyi Do you mind having a look at this one?

robzienert:
This should be fine.

PR #525, opened 2020-02-20
feat(plugins): Adding Front50 FileDownloader

robzienert:
Allows services to download plugins from Front50's proxy.


kork-plugins/src/main/java/com/netflix/spinnaker/config/Front50PluginsConfiguration.java
```diff
@@ -58,28 +57,33 @@ public static UpdateRepository pluginFront50UpdateRepository(
...
+      PluginOkHttpClientProvider pluginsOkHttpClientProvider) {
+    return new Front50FileDownloader(pluginsOkHttpClientProvider.getOkHttpClient());
+  }
+
+  @Bean
+  public static UpdateRepository pluginFront50UpdateRepository(
+      Environment environment,
+      PluginOkHttpClientProvider pluginsOkHttpClientProvider,
+      Map<String, PluginRepositoryProperties> pluginRepositoriesConfig,
+      Front50FileDownloader front50FileDownloader) {
```
robzienert: This isn't really a good, long-term solution. I imagine some people will want to use Front50 for plugin info without necessarily downloading binaries from Front50 as well. This seemed like a bigger topic to bite off than what this PR is for. Do we add a `CompositeFileDownloader` class that lets `FileDownloader`s to inspect a URL and see if it's something they should be downloading? Seems like this might be the only path forward...?

kork-plugins/src/main/java/com/netflix/spinnaker/config/Front50PluginsConfiguration.java
```diff
@@ -58,28 +57,33 @@ public static UpdateRepository pluginFront50UpdateRepository(
...
-
     OkHttpClient okHttpClient =
         new OkHttp3ClientConfiguration(okHttpClientProperties)
             .create()
             .retryOnConnectionFailure(okHttpClientProperties.isRetryOnConnectionFailure())
             .build();
 
+    return new PluginOkHttpClientProvider(okHttpClient);
+  }
+
```
srekapalli: With this we will loose the ability to download from github ? Correct.?

kork-plugins/src/main/java/com/netflix/spinnaker/config/Front50PluginsConfiguration.java
srekapalli: What if we assign a  type of Downloader to a Plugin. ?  So when we register a plugin.. assign the Downloader type to it. That way we can dynamically swtich and allow us to be flexible.

srekapalli: You are correct. We don't need to address this issue immdtly.

srekapalli: Yh. CompositeFileDownloader with each FileDownloader saying the type/name of it which we can hold in PluginInfo instead of URL inspection strategy will be more clear I think.

robzienert:
@srekapalli Take another look - I made a change that adds a `CompositeFileDownloader` and `SupportingFileDownloader` so we can support different `FileDownloader`s for any particular repository.

robzienert:
I opted _not_ to assign a downloader to a plugin because it would be a large change (we'd have to stop using `pf4j-update` almost entirely). URL inspection gets the immediate requirement out of the way, but if we absolutely need more power to inspect `PluginInfo`, we can go down that path later.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/downloader/Front50FileDownloader.kt
```diff
@@ -0,0 +1,60 @@
...
+import java.nio.file.Paths
+
+/**
+ * Downloads plugin binaries from Front50.
+ */
+class Front50FileDownloader(
+  private val okHttpClient: OkHttpClient,
+  private val front50BaseUrl: URL
+) : SupportingFileDownloader {
+
```
srekapalli: neat!!

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/internal/Front50Service.kt
```diff
@@ -12,10 +12,9 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
- *
  */
 
-package com.netflix.spinnaker.kork.plugins.update.front50
+package com.netflix.spinnaker.kork.plugins.update.internal
```
jonsie: 👍

PR #526, opened 2020-02-21
feat(plugins): Use configured plugin version (falling back to latest if not specified) and check service version requirement constraints

jonsie:
Todo:
- [x] Write tests


kork-core/src/main/java/com/netflix/spinnaker/kork/version/ManifestVersionResolver.java
```diff
@@ -35,18 +35,28 @@
...
+ * attribute is useful if JARs are built in a process outside the context of OSS, but you need a
+ * mechanism to get the underlining OSS version.
  */
 @Slf4j
 public class ManifestVersionResolver implements VersionResolver {
 
   private static final String GROUP = "com.netflix.spinnaker";
 
   private final String group;
+  private boolean ossAttribute = false;
```
robzienert: I don't understand the purpose of this flag. Can't the logic just be something like this:

```kotlin
val version = attributes.getValue("Implementation-OSS-Version") ?: attributes.getValue("Implementation-Version")
```

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerServiceVersionManager.kt
```diff
@@ -0,0 +1,48 @@
...
+ */
+class SpinnakerServiceVersionManager(
+  private val serviceName: String
+) : VersionManager {
+
+  override fun checkVersionConstraint(version: String, requires: String): Boolean {
+    val serviceVersionRequirements =
+      VersionRequirementsParser
+        .parseAll(requires)
+        .find { it.service == serviceName }
```
robzienert: Should make this match case insensitive.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/PluginDownloadService.kt
```diff
@@ -0,0 +1,135 @@
...
+
+    availablePlugins
+      .filter { !pluginStatusProvider.isPluginDisabled(it.id) }
+      .forEach { enabledPlugin ->
+        val configuredPluginVersion = pluginStatusProvider.configuredPluginVersion(enabledPlugin.id)
+
+        val pluginRelease = if (configuredPluginVersion == null) {
+          val fallbackRelease = updateManager.getLastPluginRelease(enabledPlugin.id)
+            ?: throw PluginNotFoundException(enabledPlugin.id, configuredPluginVersion)
+          run {
```
robzienert: Hmm, what object is this `run` on?

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/PluginDownloadService.kt
```diff
@@ -0,0 +1,135 @@
...
+      throw UnsupportedOperationException("This operation is only supported on the specified plugins root directory.")
+    }
+  }
+
+  internal class PluginNotFoundException(pluginId: String, configuredPluginVersion: String?) :
+    IntegrationException(
+      "'$pluginId' is enabled with version '${configuredPluginVersion ?: "undefined" }', but a " +
+        "release version could not be found that satisfies the version and/or the service " +
+        "requirement constraints."
+  )
```
robzienert: Great exception message.

kork-core/src/main/java/com/netflix/spinnaker/kork/version/ManifestVersionResolver.java
jonsie: Actually, my original intention here was to not fallback to `Implementation-Version` if `Implementation-OSS-Version` is desired (which I didn't do, got distracted but I will revisit this now).  I don't think we want to fallback if the OSS version is desired since that'll basically throw off all our service version constraint stuff.

robzienert:
Progress looks legit.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/PluginDownloadService.kt
jonsie: it runs on `PluginDownloadService` but it's unnecessary - I removed this.

kork-core/src/main/java/com/netflix/spinnaker/kork/version/ManifestVersionResolver.java
jonsie: This will now not fall through to `Implementation-Version` if the property `use-oss-version-manifest-attribute` is set in `spinnaker.extensibility`.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/ExtensionBeanDefinitionRegistryPostProcessor.kt
```diff
@@ -47,8 +49,10 @@ class ExtensionBeanDefinitionRegistryPostProcessor(
   override fun postProcessBeanDefinitionRegistry(registry: BeanDefinitionRegistry) {
     val start = System.currentTimeMillis()
     log.debug("Preparing plugins")
-    updateManagerService.checkForUpdates()
     pluginManager.loadPlugins()
+    updateManager.downloadPlugins(
```
jonsie: `pluginManager.loadPlugins()` needs to run before we download plugins so we know what is loaded and if there is a new version that should be downloaded.

kork-core/src/main/java/com/netflix/spinnaker/kork/version/ManifestVersionResolver.java
robzienert: nit: How about `spinnaker.extensibility.versioning.use-oss-version-manifest-attribute` instead? I suspect we'll end up having other config informing Spinnaker how to resolve versions in the future, it'd be nice to have them all under one place.

kork-plugins/src/main/java/com/netflix/spinnaker/config/PluginsConfigurationProperties.java
```diff
@@ -98,4 +98,12 @@ public boolean isEnabled() {
       return enabled;
     }
   }
+
+  /**
+   * If set to true, the service version resolution will use the MANIFEST.MF attribute
+   * "Implementation-OSS-Version". By default, this is false which means the version attribute
+   * lookup will be "Implementation-Version" -- for most OSS users that will still be the OSS
+   * version.
```
robzienert: Worded a little confusingly. Perhaps:

> Defines what MANIFEST.MF attribute to use in determining a service version. By default, a service version is determined by reading the "Implementation-Version" attribute. If this property is set to true, however, "Implementation-OSS-Version" will be used to support use cases where services are being extended and rebuilt. Unless you're re-building services, this should remain its default value.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/ExtensionBeanDefinitionRegistryPostProcessor.kt
robzienert: Code comment instead of PR comment. That's valuable context that shouldn't require hunting for this PR comment.

kork-plugins/src/main/java/com/netflix/spinnaker/config/PluginsConfigurationProperties.java
jonsie: Appreciated.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/ExtensionBeanDefinitionRegistryPostProcessor.kt
```diff
@@ -47,8 +49,19 @@ class ExtensionBeanDefinitionRegistryPostProcessor(
...
+    // 2) Determine the plugins for release from the list of available plugins
+    val releases = pluginInfoReleaseProvider.getReleases(updateManager.availablePlugins)
+
+    // 3) Download releases, updating previously loaded plugins where necessary
+    updateManager.downloadPluginReleases(releases).forEach { pluginPath ->
+      pluginManager.loadPlugin(pluginPath)
+    }
+
+    // 4) Start plugins - should only be called once in kork-plugins
     pluginManager.startPlugins()
```
jonsie: Feels like steps 1-4 could eventually be broken out into a separate service - would make it much easier to test.

jonsie:
This should be ready to go.  We now load plugins from the filesystem first, then gather up our list of plugin releases before passing them to the `SpinnakerUpdateManager` which then decides which plugins to download.  Loading and starting plugins is still the job of `ExtensionBeanDefinitionRegistryPostProcessor` though we may want to break that out at some point so it's easy to test.

PR #528, opened 2020-02-25
chore(plugins): removing mvp plugins implementation

link108:
Removing the mvp plugins implementation, as it has been deprecated for some time now. https://github.com/spinnaker/orca/pull/3457 should be merged first.


plumpy:
Yay! Thanks!

PR #529, opened 2020-02-25
feat(metrics): Add extra tags capability for OkHttp metrics interceptor

robzienert:
There was [feedback](https://github.com/spinnaker/kork/pull/507#discussion_r376753588) on the HttpClient SDK that it would be valuable to attach additional metrics tags to okhttp requests. Adding a `ThreadLocal` appears to be the only realistic path forward here since the interceptor is buried far away.

### DIFF
--- a/kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
+++ b/kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
@@ -1,15 +1,15 @@
 package com.netflix.spinnaker.okhttp;
 
+import com.netflix.spectator.api.Id;
 import com.netflix.spectator.api.Registry;
+import com.netflix.spectator.api.Tag;
 import com.netflix.spinnaker.kork.common.Header;
 import com.squareup.okhttp.Interceptor;
 import com.squareup.okhttp.Request;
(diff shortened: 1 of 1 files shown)


kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
```diff
@@ -134,4 +143,25 @@ private static String bucket(int statusCode) {
...
+    }
+    return id.withTags(tags);
+  }
+
+  /**
+   * Add additional metric tags to OkHttp client requests.
+   *
+   * <p>Additional tags cannot override default tags and will be silently dropped.
+   */
+  public static void addTags(Collection<Tag> tags) {
```
jonsie: Should we dedupe extra tags too?

robzienert:
@ajordens Interaction is weird because the Interceptor is completely unreachable from user-land code. If you have suggestions on how this could be improved, I'm all ears.

kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
robzienert: I don't think there will be dupes, will there?

PR #530, opened 2020-02-27
fix(plugins): Allow front50 to be inaccessible while loading plugins

robzienert:
We don't want services to be totally unable to start if Front50 is unavailable.


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/repository/Front50UpdateRepository.kt
```diff
@@ -57,10 +58,17 @@ class Front50UpdateRepository(
...
+      val response = retry.executeSupplier { front50Service.listAll().execute() }
 
       if (!response.isSuccessful) {
-        throw SystemException("Unable to list front50 plugin info", response.message())
+        // We can't throw an exception here when we fail to talk to Front50 because it will prevent a service from
+        // starting. We would rather a Spinnaker service start and be potentially misconfigured than have a hard
+        // startup dependency on front50.
+        log.error(
+          "Failed listing plugin info from front50: This service may not download plugins that it needs: {}",
+          response.message()
```
dreynaud: is there a way to log the exception here?

robzienert: Updated. It's not the greatest, but it'll get the job done.

PR #531, opened 2020-02-28
fix(plugins): Rework ServiceVersion bean wiring

robzienert:
Some problems with `ServiceVersion` wiring after adding it to the plugin framework.

- There are potentially multiple `VersionResolver`s; `ServiceVersion` should be used by the codebases.
- The plugin framework initializes before normal Spring wiring, so beans wired up by other areas of kork or an application will not be available yet.

Also moved the config to a more general location (`service-version`, instead of `spinnaker.extensibility`, since `ServiceVersion` is not part of the plugin framework itself.


robzienert:
This will fix https://github.com/spinnaker/orca/pull/3471

PR #535, opened 2020-03-02
fix(web): GenericExceptionHandlers should handle standard web mvc exceptions with correct http status

Aloren:
As requested by: https://github.com/spinnaker/kayenta/pull/662#issuecomment-585404962

I've added mvc tests to verify the http status is correct. I haven't found any response body formats to verify though.


Aloren:
@robzienert fyi

robzienert:
Sweet, thanks!

PR #537, opened 2020-03-02
fix(plugins): Misc. fixes in path to downloading and loading configured plugin version

jonsie:


kork-plugins/src/main/java/com/netflix/spinnaker/config/PluginsConfigurationProperties.java
```diff
@@ -98,13 +98,4 @@ public boolean isEnabled() {
...
   }
-
-  /**
-   * Defines what MANIFEST.MF attribute to use in determining a service version. By default, a
-   * service version is determined by reading the "Implementation-Version" attribute. If this
-   * property is set to true, however, "Implementation-OSS-Version" will be used to support use
-   * cases where services are being extended and rebuilt. Unless you're re-building services, this
-   * should remain its default value.
-   */
-  public boolean useOssVersionManifestAttribute = false;
```
jonsie: This was moved to a different properties file and no longer necessary here.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/ExtensionBeanDefinitionRegistryPostProcessor.kt
```diff
@@ -53,8 +53,8 @@ class ExtensionBeanDefinitionRegistryPostProcessor(
     // 1) Load plugins prior to downloading so we can resolve what needs to be updated
     pluginManager.loadPlugins()
 
-    // 2) Determine the plugins for release from the list of available plugins
-    val releases = pluginInfoReleaseProvider.getReleases(updateManager.availablePlugins)
+    // 2) Determine the plugins for release from the list of plugins
+    val releases = pluginInfoReleaseProvider.getReleases(updateManager.plugins)
```
jonsie: Apparently, `availablePlugins` only returns the available plugins that are not yet loaded by the plugin manager.  This will allow us to determine the release against all plugins.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/SpinnakerUpdateManager.kt
```diff
@@ -58,7 +58,9 @@ class SpinnakerUpdateManager(
         if (loadedPlugin != null) {
           val loadedPluginVersion = loadedPlugin.descriptor.version
 
-          if (pluginManager.versionManager.compareVersions(release.props.version, loadedPluginVersion) > 0) {
+          // If a plugin was built without a version specified (via the Plugin-Version MANIFEST.MF
+          // attribute), to be safe we always check for the configured plugin version.
+          if (loadedPluginVersion == "unspecified" || pluginManager.versionManager.compareVersions(release.props.version, loadedPluginVersion) > 0) {
```
jonsie: We should be making a point of building all our plugins with the `Plugin-Version` attribute (which we aren't doing in the helloworld plugin I was testing with), but just to be safe this will force us to use the correctly configured version.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/downloader/Front50FileDownloader.kt
```diff
@@ -48,13 +48,14 @@ class Front50FileDownloader(
...
-    return downloadDir.resolve(Paths.get(fileUrl.path).fileName).also {
+    return downloadDir.resolve(Paths.get(fileUrl.path + binaryExtension).fileName).also {
       Files.write(it, body.bytes())
     }
   }
 
   companion object {
     private val downloadDir = Files.createTempDirectory("plugin-downloads")
     private val retry = Retry.ofDefaults("plugin-front50-downloader")
+    private const val binaryExtension = ".zip"
```
jonsie: Since we don't ask Front50 for file + extension type (though it is constructed in Front50's `PluginBinaryController`) we need to do this to correctly save the file (otherwise it is saved as a binary file with no extension).

srekapalli:
LGTM.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/SpinnakerUpdateManager.kt
robzienert: I thought this was being populated by the gradle project?

PR #539, opened 2020-03-03
feat(plugins): Add plugin ID and extension name to MDC on extension method invocation

jonsie:


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/config/SpringEnvironmentConfigResolver.kt
```diff
@@ -95,6 +95,8 @@ class SpringEnvironmentConfigResolver(
       return missingCallback()
     }
 
+    log.debug("Found config at '$pointer'")
```
jonsie: Unrelated to this change, but this log helps when verifying configuration loading.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/proxy/aspects/InvocationAspect.kt
```diff
@@ -60,4 +60,13 @@ interface InvocationAspect<I : InvocationState> {
...
   fun error(e: InvocationTargetException, invocationState: I)
+
+  /**
+   * Called last and always called, regardless of invocation success or failure.
+   *
+   * Optional, default implementation is a no-op.
+   *
+   * @param invocationState The invocationState object created via [before]
+   */
+  fun finally(invocationState: I) { /* default implementation */ }
```
jonsie: I originally had a `finally` here and I thought it would be useful but it wasn't being used so I removed it.  Alas, it's back.

PR #545, opened 2020-03-06
fix(plugins): Do not enable configuration properties class OkHttpClientConfigurationProperties in static configuration context

jonsie:


kork-plugins/src/main/java/com/netflix/spinnaker/config/HttpClientSdkConfiguration.java
```diff
@@ -25,30 +26,39 @@
...
+import org.springframework.beans.factory.BeanCreationException;
+import org.springframework.boot.context.properties.bind.Bindable;
+import org.springframework.boot.context.properties.bind.Binder;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.context.annotation.Import;
 import org.springframework.core.env.Environment;
 
 @Configuration
-@EnableConfigurationProperties(OkHttpClientConfigurationProperties.class)
```
jonsie: Because we enabled configuration properties for `OkHttpClientConfigurationProperties` in this configuration static context (wherein beans are created early in the app lifecycle), I believe this properties object was being instantiated without all the parameters loaded from the configuration.  This resulted in missing configuration properties that are important for a functioning http client (like a keystore and a truststore, for example).

kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/proxy/aspects/LogInvocationAspectTest.kt
```diff
@@ -49,12 +50,15 @@ class LogInvocationAspectTest : JUnit5Minutests {
...
-      expectThat(mdcPluginExtension).isEqualTo(target.javaClass.simpleName.toString())
+    test("Plugin ID and extension name added to the MDC before invocation, removes in finally") {
+      val invocatonState = subject.before(target, proxy, method, args, spinnakerPluginDescriptor)
+      expectThat(MDC.get(Header.PLUGIN_ID.header)).isEqualTo(spinnakerPluginDescriptor.pluginId)
+      expectThat(MDC.get(Header.PLUGIN_EXTENSION.header)).isEqualTo(target.javaClass.simpleName.toString())
+
+      subject.finally(invocatonState)
+
+      expectThat(MDC.get(Header.PLUGIN_ID.header)).isNull()
+      expectThat(MDC.get(Header.PLUGIN_EXTENSION.header)).isNull()
```
jonsie: Unrelated test change I had locally, figure I may as well add it.

kork-plugins/src/main/java/com/netflix/spinnaker/config/PluginsAutoConfiguration.java
```diff
@@ -66,7 +66,7 @@
 
 @Configuration
 @EnableConfigurationProperties(PluginsConfigurationProperties.class)
-@Import({Front50PluginsConfiguration.class, HttpClientSdkConfiguration.class})
+@Import({Front50PluginsConfiguration.class})
 public class PluginsAutoConfiguration {
```
jonsie: I'd also like to not wire `HttpClientSdkConfiguration` as we work through some additional issues with plugin http clients.

PR #546, opened 2020-03-06
fix(plugins): Disable HttpClientSdkConfiguration

jonsie:
There are downstream failures due to this config being enabled by default, and honestly I just want to fix other bugs and move on and not deal with this problem right now.


plumpy:
What issues does this fix?

jonsie:
@plumpy Downstream Spring test failures (like Orca's `MainSpec` test).

plumpy:
Ah, I see, 1.19 was cut with kork 7.24.0, so we don't need to cherry-pick this then. Thanks.

jonsie:
@plumpy Hm, OkHttp clients on 7.24.0 are bugged.   I think you'll want to pull in 7.25.1 - mostly for this fix https://github.com/spinnaker/kork/pull/545.  That fix allows the configuration to load correctly which is important if you're configuring clients with a keystore and truststore (that was the specific bug we found).

PR #549, opened 2020-03-10
feat(plugins): Util/sdk to load YML resources from plugin source zips

srekapalli:
- This change allows us to load job configurations defined in yml resources from with in the plugin jar/zip.
- Can be used for other usecases where we declare any resource in Yml structures easily.

### DIFF
--- a/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/sdk/yaml/YamlResourceLoaderService.kt
+++ b/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/sdk/yaml/YamlResourceLoaderService.kt
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.kork.plugins.sdk.yaml
+
+import com.fasterxml.jackson.databind.DeserializationFeature
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory
+import com.netflix.spinnaker.kork.plugins.api.yaml.YamlResourceLoader
+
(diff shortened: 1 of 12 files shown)


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/sdk/yaml/YamlResourceLoaderService.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.fasterxml.jackson.databind.DeserializationFeature
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory
+import com.netflix.spinnaker.kork.plugins.api.yaml.YamlResourceLoader
+
+class YamlResourceLoaderService(private val extensionClass: Class<*>) : YamlResourceLoader {
+
+  override fun <T : Any?> loadResource(resourceName: String, toValueType: Class<T>): T? {
+
+    val mapper = ObjectMapper(YAMLFactory()).disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
```
jonsie: I think `mapper` should be a member field of `YamlResourceLoaderService` so we're not creating an object mapper on each `loadResource` call.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/sdk/yaml/YamlResourceLoaderService.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    this.extensionClass.classLoader.getResourceAsStream(resourceName).use { inputStream ->
+      if (inputStream != null) {
+        return mapper.readValue(inputStream, toValueType)
+      }
+      throw IllegalArgumentException("Cannot load specified resource: $resourceName")
```
jonsie: Include the extension class name in this error.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/sdk/yaml/YamlResourceLoaderService.kt
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package com.netflix.spinnaker.kork.plugins.sdk.yaml
+
+import com.fasterxml.jackson.databind.DeserializationFeature
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory
+import com.netflix.spinnaker.kork.plugins.api.yaml.YamlResourceLoader
+
+class YamlResourceLoaderService(private val extensionClass: Class<*>) : YamlResourceLoader {
```
jonsie: Nitpick - I would probably name this something more like `JacksonYamlResourceLoader` since it's really more about the implementation (using Jackson for deserialization) and not really a "service" if that makes sense.

PR #552, opened 2020-03-14
fix(plugins): Set the service version to the actual resolved value in…

srekapalli:
- This fixes the issue where gate was attempting to download all plugins instead of just deck related plugins.
- Service version was always set to 'unknown' before and this fix will actually set it what  itwas resolved to.


kork-core/src/main/java/com/netflix/spinnaker/kork/version/ServiceVersion.java
```diff
@@ -62,7 +62,10 @@ public String resolve() {
         }
       }
 
-      resolvedServiceVersion = UNKNOWN_VERSION;
+      if (resolvedServiceVersion == null || resolvedServiceVersion.isEmpty()) {
+        log.warn("Unable to determine the service version, setting it to unknown");
+        resolvedServiceVersion = UNKNOWN_VERSION;
+      }
```
robzienert: lol oops 💯

robzienert: note: I've setup an alert in Sumo to track for this log message once it's added.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/SpinnakerUpdateManager.kt
```diff
@@ -108,8 +108,15 @@ class SpinnakerUpdateManager(
...
 
+    // If service version cannot be determined(i.e 0.0.0),
+    // just check if the release requirement mentions the service to see if the plugin is applicable. This case should
+    // never happen, we should always know the system version but if it happens we can at least give a matching plugin
+    // to the service name in question.
+    //
+    // If version is determined, check the version constraint explicitly to decide if the plugin is applicable or not.
     for (release in pluginInfo.releases) {
-      if (systemVersion == "0.0.0" || versionManager.checkVersionConstraint(systemVersion, release.requires)) {
+      if ((systemVersion == "0.0.0" && release.requires.contains(serviceName, ignoreCase = true)) ||
```
robzienert: nit: Seeing `0.0.0` across two different files. Worth making a const somewhere instead?

PR #560, opened 2020-03-18
feat(okhttp): support for dispatcher max-conn tuning

asher:
I noticed that `okhttp3`'s async dispatcher implements a queue in order to enforce a maximum number of in-flight requests per hostname, with a default of 5. since we’ve seen its queues block and since keel attempts to support lots of concurrent resource checks which all require clouddriver calls (a single hostname), I think the ability to tune this would be nice.


kork-web/src/main/groovy/com/netflix/spinnaker/okhttp/OkHttpClientConfigurationProperties.groovy
```diff
@@ -29,6 +29,8 @@ import org.springframework.core.annotation.AliasFor
 class OkHttpClientConfigurationProperties {
   long connectTimeoutMs = 15000
   long readTimeoutMs = 20000
+  int maxRequests = 64
+  int maxRequestsPerHost = 5
```
asher: These are the default values within okHttp3.

PR #564, opened 2020-03-20
fix(plugins): allow for plugin configs on all services

link108:
Halyard does not currently have knowledge of which plugins should be placed onto each spinnaker service. For now, we can log a warning if there are plugin configs placed onto services that are not required for the plugin to function, so that halyard can successfully deploy plugins to Spinnaker.
Currently Spinnaker throws an error if plugin configs are laid down upon a service that does not require it

### DIFF
--- a/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginManager.kt
+++ b/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginManager.kt
@@ -138,9 +138,11 @@ open class SpinnakerPluginManager(
   override fun loadPlugin(pluginPath: Path?): String? {
     require(!(pluginPath == null || Files.notExists(pluginPath))) { "Specified plugin '$pluginPath' does not exist!" }
     log.debug("Loading plugin from '{}'", pluginPath)
-    val pluginWrapper = loadPluginFromPath(pluginPath) ?: return null
-    // try to resolve  the loaded plugin together with other possible plugins that depend on this plugin
-    resolvePlugins()
-    return pluginWrapper!!.descriptor.pluginId
+    return loadPluginFromPath(pluginPath)
+      ?.let {
+        // try to resolve  the loaded plugin together with other possible plugins that depend on this plugin
+        resolvePlugins()
+        it.descriptor.pluginId
+      }
   }
 }
--- a/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginManager.kt
+++ b/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginManager.kt
@@ -136,7 +136,7 @@ open class SpinnakerPluginManager(
 
   // TODO (link108): remove this override, once plugin deployments via halyard are fixed
   override fun loadPlugin(pluginPath: Path?): String? {
-    require(!(pluginPath == null || Files.notExists(pluginPath))) { String.format("Specified plugin %s does not exist!", pluginPath) }
+    require(!(pluginPath == null || Files.notExists(pluginPath))) { "Specified plugin '$pluginPath' does not exist!" }
     log.debug("Loading plugin from '{}'", pluginPath)
     val pluginWrapper = loadPluginFromPath(pluginPath) ?: return null
     // try to resolve  the loaded plugin together with other possible plugins that depend on this plugin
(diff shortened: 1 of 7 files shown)


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginManager.kt
```diff
@@ -127,4 +131,14 @@ open class SpinnakerPluginManager(
     .add(super.createPluginRepository())
 
   override fun getPluginFactory(): PluginFactory = spinnakerPluginFactory
+
+  // TODO (link108): remove this override, once plugin deployments via halyard are fixed
+  override fun loadPlugin(pluginPath: Path?): String? {
```
link108: add a config / flag to allow for strict mode

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/bundle/PluginBundleExtractor.kt
```diff
@@ -55,12 +58,20 @@ class PluginBundleExtractor {
...
-    // bundles should actually be downloaded, or B) The plugin author incorrectly identified this [service] as one
-    // that the plugin extends (via the PluginInfo `requires` list).
-    throw IntegrationException("Downloaded plugin bundle does not have plugin for service '$service'")
+    if (isStrictPluginLoading()) {
+      // If thrown, this is an indicator that either: A) There's a bug in the plugin framework resolving which plugin
+      // bundles should actually be downloaded, or B) The plugin author incorrectly identified this [service] as one
+      // that the plugin extends (via the PluginInfo `requires` list).
+      throw IntegrationException("Downloaded plugin bundle does not have plugin for service '$service'")
+    } else {
+      log.warn("Downloaded plugin bundle: {}, does not have plugin for service: {}", bundlePath.fileName, service)
```
robzienert:
```suggestion
      log.warn("Downloaded plugin bundle '{}' does not have plugin for service: {}", bundlePath.fileName, service)
```

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginManager.kt
```diff
@@ -127,4 +133,14 @@ open class SpinnakerPluginManager(
...
+    require(!(pluginPath == null || Files.notExists(pluginPath))) { String.format("Specified plugin %s does not exist!", pluginPath) }
```
robzienert:
```suggestion
    require(!(pluginPath == null || Files.notExists(pluginPath))) { "Specified plugin '$pluginPath' does not exist!" }
```

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginManager.kt
```diff
@@ -127,4 +133,14 @@ open class SpinnakerPluginManager(
...
+    log.debug("Loading plugin from '{}'", pluginPath)
+    val pluginWrapper = loadPluginFromPath(pluginPath) ?: return null
+    // try to resolve  the loaded plugin together with other possible plugins that depend on this plugin
+    resolvePlugins()
+    return pluginWrapper!!.descriptor.pluginId
```
robzienert: `!!` is a code smell, let's avoid it when it's easy to do so:

```kotlin
return loadPluginFromPath(pluginPath)
  ?.let {
    // try to resolve  the loaded plugin together with other possible plugins that depend on this plugin
    resolvePlugins()
    it.descriptor.pluginId
  }
```

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginManager.kt
```diff
@@ -54,16 +57,19 @@ open class SpinnakerPluginManager(
...
 
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
   private val springExtensionFactory: ExtensionFactory = SpinnakerExtensionFactory(
     this,
     configFactory,
     sdkFactories
   )
-  private val bundleExtractor = PluginBundleExtractor()
+  private val bundleExtractor = PluginBundleExtractor(environment)
```
jonsie: Nit: Mayne pass an instance of `PluginBundleExtractor` to the `SpinnakerPluginManager` and avoid adding a new member field `environment` which has the sole purpose of being used as a constructor arg for `PluginBundleExtractor`.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginManager.kt
link108: makes sense, updating 👍

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginManager.kt
link108: totally, I like that approach 👍

PR #566, opened 2020-03-20
fix(plugins): Use default okhttp properties if no config is bindable

robzienert:


robzienert:
yolo. This unblocks autobump.

PR #569, opened 2020-03-25
refactor(security): Use dedicated ThreadLocal for auth data store

robzienert:


kork-security/src/main/java/com/netflix/spinnaker/security/ThreadSecurityStore.java
```diff
@@ -0,0 +1,84 @@
...
+    MDC.setContextMap(map);
+  }
+
+  public void clear() {
+    STORE.remove();
+    MDC.clear();
+    try {
+      // force clear to avoid the potential for a memory leak if log4j is being used
+      Class<?> log4jMDC = Class.forName("org.apache.log4j.MDC");
+      log4jMDC.getDeclaredMethod("clear").invoke(null);
```
marchello2000: i wonder if this still a thing..?

marchello2000:
some tests failing, but otherwise lgtm

marchello2000:
Also, there are tons of places in `orca` that shove things direct into `MDC`... just FYI

kork-security/src/main/java/com/netflix/spinnaker/security/ThreadSecurityStore.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Thin adapter around a ThreadLocal used to store request-scoped security data.
+ *
+ * <p>This class also forwards writes to MDC so that these values can be available for logging as
+ * well.
+ */
+public class ThreadSecurityStore {
+
+  private final ThreadLocal<Map<String, String>> STORE = new ThreadLocal<>();
```
cfieber: I wonder if this should be InheritableThreadLocal.. what is it in the implementation of the MDC?

robzienert: MDC uses `InheritableThreadLocal`, yes. I'll switch it.

robzienert: After looking into this more, it looks like MDC uses vanilla ThreadLocal. Some implementations will use `InheritableThreadLocal`, but not in the places we use it.

ajordens: We specifically turned off the inheritance ... because it hindered cleanup when it crossed threads.

PR #573, opened 2020-03-26
fix(plugins): allow for relaxed plugin loading

link108:
These changes were meant to be part of https://github.com/spinnaker/kork/pull/564, I had some local commits I forgot to push 😬

### DIFF
--- a/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpringStrictPluginLoaderStatusProvider.kt
+++ b/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpringStrictPluginLoaderStatusProvider.kt
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2020 Armory, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


jonsie:
@link108 Any reason not to just add this functionality to the `SpringPluginStatusProvider`?

link108:
@jonsie I took a look at putting it there before, that class seems like its focused on specific plugin properties vs properties under the `spinnaker.extensibility` key, where the strict loading config goes. Can move it there, but want to be careful about overloading that class

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpringStrictPluginLoaderStatusProvider.kt
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Backs strict plugin loading by the Spring environment, instead of using text files.
+ */
+class SpringStrictPluginLoaderStatusProvider(
+  private val environment: Environment
+) {
+
+  fun isStrictPluginLoading(): Boolean =
+    environment.getProperty("spinnaker.extensibility.strict-plugin-loading")?.toBoolean() ?: false
```
jonsie: This should default to true right?

jonsie:
@link108 Ah yea, that makes sense.  I wonder if we should just pass in a boolean to these classes  rather than making it a dedicated class?  If we need to add more configs for this type of "strict" loading stuff we could make it a dedicated class.

link108:
@jonsie I played around with passing a boolean instead of the class and it made testing a bit harder, eg since fixtures are used in testing, we'd be passing a boolean to the constructors, which I found hard to override for specific tests.

jonsie:
@link108 Sounds good, LGTM

PR #574, opened 2020-03-26
fix(telemetry): Instrumented methods map should be concurrent

robzienert:
Should fix https://github.com/spinnaker/spinnaker/issues/5571

If not, we'll have a little better exception to understand what's going on here.

### DIFF
--- a/kork-telemetry/src/main/java/com/netflix/spinnaker/kork/telemetry/InstrumentedProxy.java
+++ b/kork-telemetry/src/main/java/com/netflix/spinnaker/kork/telemetry/InstrumentedProxy.java
@@ -28,6 +28,7 @@ import com.netflix.spinnaker.kork.telemetry.MetricTags.ResultValue;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.*;
 import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 
 /**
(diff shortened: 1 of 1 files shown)


kork-telemetry/src/main/java/com/netflix/spinnaker/kork/telemetry/InstrumentedProxy.java
```diff
@@ -177,7 +178,7 @@ private MethodMetrics getMethodMetrics(Method method) {
         }
       }
 
-      if (!processed && !instrumentedMethods.containsKey(method) && !seenMethods.contains(method)) {
+      if (!processed && !instrumentedMethods.containsKey(method)) {
```
robzienert: This conditional was also incorrect. Just above, the method is added to `seenMethods`, so this would never evaluate to true.

kork-telemetry/src/main/java/com/netflix/spinnaker/kork/telemetry/InstrumentedProxy.java
```diff
@@ -208,12 +209,10 @@ private void addInstrumentedMethod(
     }
 
     for (Map.Entry<Method, MethodMetrics> existingMethodMetric : existingMethodMetrics.entrySet()) {
-      if (existingMethodMetric
-          .getValue()
-          .invocationsId
-          .name()
-          .equals(methodMetrics.invocationsId.name())) {
-        throw new MetricNameCollisionException(target, existingMethodMetric.getKey(), method);
+      String invocationsIdMetricName = existingMethodMetric.getValue().invocationsId.name();
```
ajordens: Not sure if actually treating `existingMethodMetrics` as a `ConcurrentHashMap` and using the `putIfAbsent` semantics would be better/prefable.

In theory that would allow you to avoid this collision check and just safely use what pre-exists.

robzienert: Ah, yes true. I will change. It'll still need to be a CHM, though.

PR #575, opened 2020-03-26
fix(plugins): Clarify error when no httpclientfactory exists supporting baseurl

robzienert:
This error message really threw me for a loop until I realized the `MetatronHttpClientFactory` only supports baseUrls with a scheme of `metatron://`. 🤦‍♂


robzienert:
Closing. Have more fixes for this.

PR #576, opened 2020-03-26
chore(build): Move over relevant Github Action to try to build a kork release for 1.19.x

jonsie:


.github/workflows/release.yml
```diff
@@ -0,0 +1,29 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+-patch[0-9]+"
```
jonsie: I'm not sure if this will work, or if Nebula will choke on this version. If that's the case, perhaps we can create a version like `7.25.12`

jonsie:
@ezimanyi Can you review and merge?  I'm not authorized to merge to this branch.

jonsie:
@ezimanyi Because I made a change to the mergify file here, you'll need to merge it by hand.

PR #582, opened 2020-03-27
fix(build): Remove enablePublishing on branch build, run release build on standard tag format

jonsie:


.github/workflows/release.yml
```diff
@@ -3,7 +3,7 @@ name: Release
 on:
   push:
     tags:
-      - "v[0-9]+.[0-9]+.[0-9]+-patch[0-9]+"
+      - "v[0-9]+.[0-9]+.[0-9]+"
```
jonsie: Tested this out locally, as expected Nebula does not allow this `-patch` format.

PR #586, opened 2020-03-31
fix(Plugins): allow for relaxed plugin loading

link108:


jonsie:
I can't merge it, we'll have to pull in @ezimanyi for that.

PR #588, opened 2020-04-03
fix(plugins): Configured ObjectMapper and better respoonse body handling

robzienert:
Had two problems today. First, the ObjectMapper wasn't configured, oops. Second, `Ok3HttpClient` was closing responses before user-land code would have the opportunity to read response bodies, so I changed all that around.

Also added an integration test for the httpclient, since obviously the unit tests didn't catch it.

### DIFF
--- a/kork-plugins-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/httpclient/HttpClientConfig.java
+++ b/kork-plugins-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/httpclient/HttpClientConfig.java
@@ -34,6 +34,8 @@ public class HttpClientConfig {
   /** Connection pool-related configuration options. */
   private final ConnectionPoolConfig connectionPool = new ConnectionPoolConfig();
 
+  private final LoggingConfig logging = new LoggingConfig();
+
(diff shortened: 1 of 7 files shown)


kork-plugins-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/httpclient/HttpClientConfig.java
```diff
@@ -278,6 +284,40 @@ public int hashCode() {
     }
   }
 
+  public static class LoggingConfig {
```
jonsie, line 287: Would be kinda cool if we could configure some of this on a per-plugin level if we want to debug a plugin's HTTP calls with more granularity.

robzienert: We can! `HttpClientConfig` is provided by the plugin itself.

jonsie: Oh nice, didn't notice this was a change to the kork-plugins-api module. 👍

PR #591, opened 2020-04-03
refactor(metrics/prometheus): Add this dependency in order to expose …

jorgebee65:
Add this dependency in order to expose metrics in Prometheus format for each service adding this property:
```yaml
management:
  endpoints:
    web:
      exposure:
        include: "health,prometheus"
```


robzienert:
I'd rather see this introduced as a system plugin. I really want to stay away from adding micrometer backends to our global dependencies.

cc @spinnaker/plugins-working-group

kork-web/kork-web.gradle
```diff
@@ -9,6 +9,7 @@ dependencies {
   api project(":kork-exceptions")
   api "org.codehaus.groovy:groovy-all"
   api "org.springframework.boot:spring-boot-starter-web"
+  api "io.micrometer:micrometer-registry-prometheus"
```
cfieber, line 12: this should be implementation not api

jonsie:
Does this open the door to us one-by-one pulling in all these various Micrometer registry backends?  That's my main concern.  To add on to @robzienert, we've thought about this and we think this may be a good use case for a Spinnaker system plugin.  If others agree, perhaps we should prioritize building that plugin (or plugins).

robzienert:
Since this is just a new dependency, it seems to me like a perfect case for building out some system plugins for people to use.

robzienert:
Would recommend the use of https://github.com/armory-plugins/armory-observability-plugin instead.

PR #592, opened 2020-04-06
fix(core): Fix micrometer integration by using spectator-reg-micrometer library

Aloren:
feature(web): MetricInterceptor to publish equal set of tags; MetricInterceptor now supports publishing query parameters to metrics (extracted from Kayenta)

We are using standalone Kayenta in our environment. By default Kayenta is exposing all it's internal metrics into Spectator registry -- currently these metrics are available only on `/spectator/metrics` endpoint. This PR makes all Spectator metrics to be available for whatever metrics exporter is used in the app. Please also note that this change also exposes all other metrics that were previously not available on `/spectator/metrics`.


kork-web/src/main/java/com/netflix/spinnaker/kork/web/interceptors/MetricsInterceptor.java
```diff
@@ -111,13 +127,24 @@ public void afterCompletion(
...
+        } else {
+          id = id.withTag(queryParamName, "None");
         }
       }
 
       if (ex != null) {
         id = id.withTag("success", "false").withTag("cause", ex.getClass().getSimpleName());
       } else {
-        id = id.withTag("success", "true");
+        id = id.withTag("success", "true").withTag("cause", "None");
```
Aloren, line 147: We are exporting our metrics to Prometheus and requirement is that all meters with the same name must have the same set of tag keys. I've used default behavior for this kind of metrics from spring mvc metrics filter. If you guys think this is not backward compatible I could probably make it customizable via property. What do you think?

jonsie:
@Aloren Thanks for the contribution!  We've tried this change a few times and have hit a few blockers (last one was due to an upgrade to the underlying Atlas client lib).  This looks correct, and backwards compatibility with spinnaker-monitoring will be maintained here since metrics will still surface at `/spectator/metrics`.  

However, I'd like to deploy this to one of our test environments prior to merging this change.  If all goes well then we'll merge this!

kork-core/kork-core.gradle
```diff
@@ -13,6 +13,7 @@ dependencies {
   api "org.springframework.boot:spring-boot-starter-actuator"
   api "com.netflix.eureka:eureka-client"
   api "com.netflix.spectator:spectator-api"
+  api "com.netflix.spectator:spectator-reg-micrometer"
```
cfieber, line 16: Should be implementation not api

Aloren:
Thank you for the review.

> However, I'd like to deploy this to one of our test environments prior to merging this change. If all goes well then we'll merge this!

I hope so, please let me know once the change is checked :)

kork-core/kork-core.gradle
Aloren: hm.. not sure, why? I suppose, we expect this dependency to be provided by `kork-core` module, otherwise `SpectatorConfiguration.java` will fail with `ClassNotFoundException` if `spectator-reg-micrometer` is not on the classpath.

cfieber: API exports the dependency as a compile time dependency to consumers of kork-core which is unnecessary, they only need it as a runtime dependency.

Most things in here should be implementation not api but that is something I will be addressing separately.

jonsie:
@Aloren This is looking good in our testing environment - I'm not seeing any blockers ( 🤞).  If you can make the change the @cfieber suggests (changing the spectator-reg-micrometer dependency from `api` to `implementation`), then we can merge this!  Thanks again for the contribution.  👍

kork-core/kork-core.gradle
Aloren: fixed

Aloren:
Made the requested changes. Thank you.

PR #595, opened 2020-04-08
feat(plugins): Add kork-plugins-tck module to support testing plugins in services

jonsie:
This also includes two fixes -- 1) To read the Implementation-Version from the package that contains the Spring app and 2) To resolve the root plugins dir if a relative path is provided.

Implementation example:

https://github.com/spinnaker/orca/pull/3579


kork-core/src/main/java/com/netflix/spinnaker/kork/PlatformComponents.java
```diff
@@ -63,9 +61,9 @@ ServiceVersion serviceVersion(
   }
 
   @Bean
-  @ConditionalOnMissingBean(ManifestVersionResolver.class)
-  VersionResolver manifestVersionResolver(ManifestVersionResolver.Properties properties) {
-    return new ManifestVersionResolver(properties.useOssVersionManifestAttribute);
+  @ConditionalOnMissingBean(SpringPackageVersionResolver.class)
+  VersionResolver springPackageVersionResolver(ApplicationContext applicationContext) {
+    return new SpringPackageVersionResolver(applicationContext);
```
jonsie, line 66: `ManifestVersionResolver` was flawed in a few ways.  It assembles URLS to manifest files from the filesystem, but then it tries to match the manifest against a package (`com.netflix.spinnaker.orca`, for example). This works in the unit test when looking up the manifest for `org.springframework.boot` because the dependency comes from the local maven cache which stores deps on the filesystem according to their package (`/org/springframework/boot`). However, this doesn’t work when running a test on a service locally and the service JAR exists at something like `~/orca/orca-web/build/libs/orca-web.jar` or when running in an env wherein the JAR is at some non-package specific path.

kork-core/src/main/java/com/netflix/spinnaker/kork/version/SpringPackageVersionResolver.java
```diff
@@ -0,0 +1,54 @@
...
+
+  public SpringPackageVersionResolver(ApplicationContext applicationContext) {
+    this.applicationContext = applicationContext;
+  }
+
+  @Nullable
+  @Override
+  public String resolve(@Nonnull String serviceName) {
+    Map<String, Object> annotatedBeans =
+        applicationContext.getBeansWithAnnotation(SpringBootApplication.class);
```
jonsie, line 44: We use `SpringBootApplication` annotation on some of our main classes, but not all.  As a follow up to this PR, I will add the annotation where necessary.

kork-plugins/src/main/java/com/netflix/spinnaker/config/PluginsAutoConfiguration.java
```diff
@@ -140,15 +134,24 @@ public static SpinnakerPluginManager pluginManager(
...
+   * returns the default root path 'plugins'.
+   */
+  private static Path determineRootPluginPath(ApplicationContext applicationContext) {
+    String rootPathConfig =
+        applicationContext
+            .getEnvironment()
+            .getProperty(PluginsConfigurationProperties.ROOT_PATH_CONFIG);
+    return rootPathConfig == null
+        ? Paths.get(PluginsConfigurationProperties.DEFAULT_ROOT_PATH)
+        : Paths.get(rootPathConfig).toAbsolutePath();
```
jonsie, line 152: This change allows the path to the root plugins dir to be relative, so that we can easily test plugins in Spinnaker services.

kork-plugins-tck/src/main/kotlin/com/netflix/spinnaker/kork/plugins/tck/PluginsTck.kt
```diff
@@ -0,0 +1,77 @@
...
+import strikt.assertions.isGreaterThanOrEqualTo
+import strikt.assertions.isNull
+import strikt.assertions.isTrue
+
+/**
+ * The default tests that run in a plugin test suite.
+ */
+abstract class PluginsTck<T : PluginsTckFixture> : JUnit5Minutests {
+
+  fun ContextBuilder<T>.defaultPluginTests() {
```
jonsie, line 36: Does this seem like a reasonable set of default tests?

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/SpinnakerPluginManager.kt
```diff
@@ -143,4 +143,8 @@ open class SpinnakerPluginManager(
         it.descriptor.pluginId
       }
   }
+
+  init {
+    systemVersion = getSystemVersion()
```
jonsie, line 148: `systemVersion` needs to be set on initialization, otherwise it is never set and defaults to "0.0.0" in `AbstractPluginManager` which causes all sorts of problems.

kork-core/src/main/java/com/netflix/spinnaker/kork/version/SpringPackageVersionResolver.java
robzienert: If more service alignment falls out of this work, I'm all for it! Do you know if there's any reason to _not_ add that annotation to each `Main`?

kork-plugins-tck/src/main/kotlin/com/netflix/spinnaker/kork/plugins/tck/PluginsTck.kt
robzienert: Yes, I think this is a good start. We'll eventually need tests around the lifecycle as well (verification that start & stop method are invoked when expected, etc). I don't think we need to add these now.

kork-plugins-tck/src/main/kotlin/com/netflix/spinnaker/kork/plugins/tck/PluginsTckFixture.kt
```diff
@@ -0,0 +1,91 @@
...
+   * The PF4J [org.pf4j.PluginManager] implementation, used to assert plugin and extension loading.
+   */
+  val spinnakerPluginManager: SpinnakerPluginManager
+
+  /**
+   * The path to write plugins to the filesystem.  The path should match the configuration:
+   *
+   * spinnaker:
+   *   extensibility:
+   *     plugins-root-path: build/plugins
```
robzienert, line 39: Wrap in code block so that this will render correctly in dokka (same as the rest below)

kork-core/src/main/java/com/netflix/spinnaker/kork/version/SpringPackageVersionResolver.java
jonsie: I couldn't find a reason not to add the annotation, seems harmless.

kork-plugins-tck/src/test/kotlin/com/spinnaker/netflix/kork/plugins/tck/PluginsTckFixtureTest.kt
```diff
@@ -0,0 +1,96 @@
...
+import dev.minutest.rootContext
+import java.io.File
+import org.junit.runner.RunWith
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.autoconfigure.SpringBootApplication
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.context.annotation.Import
+import org.springframework.test.context.junit4.SpringRunner
+
+class PluginsTckFixtureTest : PluginsTck<PluginsTckFixtureImpl>() {
```
jonsie, line 36: Added a PluginsTck test.

PR #603, opened 2020-04-13
feat(plugins): Serde SDK, Service SDK

robzienert:
Working on some higher-level extension points for Orca and I need to:

1. Translate objects from Orca's stage context (a map) into a typed object
2. Expose Orca-specific interfaces for plugins (specifically, a `ClouddriverService`)


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/sdk/serde/SerdeServiceSdkFactory.kt
```diff
@@ -0,0 +1,57 @@
...
+
+  private val serdeService by lazy {
+    applicationContext.getBeansOfType(ObjectMapper::class.java)
+      .values
+      .let {
+        if (it.isEmpty()) {
+          // This should never happen. Obviously it probably will at some point, but it's better to explode than
+          // provide a "reasonable default" since that seems to change depending on who you ask. Better to ensure
+          // a service has wired up an ObjectMapper to matches what it has defined as reasonable defaults.
+          throw SystemException("Failed to locate ObjectMapper in application context")
```
jonsie, line 39: 👍

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/sdk/serde/SerdeServiceSdkFactory.kt
```diff
@@ -0,0 +1,57 @@
...
+        } else {
+          if (it.size > 1) {
+            // This is an indication that the service has multiple object mappers available, which will introduce
+            // uncertainty in exactly how objects will be serialization/deserialized. We should be converging on one
+            // ObjectMapper instance in the ApplicationContext in services, so this is really up to service maintainers
+            // to resolve if it occurs, but it isn't the end of the world since most of our ObjectMappers are
+            // configured roughly the same.
+            log.warn("Found more than one ObjectMapper, selecting first")
```
jonsie, line 47: If there's more than one ObjectMapper bean in the context, I'm assuming they'll have unique bean names.  Maybe we should log out the bean name here?

PR #608, opened 2020-04-14
refactor(dependencies): Reduce transitive dependencies

ezimanyi:
A number of commonly-used kork modules export their implementation dependencies as api dependencies. In cases where a depdendency is not used in the API of a module and is just an implementation detail, move it to an implementation dependency. (In some cases the dependency could be removed entirely.)

There are definitely many other places in kork where we can clean things up (reducing things on the compile path in all services) but this is just a small thing I could get done quickly.

I've already updated the microservices (or have an open PR) to explicilty add in dependencies in cases where we were depending on an accidental transitive dependency.


ezimanyi:
spinnaker/gate#1150 is the only remaining PR that should be merged before we cut a kork release.

I've tested all the microservices with this change and they all compile and tests pass.

PR #611, opened 2020-04-16
fix(plugins): consider request params while building ok http client

srekapalli:
- Adding supplied request params to the URL.


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/sdk/httpclient/Ok3HttpClient.kt
```diff
@@ -93,11 +94,18 @@ class Ok3HttpClient(
     }
   }
 
-  private fun requestBuilder(request: Request): okhttp3.Request.Builder =
-    okhttp3.Request.Builder()
+  private fun requestBuilder(request: Request): okhttp3.Request.Builder {
+    val url = (baseUrl + request.path).replace("//", "/")
+    val httpUrlBuilder = HttpUrl.parse(url)?.newBuilder()
+      ?: throw IntegrationException("Unable to parse url '$baseUrl'")
+    for (param in request.queryParams) {
```
jonsie, line 101: Opinionated nit: 
```request.queryParams.forEach { httpUrlBuilder.addQueryParameter(it.key, it.value) }```

PR #621, opened 2020-04-21
feat(plugins): Adding server group plugin version pinning

robzienert:
Depends on: https://github.com/spinnaker/front50/pull/791

### DIFF
--- a/kork-plugins/src/main/java/com/netflix/spinnaker/config/Front50PluginsConfiguration.java
+++ b/kork-plugins/src/main/java/com/netflix/spinnaker/config/Front50PluginsConfiguration.java
@@ -22,14 +22,21 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.module.kotlin.KotlinModule;
 import com.netflix.spinnaker.config.PluginsConfigurationProperties.PluginRepositoryProperties;
+import com.netflix.spinnaker.kork.plugins.update.EnvironmentServerGroupLocationResolver;
+import com.netflix.spinnaker.kork.plugins.update.EnvironmentServerGroupNameResolver;
+import com.netflix.spinnaker.kork.plugins.update.ServerGroupLocationResolver;
+import com.netflix.spinnaker.kork.plugins.update.ServerGroupNameResolver;
 import com.netflix.spinnaker.kork.plugins.update.downloader.FileDownloaderProvider;
 import com.netflix.spinnaker.kork.plugins.update.downloader.Front50FileDownloader;
 import com.netflix.spinnaker.kork.plugins.update.internal.Front50Service;
 import com.netflix.spinnaker.kork.plugins.update.internal.PluginOkHttpClientProvider;
+import com.netflix.spinnaker.kork.plugins.update.release.source.Front50PluginInfoReleaseSource;
+import com.netflix.spinnaker.kork.plugins.update.release.source.PluginInfoReleaseSource;
 import com.netflix.spinnaker.kork.plugins.update.repository.Front50UpdateRepository;
 import com.netflix.spinnaker.okhttp.OkHttpClientConfigurationProperties;
 import java.net.URL;
 import java.util.Map;
+import java.util.Objects;
 import okhttp3.OkHttpClient;
 import org.pf4j.update.UpdateRepository;
 import org.pf4j.update.verifier.CompoundVerifier;
@@ -41,6 +48,7 @@ import org.springframework.boot.context.properties.bind.Bindable;
 import org.springframework.boot.context.properties.bind.Binder;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Primary;
 import org.springframework.core.env.Environment;
 import retrofit2.Retrofit;
 import retrofit2.converter.jackson.JacksonConverterFactory;
@@ -84,12 +92,10 @@ public class Front50PluginsConfiguration {
   }
 
   @Bean
-  public static UpdateRepository pluginFront50UpdateRepository(
+  public static Front50Service pluginFront50Service(
       Environment environment,
       PluginOkHttpClientProvider pluginsOkHttpClientProvider,
-      Map<String, PluginRepositoryProperties> pluginRepositoriesConfig,
-      FileDownloaderProvider fileDownloaderProvider) {
-
+      Map<String, PluginRepositoryProperties> pluginRepositoriesConfig) {
     PluginRepositoryProperties front50RepositoryProps =
         pluginRepositoriesConfig.get(PluginsConfigurationProperties.FRONT5O_REPOSITORY);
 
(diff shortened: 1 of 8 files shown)


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/release/VersionPinningPluginInfoReleaseProvider.kt
```diff
@@ -0,0 +1,81 @@
...
+ * loaded. These resolved plugins are then sent to Front50 for pinning.
+ *
+ * If the server group has not been seen before, Front50 will save the service-resolved versions as the pinned versions
+ * for the server group. If the server group has been seen before, Front50 will return a corrected set of plugin
+ * versions back to the service, which it will use instead.
+ *
+ * Should any part of this process fail, the service-selected versions will be used as a fallback, making this
+ * functionality optional.
+ */
+class VersionPinningPluginInfoReleaseProvider(
```
jonsie, line 37: What if we call this `Front50PluginInfoReleaseProvider`?  Similar to how we have the `Front50UpdateRepository`.

robzienert: That makes sense to me. 👍

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/release/Front50PluginInfoReleaseProvider.kt
```diff
@@ -0,0 +1,81 @@
...
+class Front50PluginInfoReleaseProvider(
+  private val releaseProvider: PluginInfoReleaseProvider,
+  private val front50Service: Front50Service,
+  private val serverGroupNameResolver: ServerGroupNameResolver
+) : PluginInfoReleaseProvider {
+
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  override fun getReleases(pluginInfo: List<PluginInfo>): Set<PluginInfoRelease> {
+    val resolvedPlugins = releaseProvider.getReleases(pluginInfo)
```
jonsie, line 46: Instead of passing in a `PluginInfoReleaseProvider`, what if we pass a list of these providers to a resolver class? Then in `ExtensionBeanDefinitionRegistryPostProcessor` we use said resolver?  Then we have this `Front50PluginInfoReleaseProvider` enabled behind a config (since we should probably roll it out as optional?).

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/release/Front50PluginInfoReleaseProvider.kt
```diff
@@ -0,0 +1,81 @@
...
+import org.pf4j.update.PluginInfo
+import org.slf4j.LoggerFactory
+
+/**
+ * Attempts to use Front50 to pin plugin versions that are loaded for a particular server group.
+ *
+ * When a service starts up, it will load local plugins, reaching out to an UpdateRepository for additional
+ * plugins to download. The service will then use [releaseProvider] to resolve the plugin versions that should be
+ * loaded. These resolved plugins are then sent to Front50 for pinning.
+ *
```
srekapalli, line 29: In what scenario , will we see a drift between what the service is resolving(a plugin version) to and what front50 says ?

For the service, source of truth is in the config bundled and that is what gets pull down and loaded. 

Day #1
Initial plugin v0.0.1 , ASG V001 -- registers with Front50 .
Day #2 
Autoscale happens . config says v0.0.1 and Front50 says v0.0.1 ( everything is good )
Day #3
New commit happens in the service config now plugin version moves to v0.0.2 , gets launched into a new ASG V002 and registers with Front50, but operator can decide to stay on v0.0.1.  Something like that..?

jonsie: I think there would be a record in front50 that operators maintain like "latest" for a service.  This way we don't automatically bump folks to the newest plugin version. This addresses the "day #3" scenario, I think.

srekapalli: When do we bump to the latest version..?  If the service asks for it..?

Something is not adding up.

jonsie: So basically, if a plugin version doesn't resolve - then we ask front50 for the "latest".  If there isn't a latest record for that plugin, we just fallback to the most recent version according to plugin info.  Does that make sense?

robzienert: Keep in mind, no service YAML config is immutable: _Every config property in Spinnaker can be influenced by Fast Properties._

This is forward-compatible with a Spinnaker Admin UI which will manage what plugins are available and what version those plugins should be regardless of the latest version available or what the service config says should be loaded. Once the admin UI is available, front50 will be the SOT of what plugin version to load: Not the service config (or Fast Properties).

In today's world, you are correct, the service's config will be authoritative with front50 only being an alignment mechanism. If no version is specified in the service config, latest release is always selected.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/release/Front50PluginInfoReleaseProvider.kt
jonsie: This comment is mostly around the concept of adding a 3rd `PluginInfoReleaseProvider` and for now I think that's probably not necessary so let's just leave it be. 👍

robzienert:
Ready for review.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/release/Front50PluginInfoReleaseProvider.kt
```diff
@@ -0,0 +1,87 @@
...
+) : PluginInfoReleaseProvider {
+
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  override fun getReleases(pluginInfo: List<PluginInfo>): Set<PluginInfoRelease> {
+    val resolvedPlugins = releaseProvider.getReleases(pluginInfo)
+
+    val serverGroupName = serverGroupNameResolver.get()
+    if (serverGroupName == null) {
+      log.warn("Could not resolve my server group name: Plugin versioning is disabled")
```
jonsie, line 51: Log is a bit misleading.  Probably something like ```Could not resolve my server group name, skipping front50 plugin version resolution.```

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/release/Front50PluginInfoReleaseProvider.kt
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  private fun Set<PluginInfoRelease>.toPinnedVersions(): PinnedVersions =
+    PinnedVersions(map { it.pluginId to it.props.version }.toMap())
+
+  private fun PinnedVersions.updateReleases(pluginInfoReleases: Set<PluginInfoRelease>) {
+    pluginInfoReleases.forEach {
+      pluginVersions[it.pluginId]?.let { version ->
+        if (it.props.version != version) {
+          log.info("Aligning plugin '${it.pluginId}' to pinned version: ${it.props.version} -> $version")
```
jonsie, line 81: I would include the server group name in this log too.

kork-plugins/src/main/java/com/netflix/spinnaker/config/Front50PluginsConfiguration.java
```diff
@@ -118,6 +134,20 @@ public static UpdateRepository pluginFront50UpdateRepository(
         front50Service);
   }
 
+  @Bean
+  @Primary
+  public static PluginInfoReleaseProvider front50PluginReleaseProvider(
```
jonsie, line 139: We will probably want to disable this PluginInfoReleaseProvider until the front50 side is in place.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/release/Front50PluginInfoReleaseProvider.kt
```diff
@@ -0,0 +1,95 @@
...
+          it.props.version = version
```
jonsie, line 90: Hm.  Just realized we'll need to change `props` to a `var` in `PluginInfoRelease`.  Front50 will also need to return the PF4J `PluginInfo.Release` object here.  We can't simply swap out the `version`, because just the version field will change but we'll still end up downloading the artifact set in the `url` field.

robzienert: 🤦 🤦 🤦 🤦 🤦 

Nice catch.

PR #625, opened 2020-04-27
fix(plugins): don't blow up if plugin config does not exist and plugi…

danielpeach:
…n config class does not have no-args constructor.

Reasonable-looking config classes (https://github.com/spinnaker-plugin-examples/pf4jStagePlugin/blob/master/random-wait-orca/src/main/kotlin/io/armory/plugin/stage/wait/random/RandomWaitConfig.kt) don't load if you don't specify any config.


jonsie:
🤔 

Spring introduced `@ConstructorBinding` to address this:

https://docs.spring.io/spring-boot/docs/2.2.0.RELEASE/api/org/springframework/boot/context/properties/ConstructorBinding.html

I wonder if long term we will need to add something similar for configuration classes?

PR #628, opened 2020-04-28
chore(secrets): Don't swallow stacktrace of exceptions

german-muzquiz:
When there's an exception downloading secrets from S3, the original stacktrace is lost because the returned exception only includes the message.
It's good practice to include the full original `Exception` object, so that we can get full stacktraces for troubleshooting.


robzienert:
Looks like this PR needs to be updated? I can't do it for you.

german-muzquiz:
PR updated, but bot didn't merge :/

robzienert:
I think it's an issue of the armory-io forks not having correct fork permissions. This is a regular occurrence with Armory submissions.

PR #632, opened 2020-05-04
feat(plugins): Add SpinnakerExtensionPoint and default getExtensionClass implementation

jonsie:
This allows us to call `extension.getExtensionClass()` when we need to retrieve the underlying extension class.


kork-plugins-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/internal/ExtensionClassProvider.java
```diff
@@ -12,23 +12,22 @@
...
-/**
- * For use in service code to determine the proxied target extension class.
- */
-object ExtensionClassProvider {
-  @JvmStatic
-  fun getExtensionClass(extensionPoint: ExtensionPoint): Class<*> {
-    if (Proxy.isProxyClass(extensionPoint.javaClass)) {
-      val extensionInvocationProxy = Proxy.getInvocationHandler(extensionPoint) as ExtensionInvocationProxy
-      return extensionInvocationProxy.getTargetClass()
+class ExtensionClassProvider {
```
jonsie, line 22: This is package-private.  There should be no real reason to expose this to consumers of kork-plugins-api.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/ExtensionBeanDefinitionRegistryPostProcessor.kt
```diff
@@ -108,14 +110,16 @@ class ExtensionBeanDefinitionRegistryPostProcessor(
...
-        val bean = pluginManager.extensionFactory.create(extensionClass)
+        val extensionClassInstance = pluginManager.extensionFactory.create(extensionClass)
+
+        val bean = if (extensionClassInstance is SpinnakerExtensionPoint) {
+          ExtensionInvocationProxy.proxy(
+            extensionClassInstance,
+            invocationAspects as List<InvocationAspect<InvocationState>>,
+            plugin.descriptor as SpinnakerPluginDescriptor)
+        } else {
+          extensionClassInstance
```
jonsie, line 121: This allows us to support cases wherein an extension point is an interface that does not implement `SpinnakerExtensionPoint` (which PF4J allows with some limitations) and serves the other purpose of allowing us to migrate to `SpinnakerExtensionPoint`.

kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/update/release/source/LatestPluginInfoReleaseSourceTest.kt
```diff
@@ -59,8 +59,8 @@ class LatestPluginInfoReleaseSourceTest : JUnit5Minutests {
 
       expectThat(releases).isA<Set<PluginInfoRelease>>()
         .get { releases.size }.isEqualTo(2)
-        .get { releases.find { it?.pluginId == plugin1.id } }.isEqualTo(PluginInfoRelease(plugin1.id, plugin1ExpectedRelease))
-        .get { releases.find { it?.pluginId == plugin2.id } }.isEqualTo(PluginInfoRelease(plugin2.id, plugin2ExpectedRelease))
+        .get { releases.find { it.pluginId == plugin1.id } }.isEqualTo(PluginInfoRelease(plugin1.id, plugin1ExpectedRelease))
+        .get { releases.find { it.pluginId == plugin2.id } }.isEqualTo(PluginInfoRelease(plugin2.id, plugin2ExpectedRelease))
```
jonsie, line 63: Unrelated clean up to remove some warnings when running kork-plugins tests.

kork-plugins-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/internal/SpinnakerExtensionPoint.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.kork.plugins.api.internal;
+
+import org.pf4j.ExtensionPoint;
+
+public interface SpinnakerExtensionPoint extends ExtensionPoint {
```
cfieber, line 22: It would be worth throwing some javadoc on here, it took me a few minutes (and looking at the orca PR) to grok that this exists as the public API to things that generically want to deal with extension points and need access to the underyling class (for a minute I thought this was only needed to wire up all the proxy magic properly)

jonsie: Oh duh, this would be a pretty important thing to document.  I'll do that.

kork-plugins-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/internal/ExtensionInvocationHandler.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.kork.plugins.api.internal;
+
+import java.lang.reflect.InvocationHandler;
+
+public interface ExtensionInvocationHandler extends InvocationHandler {
```
robzienert, line 22: Docs on all interfaces and their methods.

PR #633, opened 2020-05-04
feat(plugins): Add Meter annotation, limit metering extension invocation to methods annotated with @Meter

jonsie:


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/proxy/aspects/MetricInvocationAspect.kt
```diff
@@ -123,36 +125,41 @@ class MetricInvocationAspect(
...
+          .find { it is Meter }
+          .let { meterAnnotation ->
+            (meterAnnotation as Meter)
+
+            val metricIds = MetricIds(
+              timingId = registry.createId(toMetricId(m, descriptor.pluginId, meterAnnotation.id, TIMING), mapOf(
+                Pair("pluginVersion", descriptor.version),
+                Pair("pluginExtension", target.javaClass.simpleName.toString())
+              )),
+              invocationId = registry.createId(toMetricId(m, descriptor.pluginId, meterAnnotation.id, INVOCATIONS), mapOf(
```
cfieber, line 138: I think you implicitly get invocation counts already from a timer metric (it keeps track of total time recorded and the count)

robzienert:
I'm still awfully confused why we're not re-using the `@Instrumented` stuff I did and that is used elsewhere within Spinnaker: https://github.com/spinnaker/kork/blob/master/kork-telemetry/src/main/java/com/netflix/spinnaker/kork/telemetry/Instrumented.java

jonsie:
@robzienert Just the annotation or the `InstrumentedProxy` as well?  I guess I don't mind sharing the same annotation, but the `InstrumentedProxy` implementation and the `ExtensionInvocationProxy` and `MetricInvocationAspect` patterns would need to be refactored to support use cases wherein the proxied object is either an extension or any other class in Spinnaker.  I thought about it - we could do it, but a slight amount of duplication here seemed easier to reason about and less error prone in the long run.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/proxy/aspects/MetricInvocationAspect.kt
jonsie: Well them, I shall remove this.

jonsie:
Merging, we can continue the `@Instrumented` and `InstrumentedProxy` discussion though.

robzienert:
I meant both, but mostly the annotation. As of this PR being merged, we've now gone from having one way of doing things to now two, incompatible ways of doing things. Contributors now will need to know when to use one versus the other, which I think is a very poor experience. I would prefer to see a single annotation, at minimum, and a convergence of both proxies if possible.

jonsie:
@robzienert I'll do the work to share the annotation.

As for the `MetricInvocationAspect` vs. `InstrumentedProxy` there's a few methods that we could share, but otherwise it's built on a different pattern that allows for extension proxy flexibility and proxied extension functionality in services (see https://github.com/spinnaker/kork/pull/632 for example).

PR #636, opened 2020-05-06
chore(annotations): Consolidate @Meter and @Instrumented to one annotation, @Metered, in kork-annotations

jonsie:
Additionally, this introduces a helper class MethodInstrumentation with common functions and exceptions for use when instrumenting methods with metric IDs and tags.

I'll have to swap out `@Instrumented` for `@Metered` where it's used, which looks to be here:

https://github.com/spinnaker/orca/blob/master/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/persistence/ExecutionRepository.java


robzienert:
Tasty.

Just an FYI that this autobump will break on services - new imports will need to be made. Not a problem, just a notice. I know Orca for sure, but I'm not sure if other services have used the annotations.

PR #639, opened 2020-05-07
fix(misc): allow pluggable client providers

srekapalli:
To allow customizations of OkHttpClient per destination/host, we need to provide ways to customize the client before the client gets used/wired in for making calls. Current implementation as provided in `OkHttp3ClientConfiguration` doesn't have enough hooks to allow customization of the client and this change is mostly to address that. Also to keep backward compatibility and to not have to propagate this change to every service at the same time, I'm keeping the existing OkHttp3ClientConfiguration contracts/methods as is so that we can migrate every service to use the new `Provider` implementation individually and not do all at once.
Once we modify every service to use the new `Provider` implementation, we can do some clean up , and remove the use of `OkHttp3ClientConfiguration`  and also refactor other places where we create a brand new client instead of using the singelton.

Another change in here is to use Service configuration object as the main way of defining a dependency on a spinnaker internal or external service.
```
@Autowired
OkHttpClientProvider clientProvider;

OkHttpClient client = clientProvider.getClient(serviceConfig);

client.sendRequest(serviceConfig.sendUrl)........// or inject the url and client into `Retrofit` client builders

```
We also need to revisit how we are constructing clients in the `plugin sdk` side of things and consolidate the factories/builders etc. The change is backward compatible and currently we autowire the `OkHttp3ClientConfiguration` and use that to build client impls using Retrofit or otherwise and with this change we will use the `clientProvider` and use that to build customized clients by Url.


kork-web/src/main/java/com/netflix/spinnaker/okhttp/OkHttpClientBuilderProvider.java
```diff
@@ -0,0 +1,60 @@
...
+  private final OkHttpClientConfigurationProperties okHttpClientConfigurationProperties;
+
+  public OkHttpClientBuilderProvider(
+      OkHttpClient okHttpClient,
+      OkHttpClientConfigurationProperties okHttpClientConfigurationProperties) {
+    this.builder = okHttpClient.newBuilder();
+    this.okHttpClientConfigurationProperties = okHttpClientConfigurationProperties;
+  }
+
+  /** Returns whether or not the provider supports the provided [baseUrl]. */
```
ajordens, line 33: nit ... isn't `[baseUrl]` incorrect syntax for javadoc in java (vs. kotlin)?

kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttp3ClientConfiguration.groovy
```diff
@@ -38,50 +35,30 @@ import javax.net.ssl.X509TrustManager
...
-  private final OkHttpClientConfigurationProperties okHttpClientConfigurationProperties
-  private final OkHttp3MetricsInterceptor okHttp3MetricsInterceptor
+class OkHttp3ClientConfiguration extends OkHttpClientBuilderProvider {
 
   @Autowired
-  public OkHttp3ClientConfiguration(OkHttpClientConfigurationProperties okHttpClientConfigurationProperties,
-                                    OkHttp3MetricsInterceptor okHttp3MetricsInterceptor) {
-    this.okHttpClientConfigurationProperties = okHttpClientConfigurationProperties
-    this.okHttp3MetricsInterceptor = okHttp3MetricsInterceptor
+  OkHttp3ClientConfiguration(OkHttpClient okHttpClient,
```
ajordens, line 45: Thoughts on just accepting an `OkHttpClient.Builder` since that seems to primarily what the class is dealing with?

kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttp3ClientConfiguration.groovy
```diff
@@ -38,50 +35,30 @@ import javax.net.ssl.X509TrustManager
...
 import java.security.SecureRandom
-import java.util.concurrent.TimeUnit
 
 @Slf4j
 @CompileStatic
 @Component
-class OkHttp3ClientConfiguration {
-  private final OkHttpClientConfigurationProperties okHttpClientConfigurationProperties
-  private final OkHttp3MetricsInterceptor okHttp3MetricsInterceptor
+class OkHttp3ClientConfiguration extends OkHttpClientBuilderProvider {
```
ajordens, line 42: Seems like in the future this class might be misnamed?
(would it eventually become `DefaultOkHttpClientBuilderProvider`?)

kork-web/src/main/java/com/netflix/spinnaker/config/OkHttpClientProvider.java
```diff
@@ -0,0 +1,73 @@
...
+import static java.lang.String.format;
+
+import com.netflix.spinnaker.kork.exceptions.SystemException;
+import com.netflix.spinnaker.okhttp.OkHttpClientBuilderProvider;
+import java.util.List;
+import okhttp3.OkHttpClient;
+import org.springframework.stereotype.Component;
+
+@Component
+class OkHttpClientProvider {
```
ajordens, line 28: Naming convention Q ... historically we've used `OkHttpClient3` to distinguish.

Is it more or less confusing not to do that here?

kork-web/src/main/java/com/netflix/spinnaker/config/OkHttpClientProvider.java
```diff
@@ -0,0 +1,73 @@
...
+   * @return okHttpClient
+   */
+  OkHttpClient getClient(String endPointUrl) {
+
+    OkHttpClientBuilderProvider provider = findProvider(endPointUrl);
+    if (provider == null) {
+      throw new SystemException(format("Failed to create HTTP client for url (%s)", endPointUrl));
+    }
+
+    provider.create();
```
ajordens: Any reason why `provider.create()` couldn't take `endPointUrl` and setup the host name verifier? Why does it need to be separately invoked?

kork-web/src/main/java/com/netflix/spinnaker/config/OkHttpClientProvider.java
```diff
@@ -0,0 +1,73 @@
...
+    provider.applyHostNameVerifier(endPointUrl);
+
+    return provider.getBuilder().build();
```
ajordens: `provider.build()` vs `provider.getBuilder().build()`?

Might be irrelevant.

kork-web/src/main/java/com/netflix/spinnaker/config/OkHttpClientProvider.java
```diff
@@ -0,0 +1,73 @@
...
+
+  /**
+   * Get normalized URL as decided by the provider that can serve the presented Url.
+   *
+   * @param endPointUrl
+   * @return
+   */
+  String getNormalizedUrl(String endPointUrl) {
+    OkHttpClientBuilderProvider provider = findProvider(endPointUrl);
+    return (provider != null) ? provider.getNormalizedUrl(endPointUrl) : null;
```
ajordens, line 64: What's the impact on returning `null` here ... we're throwing an exception above if we cannot find a provider.

kork-web/src/main/java/com/netflix/spinnaker/config/OkHttpClientProvider.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    provider.create();
+    provider.applyHostNameVerifier(endPointUrl);
+
+    return provider.getBuilder().build();
+  }
+
+  /**
+   * Get normalized URL as decided by the provider that can serve the presented Url.
```
ajordens, line 57: nit `URL` vs `Url` when it should just be `url`.

kork-web/src/main/java/com/netflix/spinnaker/config/RawOkHttpClientConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+import okhttp3.OkHttpClient;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@EnableConfigurationProperties(OkHttpClientConfigurationProperties.class)
+class RawOkHttpClientConfiguration {
+
+  /** OkHttpClient instance to be reused it for all HTTP calls. */
```
ajordens, line 30: Javadoc could potentially be improved.

```
/**
 * Default {@link OkHttpClient} that is correctly configured for service-to-service communication.
 */
```

^^ can likely be further improved (if we actually want/need javadoc on `@Bean`'s).

kork-web/src/main/java/com/netflix/spinnaker/config/RawOkHttpClientFactory.java
```diff
@@ -0,0 +1,62 @@
...
+import com.netflix.spinnaker.okhttp.OkHttpClientConfigurationProperties;
+import java.util.concurrent.TimeUnit;
+import okhttp3.ConnectionPool;
+import okhttp3.Dispatcher;
+import okhttp3.OkHttpClient;
+
+public class RawOkHttpClientFactory {
+
+  /**
+   * Returns a basic client which can be further customized for other needs upstream.(eg: SSL setup,
```
ajordens, line 29: nit javadoc needs some spacing ... at least move `(eg: ...` onto a separate line.

Not sure what "upstream" means in this context. I think we mean something along the lines of subsequently customized depending on destination?

kork-web/src/main/java/com/netflix/spinnaker/okhttp/OkHttpClientBuilderProvider.java
```diff
@@ -0,0 +1,60 @@
...
+  public abstract Boolean supports(String baseUrl);
+
+  /** Allows custom implementations to adjust the url before making a call. */
+  public String getNormalizedUrl(String baseUrl) {
+    return baseUrl;
+  }
+
+  /** Apply host name verifier for the provided [baseUrl] if any */
+  public OkHttpClient.Builder applyHostNameVerifier(String baseUrl) {
```
ajordens: Mentioned in an earlier comment but this may not be necessary.

kork-web/src/main/java/com/netflix/spinnaker/okhttp/OkHttpClientBuilderProvider.java
```diff
@@ -0,0 +1,60 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.okhttp;
+
+import okhttp3.OkHttpClient;
+
+public abstract class OkHttpClientBuilderProvider {
```
ajordens, line 21: Is there anything here that would stop this from actually being an interface rather than abstract class?

kork-web/src/main/java/com/netflix/spinnaker/okhttp/OkHttpClientBuilderProvider.java
```diff
@@ -0,0 +1,60 @@
...
+    return this.builder;
+  }
+
+  /**
+   * Creates an [OkHttpClient] Builder. The generated client builder is not set to talk to any
+   * endpoint and it just constructs a bare minimum client
+   */
+  public abstract OkHttpClient.Builder create();
+
+  public OkHttpClient.Builder getBuilder() {
```
ajordens: Having two methods that return an `OkHttpClient.Builder` might be confusing?

kork-web/src/main/java/com/netflix/spinnaker/okhttp/OkHttpClientBuilderProvider.java
```diff
@@ -50,8 +50,8 @@ public String getNormalizedUrl(String baseUrl) {
    */
   public abstract OkHttpClient.Builder create();
 
-  public OkHttpClient.Builder getBuilder() {
-    return this.builder;
+  public OkHttpClient getClient() {
+    return this.okHttpClient;
```
ajordens, line 54: Confusing to have a method that returns `OkHttpClient.Builder` and another that returns a full blown `OkHttpClient`?

kork-web/src/main/java/com/netflix/spinnaker/config/RawOkHttpClientConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+  @Bean
+  OkHttpClient okHttpClient(
+      OkHttpClientConfigurationProperties okHttpClientConfigurationProperties,
+      OkHttp3MetricsInterceptor okHttp3MetricsInterceptor) {
```
ajordens, line 34: More generically could this be a collection of interceptors rather than this specific one?

kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttp3ClientConfiguration.groovy
srekapalli: Yes. We need to rename after we migrate every service to use the new provider impl

kork-web/src/main/java/com/netflix/spinnaker/config/OkHttpClientProvider.java
srekapalli: Good point. I think we can overload that method and leave the existing impl for backward compatibility.  Trying to fit in all the abstractions , the interface is not looking clean.

kork-web/kork-web.gradle
```diff
@@ -1,6 +1,13 @@
 apply plugin: "java-library"
 apply plugin: "groovy"
 
+sourceSets {
+  main {
+    java { srcDirs = [] }
+    groovy { srcDirs += ['src/main/java'] }
```
robzienert, line 7: Is this actually necessary? I don't really like the idea of expanding the use of the Groovy compiler.

kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttp3ClientConfiguration.groovy
robzienert: +1, I think accepting a builder is probably better.

kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttp3ClientConfiguration.groovy
robzienert: What's preventing us from renaming it now?

kork-web/src/main/java/com/netflix/spinnaker/config/OkHttpClientProvider.java
```diff
@@ -0,0 +1,73 @@
...
+   * and delegates to that provider to build a client.
+   *
+   * @param endPointUrl
+   * @return okHttpClient
+   */
+  OkHttpClient getClient(String endPointUrl) {
+
+    OkHttpClientBuilderProvider provider = findProvider(endPointUrl);
+    if (provider == null) {
+      throw new SystemException(format("Failed to create HTTP client for url (%s)", endPointUrl));
```
robzienert, line 47:
```suggestion
      throw new SystemException(format("Failed to create HTTP client: No provider found for endpoint '%s'", endPointUrl));
```

kork-web/src/main/java/com/netflix/spinnaker/okhttp/OkHttpClientBuilderProvider.java
robzienert: +1 this definitely looks like it should be an interface. I'd also like to see docs on the interface itself.

kork-web/src/main/java/com/netflix/spinnaker/okhttp/OkHttpClientBuilderProvider.java
robzienert: Agreed. I would not expose `OkHttpClient` at all.

robzienert:
Good step! I like that we'll be passing around a provider instead of a "Configuration" class that has been hijacked for new purposes - good cleanup.

kork-web/src/main/java/com/netflix/spinnaker/okhttp/OkHttpClientBuilderProvider.java
srekapalli: Wanted to provide constructor with the dependencies that need to get autowired

kork-web/kork-web.gradle
srekapalli: Without this , I was unable to refer the existing groovy in a new java source and we don't want to add any more groovy source. 🤷‍♂️

kork-web/src/main/java/com/netflix/spinnaker/okhttp/OkHttpClientBuilderProvider.java
srekapalli: Removed.

kork-web/kork-web.gradle
robzienert: Oh, right. Ok.

kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttp3ClientConfiguration.groovy
srekapalli: If we rename it now, we have to push changes into all services at the same time. I was hoping to avoid changes  to all services at once.

kork-web/src/main/java/com/netflix/spinnaker/config/OkHttpClientProvider.java
srekapalli: Fixed now.

kork-web/src/main/java/com/netflix/spinnaker/config/RawOkHttpClientConfiguration.java
srekapalli: Accepts list of interceptors.

kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttp3ClientConfiguration.groovy
jonsie: I think it's worth taking the hit and renaming this class.  It took me a moment to realize this is more like a `DefaultOkHttpClientBuilderProvider` as I initially scanned through this.

kork-web/src/main/java/com/netflix/spinnaker/okhttp/OkHttpClientBuilderProvider.java
```diff
@@ -0,0 +1,73 @@
...
+  public OkHttpClient.Builder applyHostNameVerifier(OkHttpClient.Builder builder, String baseUrl) {
+    // Concrete impls will override.
+    return builder;
+  }
+
+  /**
+   * Creates a new OkHttpClient Builder from the client. The generated client builder is not set to
+   * talk to any endpoint and it just constructs a bare minimum client
+   */
+  public OkHttpClient.Builder create() {
```
jonsie, line 62: nit: I suggest renaming these to `get` as it's more idiomatic on a provider.  It's always a bit strange to call `create` on a builder provider, and then call `create` again on the builder.

kork-web/src/main/java/com/netflix/spinnaker/okhttp/OkHttpClientBuilderProvider.java
```diff
@@ -0,0 +1,73 @@
...
+    return okHttpClient.newBuilder();
+  }
+
+  public OkHttpClient.Builder create(String url) {
```
jonsie, line 66: Same comment as above.

kork-web/src/main/java/com/netflix/spinnaker/config/OkHttpClientProvider.java
srekapalli: @ajordens :  Made some changes to take in url as arg.

kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttp3ClientConfiguration.groovy
srekapalli: I left the original OkHttp3ClientConfiguration as is and pretty much copied the impl into the new impl (DefaultOkHttpClientProvider). We can roll this out w/o breaking all the apps.

kork-web/src/main/java/com/netflix/spinnaker/config/OkHttpClientProvider.java
srekapalli: I moved all the impls to okhttp3 package.

kork-web/src/main/java/com/netflix/spinnaker/okhttp/OkHttpClientBuilderProvider.java
srekapalli: Removed this method completely , we will do this as part of ssl builder.

kork-web/src/main/java/com/netflix/spinnaker/okhttp/OkHttpClientBuilderProvider.java
srekapalli: Ok. Converted to interface.

kork-web/src/main/java/com/netflix/spinnaker/okhttp/OkHttpClientBuilderProvider.java
srekapalli: Cleaned up the interface.

kork-web/src/main/java/com/netflix/spinnaker/config/okhttp3/OkHttpClientBuilderProvider.java
```diff
@@ -0,0 +1,53 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.config.okhttp3;
+
+import okhttp3.OkHttpClient;
+
+public interface OkHttpClientBuilderProvider {
```
cfieber, line 22: would it be worth grabbing something like the [Service](https://github.com/spinnaker/gate/blob/master/gate-core/src/main/groovy/com/netflix/spinnaker/gate/config/Service.groovy) object from gate, and making that the thing passed to this builder?

It feels like as it stands if you wanted some sort of switching behavior based on the URL you would need to encode the configuration of that into the URL itself (which is why we then have to normalize it back to the actual URL).

Maybe the config could just come along in a dedicated object?

srekapalli: Yes. Didn't feel good to pass along URL to switch behavior and kind influenced by what's out there currently :(.

kork-web/src/main/java/com/netflix/spinnaker/config/ServiceConfigurationProperties.java
```diff
@@ -0,0 +1,177 @@
...
+package com.netflix.spinnaker.config;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties
+public class ServiceConfigurationProperties {
```
cfieber, line 27: I'm not totally sure we should lift this entire structure out of gate - there is a lot of stuff in here that seems overly specific and that we won't use generically in all the services.

Maybe just a Service interface but not the opinion on how to map that class from configuration (at least to start with)

And maybe all the Service interface has to supply out of the box is a getUrl method.

For our internal needs we can extend that to something that includes additional configuration.

srekapalli: Agree, some config around sharding etc felt too specific and some configs like healthCheckableSerivces etc felt like we can reuse across all so thought will bring it in as is.

Also, thought we can make the service config consistent across all  services if we use a common construct like gate. (may be a trimmed down version)

For now, I will keep the configuration aspects out of this and see whether we can easily put the usage of `Service` object in places where we construct clients.

kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttp3ClientConfiguration.groovy
srekapalli: Reworked on this and made the impl just take the client and work with the builder API internally.

kork-web/src/main/java/com/netflix/spinnaker/config/DefaultServiceEndpoint.java
```diff
@@ -0,0 +1,41 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.config;
+
+/** Represents an service endpoint URL and name. */
```
robzienert, line 20:
```suggestion
/** Represents a service endpoint URL and name. */
```

kork-web/src/main/java/com/netflix/spinnaker/config/ServiceEndpoint.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.config;
+
+public interface ServiceEndpoint {
```
robzienert, line 20: Docs on interface. What is this for? How is it used?

kork-web/src/main/java/com/netflix/spinnaker/config/okhttp3/OkHttpClientBuilderProvider.java
```diff
@@ -0,0 +1,50 @@
...
+   * @return the builder
+   */
+  OkHttpClient.Builder get(ServiceEndpoint service);
+
+  /**
+   * Decides the order in which this provider is consulted.
+   *
+   * @return the assigned priority
+   */
+  Integer priority();
```
robzienert, line 49: Shouldn't this be an `int`? I see the code doesn't handle nullability. Would a default value also be useful?

kork-web/src/main/java/com/netflix/spinnaker/config/okhttp3/OkHttpClientBuilderProvider.java
robzienert: Would like to see docs on all interfaces. Purpose, usage, etc.

kork-web/src/main/java/com/netflix/spinnaker/config/okhttp3/OkHttpClientBuilderProvider.java
cfieber, line 49: We could use the spring @Ordered annotation on instances of these providers in which case the list of providers [here](https://github.com/spinnaker/kork/pull/639/files#diff-89375dc16ec7a991e454e284baaa56d5R31) comes in priority order

kork-web/src/main/java/com/netflix/spinnaker/config/ServiceEndpoint.java
robzienert: Are these methods actually returning nullable types? If not, `@Nonnull`. Same goes for all of other interfaces: Flag non-null method arguments and return types with `@Nonnull` where applicable.

kork-web/src/main/java/com/netflix/spinnaker/config/okhttp3/DefaultOkHttpClientBuilderProvider.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    ConnectionSpec connectionSpec = connectionSpecBuilder.build();
+
+    return builder.connectionSpecs(Arrays.asList(connectionSpec, ConnectionSpec.CLEARTEXT));
+  }
+
+  public OkHttpClient getOkHttpClient() {
+    return okHttpClient;
+  }
```
robzienert, line 155: What's the purpose of this method? Seems leaky to me.

kork-web/src/main/java/com/netflix/spinnaker/config/okhttp3/OkHttpClientProvider.java
```diff
@@ -0,0 +1,64 @@
...
+        providers.stream()
+            .sorted(Comparator.comparingInt(OkHttpClientBuilderProvider::priority).reversed())
+            .filter(provider -> provider.supports(service))
+            .findFirst()
+            .orElse(null);
+
+    if (providerImpl == null) {
+      throw new SystemException(
+          format("No client provider found for url (%s)", service.getBaseUrl()));
+    }
```
robzienert, line 60: nit: This can be shortened to `.orElseThrow(() -> new SystemException("..."));`

kork-web/src/main/java/com/netflix/spinnaker/config/okhttp3/RawOkHttpClientConfiguration.java
```diff
@@ -0,0 +1,40 @@
...
+
+@Configuration
+@EnableConfigurationProperties(OkHttpClientConfigurationProperties.class)
+class RawOkHttpClientConfiguration {
+
+  /**
+   * Default {@link OkHttpClient} that is correctly configured for service-to-service communication.
+   */
+  @Bean
+  OkHttpClient okHttpClient(
```
robzienert, line 35: If this is truly "default" should this factory method include `@ConditionalOnMissingBean(OkHttpClient.class)`?

kork-web/src/main/java/com/netflix/spinnaker/config/okhttp3/DefaultOkHttpClientBuilderProvider.java
srekapalli: Removed. Not needed.

ezimanyi:
@srekapalli : This seems to have broken `clouddriver` at HEAD once the kork bump was merged in...I'm getting:
```
***************************
APPLICATION FAILED TO START
***************************

Description:

Parameter 0 of method httpArtifactCredentials in com.netflix.spinnaker.clouddriver.artifacts.http.HttpArtifactConfiguration required a be
an of type 'com.squareup.okhttp.OkHttpClient' that could not be found.


Action:

Consider defining a bean of type 'com.squareup.okhttp.OkHttpClient' in your configuration.
```

I'm not sure exactly where that bean was coming from before and if there's some corresponding change you'll need to make in clouddriver to account for this...but can you please take a look?  Thanks!

ezimanyi:
Oh no, I think I know what's happening...I think that the new `OkHttpClient okHttpClient()` bean defined here is overriding the existing one [here](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/config/ArtifactConfiguration.java#L34).

While they have the same method name, the new one makes an OkHttp3 client, and the old one makes an OkHttp client; the artifacts code expects to be able to to autowire an `OkHttp3` client but only finds the `OkHttp` client.

If that's the only place that's broken in the services, maybe it just makes sense to update that code to use an OkHttp3 client (or just rename the bean)...but it would not surprise me if this happened elsewhere too.

srekapalli:
@ezimanyi : Sorry, just saw this. Will put in a fix ASAP

PR #642, opened 2020-05-08
feat(plugins): Add preferred plugin info release source

jonsie:
Somewhere along the way we stopped using `SpinnakerPluginInfo` so this change updates a few more files than should've been necessary.  Really, this just adds a new plugin info release source.

### DIFF
--- a/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/release/source/PreferredPluginInfoReleaseSource.kt
+++ b/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/release/source/PreferredPluginInfoReleaseSource.kt
@@ -17,19 +17,11 @@
 
 package com.netflix.spinnaker.kork.plugins.update.release.source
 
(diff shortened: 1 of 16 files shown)


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/release/source/PreferredPluginInfoReleaseSource.kt
```diff
@@ -0,0 +1,44 @@
...
+    return if (release != null) {
+      log.info("Preferred release version '{}' for plugin '{}'", release.version, pluginInfo.id)
+      PluginInfoRelease(pluginInfo.id, release)
+    } else {
+      log.info("No preferred release version found for '{}'", pluginInfo.id)
+      null
+    }
+  }
+
+  override fun getOrder(): Int = 200
```
jonsie, line 43: The order now is: latest plugin info release source -> spring configured plugin info release source -> preferred plugin info release source -> front50 version pinning.

PR #643, opened 2020-05-08
feat(plugins): Requires data class and parser

robzienert:
I've been needing to parse the `requires` field on plugin releases across a few different services, so I wrote this little utility class to consolidate the logic.

### DIFF
--- a/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/PluginReleaseRequires.kt
+++ b/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/PluginReleaseRequires.kt
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


jonsie:
Why not use the existing `VersionRequirementsParser`?

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/PluginReleaseRequires.kt
```diff
@@ -0,0 +1,77 @@
...
+import java.util.regex.Pattern
+
+/**
+ * Represents a Plugin's dependency version requirements.
+ *
+ * Each Plugin Release is able to define a set of version requirements that the Spinnaker environment must satisfy
+ * before the Release will be installed and loaded.
+ *
+ * TODO(rz): Will need to support `&` operator, negation, etc eventually.
+ *   Is there a lib out there that supports this already?
```
jonsie, line 29: Yes, https://github.com/zafarkhaja/jsemver which is what PF4J uses.  Though we'd need to do some preprocessing so we just pass it in the operator and versions, like `>= 1.0.0 & < 2.0.0` since it wont know what to make of the service name obviously.

robzienert:
Oh god. I thought this all felt a little familiar.

PR #645, opened 2020-05-12
fix(misc): Add package level annotations

srekapalli:
-- Adding package level annotations instead of repeating the annotation for every field and method args.

### DIFF
--- a/kork-web/src/main/java/com/netflix/spinnaker/config/DefaultServiceEndpoint.java
+++ b/kork-web/src/main/java/com/netflix/spinnaker/config/DefaultServiceEndpoint.java
@@ -17,6 +17,7 @@
 
 package com.netflix.spinnaker.config;
 
+import java.util.Objects;
 import javax.annotation.Nonnull;
 
 /** Represents a service endpoint URL and name. */
@@ -29,14 +30,18 @@ public class DefaultServiceEndpoint implements ServiceEndpoint {
   @Nonnull private final String baseUrl;
 
   public DefaultServiceEndpoint(@Nonnull String name, @Nonnull String baseUrl) {
-    this.name = name;
-    this.baseUrl = baseUrl;
+    this.name = Objects.requireNonNull(name);
+    this.baseUrl = Objects.requireNonNull(baseUrl);
   }
 
+  @Override
+  @Nonnull
   public String getName() {
     return name;
   }
 
+  @Override
+  @Nonnull
   public String getBaseUrl() {
     return baseUrl;
   }
--- a/kork-web/src/main/java/com/netflix/spinnaker/config/okhttp3/OkHttpClientProvider.java
+++ b/kork-web/src/main/java/com/netflix/spinnaker/config/okhttp3/OkHttpClientProvider.java
@@ -18,14 +18,14 @@ package com.netflix.spinnaker.config.okhttp3;
 
 import static java.lang.String.format;
 
-import com.netflix.spinnaker.config.DefaultServiceEndpoint;
+import com.netflix.spinnaker.config.ServiceEndpoint;
 import com.netflix.spinnaker.kork.exceptions.SystemException;
 import java.util.List;
 import okhttp3.OkHttpClient;
 import org.springframework.stereotype.Component;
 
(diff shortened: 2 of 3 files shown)


kork-web/src/main/java/com/netflix/spinnaker/config/package-info.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+@NonNullFields
+@NonNullApi
+package com.netflix.spinnaker.config;
+
+import org.springframework.lang.NonNullApi;
```
cfieber, line 21: TIL - cool!

cfieber: are these assertions true for everything in this package?

given this is the component scanned package that lives in almost every jar in spinnaker, is there any impact to having this package-info.java in this jar?

kork-web/src/main/java/com/netflix/spinnaker/config/DefaultServiceEndpoint.java
```diff
@@ -17,18 +17,16 @@
...
-  @Nonnull private final String name;
+  private final String name;
 
   /** Base API url */
-  @Nonnull private final String baseUrl;
+  private final String baseUrl;
 
-  public DefaultServiceEndpoint(@Nonnull String name, @Nonnull String baseUrl) {
+  public DefaultServiceEndpoint(String name, String baseUrl) {
     this.name = name;
```
cfieber, line 30: I don't know if we have lombok turned on for this packge, and I don't know if lombok was injecting a null-check on javax.annotation.Nonnull (I know it does on lombok.NotNull), but maybe worth adding an Objects.requireNonNull on these assignments

kork-web/src/main/java/com/netflix/spinnaker/config/package-info.java
maggieneterval: Question: if using spring's `NonNullApi` + `NonNullFields` annotations has the same behavior as using the [`NonNullByDefault`](https://github.com/spinnaker/kork/blob/master/kork-annotations/src/main/java/com/netflix/spinnaker/kork/annotations/NonnullByDefault.java) annotation that @ezimanyi added to kork-annotations, should we standardize on one or the other?

kork-web/src/main/java/com/netflix/spinnaker/config/DefaultServiceEndpoint.java
srekapalli: No lombok on this package. Added requireNonNull checks.

kork-web/src/main/java/com/netflix/spinnaker/config/package-info.java
cfieber: good point. Standardizing seems like a good idea. I suspect having our own flavor of this annotation is worthwhile so that we can use it when building plugin api modules without forcing a spring-core dependency on those APIs

srekapalli: > are these assertions true for everything in this package?
> 
> given this is the component scanned package that lives in almost every jar in spinnaker, is there any impact to having this package-info.java in this jar?

Oops. Verified that IDE applies to all classes with the same package namespace across the entire project.  Removed for `com.netflix.spinnaker.config` package and left it for the `okhttp3` package.

jonsie: Ah good to know.  I also wonder if it would be applied when in services with the kork-web dependency if the package happens to match?

srekapalli: > Ah good to know. I also wonder if it would be applied when in services with the kork-web dependency if the package happens to match?

Yes. It applies for that too. Not sure if that is good or bad ? Although we can relax any of this locally with more specific annotation.

PR #646, opened 2020-05-12
fix(misc): bean rename

srekapalli:


kork-web/src/main/java/com/netflix/spinnaker/config/okhttp3/RawOkHttpClientConfiguration.java
```diff
@@ -34,7 +34,7 @@
    */
   @Bean
   @ConditionalOnMissingBean
-  OkHttpClient okHttpClient(
+  OkHttpClient ok3HttpClient(
```
ajordens, line 37: Maybe `okHttp3Client` to match the `okhttp3` in the package name?

PR #647, opened 2020-05-13
fix(plugins): Use Jackson to convert PF4J PluginInfo to SpinnakerPluginInfo

jonsie:
The cast was a poor choice - it never would succeed.  The reason we didn't see this bug is because we were only testing with front50's update repository, which returns a `SpinnakerPluginInfo` object.  This supports any PF4J compatible update repository (though the `preferred` field is only supported if using the front50 update repository).

### DIFF
--- a/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/internal/SpinnakerPluginInfo.kt
+++ b/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/internal/SpinnakerPluginInfo.kt
@@ -31,5 +31,10 @@ class SpinnakerPluginInfo : PluginInfo() {
     releases = spinnakerReleases
   }
 
-  data class SpinnakerPluginRelease(val preferred: Boolean) : PluginRelease()
+  /**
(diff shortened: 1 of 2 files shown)


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/internal/SpinnakerPluginInfo.kt
```diff
@@ -31,5 +31,5 @@ class SpinnakerPluginInfo : PluginInfo() {
     releases = spinnakerReleases
   }
 
-  data class SpinnakerPluginRelease(val preferred: Boolean) : PluginRelease()
+  data class SpinnakerPluginRelease(val preferred: Boolean = false) : PluginRelease()
```
jonsie, line 34: Any custom fields we add will need to have default values provided.

robzienert: This should be a comment on the class.

jonsie: Done.

PR #653, opened 2020-05-15
refactor(security): refactor AuthenticatedRequest to support generic UserDetails

cfieber:
deprecates User and updates interactions to support UserDetails instead of User.
adds support for encoding allowedAccounts as granted authorities

### DIFF
--- a/kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequest.java
+++ b/kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequest.java
@@ -20,15 +20,13 @@ import static java.lang.String.format;
 
 import com.google.common.base.Preconditions;
 import com.netflix.spinnaker.kork.common.Header;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Optional;
-import java.util.UUID;
+import java.util.*;
 import java.util.concurrent.Callable;
 import lombok.SneakyThrows;
 import org.slf4j.MDC;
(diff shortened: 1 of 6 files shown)


cfieber:
some nastiness in here due to AuthenticatedRequest being all static, but this supports supplying a different opinion on what the current principal is, and how to extract information out of it.

Also deprecates the spinnaker User object that we are currently using for both the purposes of gate authentication as well as switching users.

Adds a runAs method to AuthenticatedRequest so that callers no longer have to construct and supply a User object and deprecates the propagate methods that accept a principal object.

For gate authentication, we can use the spring-security-core User object. In this PR the places where we were expecting our User object are updated to work against the generic UserDetails interface instead, and a mechanism for encoding allowed accounts as granted authorities is provided.

kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequest.java
```diff
@@ -20,19 +20,98 @@
 
 import com.google.common.base.Preconditions;
 import com.netflix.spinnaker.kork.common.Header;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Optional;
-import java.util.UUID;
+import java.util.*;
```
dreynaud, line 23: how did this get past the spotless cop? 👮‍♂️

cfieber: not sure, spotless appears to just not care about `.*`

PR #654, opened 2020-05-18
chore(redis): Remove dynomite from project

robzienert:
No services depend on dynomite anymore. Yum, yum.


marchello2000:
i think these two need to be fixed up:
1. https://github.com/spinnaker/fiat/blob/master/fiat-roles/fiat-roles.gradle#L31
2. https://github.com/spinnaker/swabbie/blob/master/swabbie-redis/swabbie-redis.gradle#L22

PR #657, opened 2020-05-20
fix(plugins): Throw PluginRuntimeException when URL does not work

edgarulg:
**Issue Scenario**
In the plugin metadata, we have a list of releases for a single plugin with the information about it, we moved the plugin to another location, so the URL does not work anymore and we start seeing errors on gate when trying to download plugins.

This is happening because when the `Update Manager` tries to download the plugin and the link is broken (or any other connection error) throws a `PluginRuntimeException` but we do not catch that Exception.

In my changes, I just catch the Exception and return `null` to ignore the plugins that failed on download.


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/SpinnakerUpdateManager.kt
```diff
@@ -150,8 +151,13 @@ class SpinnakerUpdateManager(
...
    */
-  fun downloadPluginRelease(pluginId: String, version: String): Path {
-    return downloadPlugin(pluginId, version)
+  fun downloadPluginRelease(pluginId: String, version: String): Path? {
+    return try {
+      downloadPlugin(pluginId, version)
+    } catch (e: PluginRuntimeException) {
+      log.warn("Unable to download plugin release '{}'", pluginId)
+      null
+    }
```
jonsie, line 160: Hm, I think we would prefer it if you inject `SpringStrictPluginLoaderStatusProvider` into `SpinnakerUpdateManager` and use that to determine whether or not to return null or throw the exception.

robzienert:
I still don't understand how this could possibly be desired behavior. 

If you have a plugin that has been installed and then--for whatever reason--it cannot be installed, you now have a regression in your service. If this were a normal application, you would roll back to the previous version of the software when the code was working.

This behavior seems borderline pathological. I'm assuming that plugins are meant to be offering functional value, but this PR and the others like it, can only lead me to believe these affected plugins do not offer value. That's probably not the case, however, so I'm curious why Armory would want to start their Spinnaker services in a degraded state?

With a strict mode (that defaults enabled), like the other PRs, I'm happy to look the other way but this is just an alarming code smell that it really worries me you're going down the wrong path here.

jonsie:
I think it may make more sense to use `SpringStrictPluginLoaderStatusProvider` in Gate and then determine how to handle the exception here:  https://github.com/spinnaker/gate/blob/master/gate-plugins/src/main/kotlin/com/netflix/spinnaker/gate/plugins/deck/DeckPluginCache.kt#L105

This way the change only affects Gate caching plugins for Deck.  

As it stands right now, making this change here and setting `spinnaker.extensibility.strict-plugin-loading` to `false` in Gate's config would affect both Gate and Deck plugins.

edgarulg:
@jonsie Thanks for your feedback!, make sense for me to move this to Gate and avoid any breaking code on kork, will move this change to Gate with your comments and will open a new PR for that

edgarulg:
@robzienert Thanks for take a look, this is because we have a cache on Gate for plugins that failed on the scenario mentioned, I'm going to move this fix to Gate because it's where the issue is.

jonsie:
Closing, as the change will be made in Gate.

PR #660, opened 2020-05-22
wip(models): Concept of shared models

robzienert:
Proofing out shared models.

The problem that I'm trying to solve for is that we have _basically_ the same model definitions in each service for various models. These models typically drift a little bit from each other, or worse, are just represented as haphazard, recursive `M<S, O>` objects. These models are strewn about our various services, written slightly differently, represented with slightly different (and sometimes incompatible) types that need service-to-service massaging. It's a time vampire.

This PR targets just `PluginInfo` and `PluginRelease`, which are used by kork as well as gate, front50, orca, echo and igor.

This PR has two parts:

1. All models are firstly defined as protos in `kork-proto`. These models would also be sent across the wire as binary (we can just use `jackson-datatype-binary` for this purpose without going full-in on gRPC yet). Any shared model that is going across the wire *MUST* be defined as a proto.
1. A `kork-api` module that depends on `kork-proto`, which wraps defined protos in models that are more ergonomic. This module also defines any and all common validation for these models' fields. Like any other `-api` module, dependencies added to this module should be limited.

Each service is bound to have its own data that it needs to add to a model, which can be done via decoration:

```kotlin
class PluginInfo // from kork-api

class Front50PluginInfo(
  private val korkPluginInfo: PluginInfo, 
  val myAdditionalData: Any
) : PluginInfo by korkPluginInfo
```

Notes:

- I chose Kotlin for this adventure because it's a hell of a lot easier to write and read than Java. Choosing Kotlin would be a total lean-in on Kotlin (since we'd be exposing it as an API). Do we want to risk this dependency? I suspect `kork-api` would find its way into all `service-api` modules at some point or another.
- I'm currently of the mind that each service should define its own protos and models for domains it owns. This PR is in kork because kork owns the plugin codebase: It's not an endorsement for co-locating all models into kork. Is this something we want? Advantage of everything in kork is that dependency bumps are more easily managed, which is not something to disregard...

Future enhancements:

- Documentation generator: Dokka export to Jekyll format and commit into `developer.spinnaker.github.io` repository under a `kork` sub-path. Other services would publish their own `-api` modules here in their own namespaced section. This would be fairly straight forward to do with Github Actions.

**Is this something that we want**

### DIFF
--- a/kork-api/src/main/kotlin/com/netflix/spinnaker/kork/models/validation/PluginIdValidator.kt
+++ b/kork-api/src/main/kotlin/com/netflix/spinnaker/kork/models/validation/PluginIdValidator.kt
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package com.netflix.spinnaker.kork.models.validation
+
+import com.netflix.spinnaker.kork.models.validation.constraints.PluginId
+import java.util.regex.Pattern
+import javax.validation.ConstraintValidator
+import javax.validation.ConstraintValidatorContext
+
+/**
+ * Validates that a [String] value conforms to the canonical [PluginId] format.
+ */
+class PluginIdValidator : ConstraintValidator<PluginId, String> {
+  override fun isValid(value: String, context: ConstraintValidatorContext): Boolean {
+    return pattern.matcher(value).let { matcher ->
+      if (!matcher.matches()) {
+        context
+          .buildConstraintViolationWithTemplate("Plugin IDs must follow a '{namespace}.{id}' format ($pattern)")
+          .addConstraintViolation()
+        false
+      } else {
+        true
+      }
+    }
+  }
+
+  companion object {
+    private val pattern = Pattern.compile("^(?<namespace>[\\w\\-.])+\\.(?<id>[\\w\\-]+)$")
+  }
+}
--- a/kork-api/README.md
+++ b/kork-api/README.md
@@ -1,0 +1,5 @@
+# kork-api
+
+A module for shared models across the Spinnaker services.
+Most, if not all, models in this module should be represented first via protobuf in `kork-proto`:
+This module is mostly for adapting generated proto code into something easier to work with.
--- a/kork-api/kork-api.gradle
+++ b/kork-api/kork-api.gradle
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+apply from: "$rootDir/gradle/kotlin.gradle"
+
+dependencies {
+  implementation project(":kork-annotations")
+  implementation project(":kork-exceptions")
+  implementation project(":kork-proto")
+
+  implementation "com.google.protobuf:protobuf-java"
+
+  implementation "javax.validation:validation-api"
+  implementation "org.hibernate.validator:hibernate-validator"
+  implementation "com.github.zafarkhaja:java-semver:0.9.0" // BOM
+}
--- a/kork-api/src/main/kotlin/com/netflix/spinnaker/kork/models/PluginInfo.kt
+++ b/kork-api/src/main/kotlin/com/netflix/spinnaker/kork/models/PluginInfo.kt
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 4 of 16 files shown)


kork-api/src/main/kotlin/com/netflix/spinnaker/kork/models/validation/PluginIdValidator.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import com.netflix.spinnaker.kork.models.validation.constraints.PluginId
+import java.util.regex.Pattern
+import javax.validation.ConstraintValidator
+import javax.validation.ConstraintValidatorContext
+
+/**
+ * Validates that a [String] value conforms to the canonical [PluginId] format.
+ */
+class PluginIdValidator : ConstraintValidator<PluginId, String> {
```
robzienert, line 27: There's some duplication in this PR: There's already a `CanonicalPluginId` object in `kork-plugins`, but it would be removed in favor of `PluginIdValidator`.

cfieber:
In the past when we had overly specified models where they didn't need to be, we ran into issues where a service would drop parts of a message on the floor because they weren't explicitly defined in its API model (even though that service had no opinion on what the field was and was just going to pass it through to another service).

The nice thing about this approach is that a model change will eventually get propagated to all the services.

The only concern I have is this will potentially introduce some subtle deployment ordering dependencies as these models evolve so as not to lose stuff in transit. 

Any thoughts on how to track/manage that?

robzienert:
> over-specified models

You didn't have any questions here, but this is one of my concerns going forward. I'm zealous about removing `M<S, O>`, but too tightly-defined models _could_ make things difficult going forward. I'd like to lean more heavily on black-boxed model payloads or decoration, rather than forcing every single attribute to be defined if it's not necessary.

> potentially introduce some subtle deployment ordering dependencies as these models evolve

I'm not sure this PR changes anything from our current state of affairs as far as service deployment ordering. For example, for example, adding triggers is a release train across multiple services. I think we can avoid most of the deployment coordination issues through a considered approach WRT black-boxing data across the wire. I think I might need some example scenarios to help make this question live a little more for me?

jonsie:
This proposes defining a proto schema as well as an interface for a model, then creating an implementation which allows us to call `toProto` or `toModel`.  There's a lot I like about this, but I'm also wondering if the additional complexity is worth it right now.  🤔 

If we were to go this route, I would imagine `kork-api` would become a dependency in service `-api` modules.

cfieber:
> I think I might need some example scenarios to help make this question live a little more for me?

Just thinking of the hypothetical scenario of adding a new attribute to one of the models - is that both backwards and forwards compatible? I think we get that from the proto, I think we'll need to be careful on the mappers to make sure they don't when we add new fields.

> I'd like to lean more heavily on black-boxed model payloads or decoration

Yeah I can get behind that, but I wonder how many actually shared models we will end up with if we go that route vs just nicely defined envelopes that don't necessarily care about what is inside them.

ezimanyi:
I like the idea in general!  A few comments:
* As @jonsie said, I wonder whether we want to introduce the complexity of `toProto` and `toModel` at this point.  Do we need that level of indirection (and another place to pass through new fields we add)?
* It looks like the proto definitions are using raw types `string` but the wrapper classes are using the wrapped type `StringValue`.  (And actually the conversion between `String` and `StringValue` is the primary thing these wrapper classes do.).  If we don't need to distinguish between null/empty string (and similar for null/zero for other types) and standardize on the non-wrapped types then the wrapper classes greatly simplify (to the point where maybe we don't need them?)

Around the question of backwards compatibility, proto 3.x as of 3.5 does support retaining and re-serializing [unknown fields](https://developers.google.com/protocol-buffers/docs/proto3#unknowns).  So in principle this means that:
* We add a new field to some proto
* It is passed from Service A -> Service B -> Service C
* Service A can set the field as long as it has an upgraded library; Service C and read the field as
long as it has an upgraded library.  Regardless of whether Service B knows about the field, it will still pass it along.

Of course this probably requires using the proto library to actually do the serialization/deserialization so we'd need to be careful about that.  (And I also don't have direct experience doing this, but that's my understanding from limited knowledge of how this works.)

plumpy:
> I wonder whether we want to introduce the complexity of toProto and toModel at this point. Do we need that level of indirection (and another place to pass through new fields we add)?

In my experience, since the generated protobuf API is relatively nice, it's super common for people to start out just passing around protos directly, but you almost always regret that decision and change it later. You'll want to add some basic features to your model objects beyond what the generated API gives you. Even just simple things like using `Instant` and `Optional` fields instead of `Timestamp` and `@Nullable`. Using a raw proto is okayish but it never feels quite like idiomatic Java. Although maybe Kotlin's extension methods lessen the need for that somewhat? But I'd still be in favor of keeping them, especially since this is basically intended to be a demonstration for future implementations to copy.

(I'm not sure `ProtoModel` is adding anything, though. Is anyone ever going to pass around a `ProtoModel`-typed reference? If not, what value is it adding?)

One option for dodging the missing field question is having your model objects always reference the underlying protobuf instead of defining their own fields. You do lose some flexibility this way, but then you get by on protobuf's unknown field support that Eric referenced. I think only Netflix will really have this issue because of autobumps and the fact that they upgrade all their services at the same time.

robzienert:
@plumpy hit the issue I was trying to tackle with `ProtoModel` well: I don't want proto models to be floating all around the codebase. The generated proto API is inelegant and cumbersome, uses types that we generally consider not best practice in this project, and we typically will want to attach additional logic to the models.

My thought re: `ProtoModel` interface was that we could write some middleware in our RPC client code to convert to and from the raw proto models, but use something richer within the actual application code -- this kind of converter could be written once and applied across any model with the `ProtoModel` interface. I do concede, however, that maintaining two models is not ideal. There is [pbandk](https://github.com/streem/pbandk), but it also doesn't do some of the things I'd like to see (`Instant`, `Duration`, etc). I feel like there's potential for us to generate `ProtoModel` (or something like them) from the protos that fit our needs a little more closely.

I'll try taking a different direction here that doesn't require as much elbow grease and uses the underlying proto directly. I'm not really sure how that will look, but will try.

robzienert:
I'm messing around with writing a protoc plugin to generate `Optional` and `Instant`/`Duration` methods. Guidance on whether or not this is a good idea isn't in great supply, but it seems like it could be a good alternative?

plumpy:
Personally I still think you're going to want things out of model objects besides simple type migrations. Sometimes you want to model things in a proto differently than you would model them in Java, e.g. moving fields from a submessage up into the top-level Java object.

Also I feel like we'd have more flexibility using an annotation processor than trying to write a protoc plugin. Something like

```kotlin
@ProtoWrapper(proto=PluginInfoProto::class)
class PluginInfo : ProtoWrapper_PluginInfoProto
```

where `ProtoWrapper_PluginInfoProto` is generated by `@ProtoWrapper`?

Alternately maybe something like:
```kotlin
class PluginInfo(val proto: PluginInfoProto) : WrappedProto(proto) {
  val id by proto()
  val name by proto()
  val description by proto("metadata.description")
}
```

Or something like that. So you'd still have to add one line for each field you wanted to delegate, but you have complete flexibility to implement them differently if you wanted. Since the object is backed by the actual proto, all unknown fields are maintained.

I dunno, just some more ideas.

robzienert:
OK, I like that a lot more - I do like having a separation between the API contract and internal models. Thanks!

PR #664, opened 2020-05-27
update various dependencies

robfletcher:


spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -40,14 +40,14 @@ dependencies {
   // this project and need to configure gradle plugins etc.
   api(platform("org.jetbrains.kotlin:kotlin-bom:$kotlinVersion"))
   api(platform("org.junit:junit-bom:5.6.2"))
-  api(platform("com.fasterxml.jackson:jackson-bom:2.10.3"))
+  api(platform("com.fasterxml.jackson:jackson-bom:2.10.4"))
```
robfletcher, line 43: This is in-line with the version of Spring in #663

cfieber:
have you tried starting up an app with this change in it?

robzienert:
Please also test OSS and -nflx paths.

robfletcher:
I've backed out the change to Jackson as we ended up backing down a couple of Spring patch versions. That means all the updates here, with the exception of `mysql-connector-java` are test related.

I have tested this against keel and keel-nflx.

In the long term I think there are some things (Jackson and mysql-connector-java among them) that we could and probably should derive from `spring-boot-dependencies.bom`.

PR #667, opened 2020-05-29
refactor(artifacts): Improve null-safety of Artifact class

ezimanyi:
* fix(artifacts): Improve null safety of expected artifact 

  The ExpectedArtifact class is annotated such that id and matchArtifact are not null, but this is not actually enforced anywhere.

  First, explicitly write out the AllArgsConstructor that was generated by lombok, and move the Builder annotation to it (so the builder uses that constructor when finally creating the object).

  Then update the constructor to have a ParametersAreNullable annotation as we can't control what gets passed to the constructor, and handle when we are passed a null id or matchArtifact.

  For the id case, we'll just translate to an empty string. For the matchArtifact case, we'll create an empty matchArtifact. As the point of a matchArtifact is to contain filters that determine whether a given artifact matches this expected artifact; by defaulting to no filters seems like a reasonable behavior here.

* refactor(artifacts): Remove deprecated putMetadata 

  Last time I updated this class (a few months ago) I added the Deprecated annotation to this function, which is not used anywhere in Spinnaker. Removing it now.

* refactor(artifacts): Improve null-safety of Artifact class 

  The Artifact class contains a number of strings that are often null. To make the class more null-safe, annotate it with NonnullByDefault and translate any null values into the empty string in the constructor and in setters.

  In order to do this null -> empty translation, we need to explicitly write out the constructor and the setters that were generated by lombok.

  We rarely mutate artifacts; when we do it often leads to confusing and fragile logic because we often pass these many levels down in the call stack, and also sometimes convert to/from Map<> meaning that it's not always clear if you're operating on the same artifact or a copy. To greatly simplify this, I'd like to make this class immutable. Based on my analysis of where we mutate artifacts, the few places can all be replaced by either (1) fully specifying the artifact up-front or
  (2) returning a copy of the artifact with the desired mutations.

  To start this process of making Artifact immutable, I've annotated all of the setters as Deprecated. Once all of the places where we call these are updated, I'll come back and actually remove the setters.

* refactor(artifacts): Remove redundant JsonProperty 

  These are just explicitly specifying the default values (and in any case we're using the builder to deserialize).

* refactor(artifacts): Clean up metadata creation 

  This is a tiny refactor to simplify creating the metadata in an Artifact builder; rather than check if the Map is null every time we add a key, just initialize it.

* refactor(artifacts): Make Artifact final

  If we want the class to truly be immutable in the future, it will need to be final. We don't currently extend this anywhere, so let's make it final.

### DIFF
--- a/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/Artifact.java
+++ b/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/Artifact.java
@@ -21,16 +21,21 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import com.google.common.base.Strings;
+import com.netflix.spinnaker.kork.annotations.NonnullByDefault;
 import java.util.HashMap;
 import java.util.Map;
-import lombok.AccessLevel;
-import lombok.AllArgsConstructor;
+import java.util.Optional;
+import javax.annotation.ParametersAreNullableByDefault;
 import lombok.Builder;
-import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.ToString;
 
-@AllArgsConstructor(access = AccessLevel.PRIVATE)
-@Data
-@Builder(toBuilder = true)
+@Getter
+@ToString
+@EqualsAndHashCode
+@NonnullByDefault
 @JsonDeserialize(builder = Artifact.ArtifactBuilder.class)
 public class Artifact {
   @JsonProperty("type")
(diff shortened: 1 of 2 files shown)


kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/Artifact.java
```diff
@@ -63,6 +68,99 @@
...
+    this.reference = Strings.nullToEmpty(reference);
+    this.metadata = Optional.ofNullable(metadata).orElseGet(HashMap::new);
+    this.artifactAccount = Strings.nullToEmpty(artifactAccount);
+    this.provenance = Strings.nullToEmpty(provenance);
+    this.uuid = Strings.nullToEmpty(uuid);
+  }
+
+  // All setters for this class are deprecated. In general, artifacts are passed around the pipeline
+  // context and are liberally serialized and deserialized, which makes mutating them fraught with
+  // peril (because you might or might not actually be operating on a copy of the artifact you
```
maggieneterval, line 78: "No one would do such things. Civilization has climbed above such perils."

-- Winston Churchill

maggieneterval:
(👨 🔨 📆 ➡️ 📆 🐛 means you prevented future bugs)

PR #670, opened 2020-06-01
fix(artifacts): Fix non-null annotation

ezimanyi:
There's still one annoying thing that I didn't properly fix in the last commit. Because the whole Artifact class is annotated as NullableByDefault, we get a warning every time we call .toBuilder() which *is* guaranteed to return non-null. We can't directly annotate it as it's generated by Lombok; let's fix this by using FieldsAreNullableByDefault so that it doesn't affect method values.

At first I wanted to affect method return values because the fact that the fields are nullable means the accessors are also nullable. But after testing, marking the fields as Nullable does cause the lombok-generated accessors to be marked nullable, so marking the fields nullable is enough.

### DIFF
--- a/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/Artifact.java
+++ b/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/Artifact.java
@@ -26,6 +26,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import lombok.Builder;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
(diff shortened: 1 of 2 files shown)


ezimanyi:
Added one more commit!
* refactor(artifacts): Add function getMetadata(string)

  It's way safer to roll out making metadata shallowly immutable if we remove the ability for consumers to get the raw map at all.

  Currently consumer can call getMetadata and mutate the map; we can make the map immutable but there's no safe way to do that with warning to consumers. (They will go from getting a mutable map to an immutable map and get a runtime error on mutating.)

  A safer plan is to deprecate getMetadata in favor of having clients ask for the metadata key of interest; once the function getMetadata is removed then we will be able to transparently make the metadata map immutable without affecting clients.

  This of course is all just about making getMetadata shallowly immutable. It unfortunately looks like there are some consumers that are storing entire objects in metadata (in many cases just using an artifact's metadata as a convenient place to store untyped data). If I could start over, I would have make this a Map<String, String> but it's too late for that, and we'll likely need to just live with the fact that metadata is only shallowly immutable (and, what I think is worse, can contain arbitrary untyped objects).

PR #671, opened 2020-06-02
refactor(*): Remove Hystrix

robzienert:
Depends on https://github.com/spinnaker/gate/pull/1233


ajordens:
Yes to @ezimanyi's comment ... we do have pockets that will require some rework.

`front50` might be serving fallbacks out of the cache, suspect that's still possible in a non-hystrix world (or may actually warrant resilience4j).

robzienert:
I will migrate existing, valid use cases to resilience4j.

robzienert:
Byeeee.

PR #678, opened 2020-06-11
Avoid trying to truncate DB views

robfletcher:
Turns out `DatabaseMetadata.getTables()` will include views as well as tables. If you then try to truncate those, it esplode.


robfletcher:
Turns out `DatabaseMetadata.getTables()` will include views as well as tables. If you then try to truncate those, it esplode.

PR #681, opened 2020-06-15
refactor(artifacts): Shallow copy artifact metadata

ezimanyi:
As part of making Artifact immutable, we also want to make the contents of metadata shallowly immutable.  (We can't enforce deep immutability as it can contain arbitrary Objects.)

We've already deprecated (and removed from calling code) any way for consumers to directly modify the metadata map (or even for them to know that we are storing this as a Map).

That being said, there is still the loophole of clients passing in a Map, then keeping around a reference and modifying it. Let's close that loophole my making a shallow copy of the map before storing it on the Artifact.

Also, remove an unecessary qualified reference to Artifact.ArtifactBuilder.


ezimanyi:
I've had this on a branch for the last couple of weeks but wanted to let my original artifact changes settle first.  This should be much more minor, and I'd like to get it in before the 1.21 release (so I can fully remove all the deprecated stuff before 1.22).

PR #683, opened 2020-06-15
fix(plugins): Lookup method from target class when instrumenting extension point method invocation with timing metric

jonsie:
Typically the concrete target class will not be invoked directly, instead the invocation will occur on the type (like `Task#execute` for example) -- in that instance we would fail to lookup the `Metered` annotation because the interface/abstract class methods do not and should not contain the `Metered` annotation.


kork-telemetry/src/main/java/com/netflix/spinnaker/kork/telemetry/MethodInstrumentation.java
```diff
@@ -35,7 +35,7 @@ public static boolean isMethodAllowed(Method method) {
     return
     // Only instrument public methods
     Modifier.isPublic(method.getModifiers())
-        ||
+        &&
```
jonsie, line 38: This should've always been && and not ||.  This will now return false if either of these cases are false - which is the expected behavior.

PR #693, opened 2020-06-20
chore(misc): add Dependabot

sullis:
https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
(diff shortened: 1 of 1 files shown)


cfieber:
not sure how I feel about autobumping our java dependencies in this project

some of them can be pretty disruptive when they are cascaded to the other services via kork autobump, and are not likely to manifest as problems in the kork build itself (something like bumping a minor version of jackson or spring-boot comes to mind, both of which I believe would happen if we updated to latest on those)

I'd rather try this out in one of the service repos first

PR #694, opened 2020-06-23
feat(plugins): Add plugin API Environment SDK; get environment name and region from Spring environment

jonsie:
This supports the potential use case wherein a plugin developer wants to load specific configuration for a Spinnaker environment.


jonsie:
Closing since I think I'm going to make this more generic and load up a map of properties - that way organizations can default to existing properties they may already have.

PR #701, opened 2020-07-01
refactor(artifacts): Make fields final

ezimanyi:
* refactor(artifacts): Remove deprecated functions

  These functions have been deprecated for more than a month (including a release of Spinnaker) and have been removed from all client code. It's time to remove them entirely.

* refactor(artifacts): Make fields final

  Now that we've removed all setters for Artifact, we can make its fields final.


ezimanyi:
I'm going to wait until Monday to merge this and bump `kork` in the microservices so I'm around if there are any issues.  This should be pretty safe, but the one risk is that there's a call to one of these deprecated functions in `groovy` that I missed (and that would be a runtime error rather than a compile-time error).

PR #702, opened 2020-07-01
test(artifacts): Add some tests around null metadata values

ezimanyi:
I had wanted to make the metadata map an ImmutableMap as it is now immutable, but this caused some NPEs because it is possible to have metadata with null values. In order to exercise the current functionality, add some tests around what happens when you have null values in the metadata map.

At this point this PR will just add the tests so nobody makes the same mistake; I may come back later to filter out null values, then make the map immutable, but for now just adding the tests.

### DIFF
--- a/kork-artifacts/src/test/java/com/netflix/spinnaker/kork/artifacts/model/ArtifactTest.java
+++ b/kork-artifacts/src/test/java/com/netflix/spinnaker/kork/artifacts/model/ArtifactTest.java
@@ -231,6 +231,59 @@ final class ArtifactTest {
     assertThat(artifact.getMetadata("key2")).isNull();
   }
 
+  @Test
+  void putMetadataNullValue() throws IOException {
+    Artifact artifact = Artifact.builder().putMetadata("test", null).build();
+    assertThat(artifact.getMetadata("test")).isNull();
+  }
+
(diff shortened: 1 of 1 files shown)


ezimanyi:
The issue is that unlike a regular `Map`, an `ImmutableMap` classes don't allow `null` keys or values.  So you'll get an NPE from `ImmutableMap.copyOf(map)` if `map` has any keys or values that are `null`.

It turns out that in some places we call `ArtifactBuilder.putMetadata` with something that can be null (such as [here](https://github.com/spinnaker/clouddriver/blob/eada6786aadd726448f4291e993cde282c171381/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/manifest/KubernetesManifestAnnotater.java#L197)) so we'd have to fix all those places before making this an `ImmutableMap`.

I actually think that a better fix would be to just filter out null values from the map before converting to an `ImmutableMap` because it won't change the result of `map.get("key")` which will be `null` in either case and is the only externally-visible way to query the map. (Since we don't expose `.containsKey()` or similar.)  At some point in the future I may make that change, but for now these tests will prevent me (or someone else) from just making this an `ImmutableMap` without considering the `null` case.

PR #703, opened 2020-07-02
feat(plugins): Support range constraint

edgarulg:
This PR is about to support range constraint on a service that a plugin requires on.

With this change, they would be two valid formats for the required field:

`orca>=1.0.0`  <-- currently supported
`orca>=1.0.0 & <2.0.0`  <-- new constraint supported
 
When a range constraint is found It creates a list with two `VersionRequirements` that will be validated with `SpinnakerServiceVersionManager` as it works currently.


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/VersionRequirementsParser.kt
```diff
@@ -27,41 +27,54 @@ import java.util.regex.Pattern
  * - `operator` is a version constraint operator (`>`, `<`, `>=`, `<=`)
  * - `version` is the service version that is being constrained
  *
- * TODO(jonsie): Add range constraint support (>= 1.0.0 & < 2.0.0)
  */
 object VersionRequirementsParser {
 
   private val SUPPORTS_PATTERN = Pattern.compile(
-    "^(?<service>[\\w\\-]+)(?<operator>[><=]{1,2})(?<version>[0-9]+\\.[0-9]+\\.[0-9]+)$")
+    "^(?<service>[\\w\\-]+)(?<operator>[><=]{1,2})(?<version>[0-9]+\\.[0-9]+\\.[0-9]+)( & (?<rOperator>[><=]{1,2})(?<rVersion>[0-9]+\\.[0-9]+\\.[0-9]+))?$")
```
robzienert, line 34: Oh jeez. This is what I get for using regex to begin with. 🤣 

We should switch this class to wrap [jsemver](https://github.com/zafarkhaja/jsemver) instead of doubling down on our regex pattern.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/VersionRequirementsParser.kt
```diff
@@ -27,41 +27,54 @@ import java.util.regex.Pattern
...
    * Parse a list of comma-delimited versions.
    */
-  fun parseAll(version: String): List<VersionRequirements> =
-    version.split(',').map { parse(it.trim()) }
+  fun parseAll(version: String): List<VersionRequirements> {
+    val result = mutableListOf<VersionRequirements>()
+    version.split(',').forEach {
+      result.addAll(parse(it.trim()))
+    }
+    return result
```
robzienert, line 76: Why?

robzienert:
Great improvement, but I think we should clean things up while we're here instead.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/VersionRequirementsParser.kt
edgarulg: I agree, I don't like to have the regex growing when we want to support more constraints, I changed it to have two groups one for the `service` and another for the `constraint`, then I uses `jsemver` to validate if the constraint is a valid.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/VersionRequirementsParser.kt
edgarulg: I did this because having a range constraint I could return multiple `VersionRequirements` to be validated by the `SpinnakerServiceVersionManager` but using [jsemver](https://github.com/zafarkhaja/jsemver) it isn't necessary to keep doing that

edgarulg:
@robzienert Thanks for your comments! I reviewed and changed my code to use `jsemVer` library, I keep the regex to get the service name but I am using `jsemver` to validate the input constraint. Looking forward for your comments.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/VersionRequirementsParser.kt
```diff
@@ -47,12 +47,12 @@ object VersionRequirementsParser {
         if (!it.matches()) {
           throw InvalidPluginVersionRequirementException(version)
         }
+        try { CONSTRAINT_VALIDATOR.satisfies(it.group(SUPPORTS_PATTERN_CONSTRAINT_GROUP)) } catch (e: ParseException) { throw InvalidPluginVersionRequirementException(version) }
```
robzienert, line 50: nit: Let's break this up into multiple lines.

edgarulg: its done!

edgarulg:
> One little nitpick, but this is greatly improved, thanks!

Thanks for your comments!

PR #710, opened 2020-07-08
 refactor(eureka): Move Eureka out of kork-core into kork-eureka

robzienert:
- Moves Eureka code and dependencies from `kork-core` into `kork-eureka`; importing into `kork-runtime`.
- Decouples Eureka from `RemoteStatusChangedEvent`; opening the door for other service discovery techs to be used with Spinnaker.
- Refactors existing `kork-eureka` code to use new discovery abstraction.

This PR will require modifications to downstream services.


kork-core/kork-core.gradle
```diff
@@ -11,15 +11,13 @@ dependencies {
...
-  api "com.netflix.eureka:eureka-client"
   api "com.netflix.spectator:spectator-api"
   api "com.google.code.findbugs:jsr305"
   api "io.github.resilience4j:resilience4j-annotations"
   api "io.github.resilience4j:resilience4j-retry"
   api "io.github.resilience4j:resilience4j-spring-boot2"
   api "javax.inject:javax.inject:1"
 
-  implementation "com.netflix.netflix-commons:netflix-eventbus"
   implementation "com.netflix.archaius:archaius-core"
```
cfieber, line 21: I think the only reason that we have archaius config in here is to bridge spring configuration to the legacy eureka client - I think this should move along into the kork-eureka module

robzienert: Well, there's other Archaius stuff in the module - I was going to deal with all that in a separate PR, but I'll investigate doing it in one shot.

kork-eureka/src/main/java/com/netflix/spinnaker/kork/eureka/EurekaComponents.java
```diff
@@ -16,6 +16,8 @@
 
 package com.netflix.spinnaker.kork.eureka;
```
cfieber, line 17: I think there should be a spring.factories in here referencing this config, and I also think that PlatformComponents in kork-core needs to remove this from its Import but I don't see that change in here

kork-eureka/src/main/java/com/netflix/spinnaker/kork/eureka/EurekaComponents.java
```diff
@@ -16,6 +16,8 @@
...
 
+import static com.netflix.spinnaker.kork.eureka.InstanceStatusBridge.fromEureka;
+
 import com.netflix.appinfo.*;
```
cfieber, line 21: I think it might be worth adding another autoconfiguration class here conditional on eureka.enabled=false that publishes a Status UP event (I know something similar exists in orca and others that try to control polling activity based on in-service status for the case where there is no discovery service):
https://github.com/spinnaker/orca/blob/master/orca-eureka/src/main/java/com/netflix/spinnaker/orca/eureka/NoDiscoveryApplicationStatusPublisher.java

cfieber: alternatively there could be an autoconfiguration in kork-core that supplies that ConditionalOnMissingBean, and this configuration could be flagged to AutoConfigureBefore that configuration so the conditional stuff actually works

kork-eureka/src/main/java/com/netflix/spinnaker/kork/eureka/EurekaComponents.java
robzienert: Ah, yarp. I did do that, but didn't push it.

robzienert:
Much improved, very done.

kork-eureka/src/test/kotlin/com/netflix/spinnaker/kork/EurekaSystemTest.kt
```diff
@@ -0,0 +1,94 @@
...
+import org.springframework.boot.test.context.assertj.AssertableApplicationContext
+import org.springframework.boot.test.context.runner.ApplicationContextRunner
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import org.springframework.context.annotation.Import
+import strikt.api.expect
+import strikt.assertions.isA
+import strikt.assertions.isEqualTo
+
+class EurekaSystemTest : JUnit5Minutests {
```
cfieber, line 38: nice addition, thanks!

kork-security/kork-security.gradle
```diff
@@ -5,14 +5,17 @@ apply from: "$rootDir/gradle/lombok.gradle"
 dependencies {
   api(platform(project(":spinnaker-dependencies")))
 
-  api project(":kork-core")
+  implementation(project(":kork-core")) { transitive = false }
```
cfieber, line 8: does this still need to be `transitive=false`? (I had only flipped it that way to try and figure out what dependencies might have been confusing the groovy compiler)

robzienert: Ah, no it appears not. Nice.

PR #712, opened 2020-07-09
feat(sql): Add a feature flag for liquibase migrations

jervi:
Right now it is hard to run e.g. a read-only clouddriver cluster towards a read-only database, because there's no way to disable the migrations (which will cause Clouddriver to fail on startup if it can't write to the DB).
I'm piggybacking on the `sql.read-only` feature flag.


kork-sql/src/main/kotlin/com/netflix/spinnaker/kork/sql/config/DefaultSqlConfiguration.kt
```diff
@@ -64,11 +65,13 @@ class DefaultSqlConfiguration {
 
   @Bean
   @ConditionalOnMissingBean(SpringLiquibase::class)
+  @ConditionalOnExpression("\${sql.read-only:false} == false")
```
cfieber, line 68: I tend to prefer `@ConditionalOnProperty(value = "sql.read-only", havingValue = "false", matchIfMissing = true)` since it more clearly shows which properties are being used and hopefully one day we can get some tooling to help generate some property documentation...

jervi: I actually wrote that first, but I changed it to a `ConditionalOnExpression` because I can't nest multiple `ConditionalOnProperty` annotations without creating a new `AllNestedConditions` meta-annotation (needed for the `secondaryLiquibase` bean below)

jervi: WDYT? Should I create a new annotation just for this purpose?

cfieber: nah, this seems fine in that case, thanks for the explaination

PR #716, opened 2020-07-10
feat(sql): Add PostgreSQL test container utils

jcavanagh:
This adds new SQL testcontainers functionality for Postgres.  

The no-args versions of `initDatabase()` and `initPreviousDatabase()` as marked as deprecated, since they don't seem to be in use.


kork-sql-test/src/main/java/com/netflix/spinnaker/kork/sql/test/SqlTestUtil.java
```diff
@@ -44,16 +44,22 @@
...
 import org.testcontainers.containers.MySQLContainer;
+import org.testcontainers.containers.PostgreSQLContainer;
 
 public class SqlTestUtil {
 
   public static String tcJdbcUrl = "jdbc:tc:mysql:5.7.22://somehostname:someport/somedb";
 
+  public static String tcPgJdbcUrl = "jdbc:tc:postgres:10.13:///test";
+
+  @Deprecated
```
robzienert, line 57: When adding `@Deprecated` annotations, we require javadoc about the deprecation as well, since the annotation on its own isn't very useful.

jcavanagh: No problem - I've added a short comment to use the TestContainers implementations instead.  Please let me know if you'd prefer more or different detail.

PR #718, opened 2020-07-10
Revert "feat(sql): Add a feature flag for liquibase migrations"

cfieber:
Reverts spinnaker/kork#712

I'm seeing an issue with this change and read-only mode where things fail due to some unsatisfied dependencies

there is an `@DependsOn("liquibase")` on the data source class and it seems like with read-only set that is either blocking a whole chain of components from initializing or introducing a circular dependency (because there is another bean named liquibase that this ends up dependant on)

the purpose of the dependsOn annotation was to block the context from continuing to initialize while the migrations are running, so if we can find another way to do that we should be good to get this change back in

alternatively could we just set spring.liquibase.enabled=false .. that stops liquibase autoconfiguration (that probably blows something else up... heh)

### DIFF
--- a/kork-sql/src/main/kotlin/com/netflix/spinnaker/kork/sql/config/DefaultSqlConfiguration.kt
+++ b/kork-sql/src/main/kotlin/com/netflix/spinnaker/kork/sql/config/DefaultSqlConfiguration.kt
@@ -36,7 +36,6 @@ import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.BeanCreationException
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration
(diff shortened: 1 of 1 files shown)


cfieber:
I think the same ordering constraint can be achieved by removing the dependsOn annotation and adding `fun dataSource(liquibase: Optional<SpringLiquibaseProxy>, ...)` but I think there is still an issue with needing the default liquibase autoconfiguration to not happen as I'm still seeing startup failures

jervi:
I tried the `spring.liquibase.enabled=false`, but it didn’t seem to have any effect, the migrations were still running. I thought it was because liquidate was running outside of the spring boot built-in migrations or something, but I didn’t spend much time on it.

jervi:
I can create a new PR that doesn’t remove the Liquibase beans from the context, but instead sets the `shouldRun` flag to `false` if read only is `true`.

PR #719, opened 2020-07-10
feat(sql): Add a feature flag for liquibase migrations

jervi:
Right now it is hard to run e.g. a read-only clouddriver cluster towards a read-only database, because there's no way to disable the migrations (which will cause Clouddriver to fail on startup if it can't write to the DB).
I'm piggybacking on the `sql.read-only` feature flag.

This is take 2 of this PR, after #712 had to be reverted due to Spring startup issues. See #718. I think this way of disabling the migration is less intrusive, because the liquibase beans will still be present in the context.


jervi:
PTAL @cfieber

robzienert:
We have readonly clusters for clouddriver without any problems: Our `migration` JDBC url doesn't use the readonly endpoint.

PR #720, opened 2020-07-13
fix(spel): Fix flattening of nested arrays in SpEL evaluation

ezimanyi:
Fixes spinnaker/spinnaker#5865.

* test(spel): Add tests on ExpressionTransform

  The next commit will fix a bug in ExpressionTransform; first add some tests to document the current behavior. In particular, test the behavior in handling nested maps and nested lists.

* fix(spel): Fix flattening of nested arrays in SpEL evaluation

  Currently nested lists are being flattened when evaluating expressions. This is incorrect, as expression evaluation should not affect the shape of the data structure; it should only handle evaluating expressions in the components of that data structure.

  There are some situations in the Kubernetes provider where we need a list of lists in the context; this currently breaks because upon evaluating expressions these lists get flattened.

### DIFF
--- a/kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/ExpressionTransform.java
+++ b/kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/ExpressionTransform.java
@@ -132,7 +132,7 @@ public class ExpressionTransform {
       if (obj instanceof Map) {
         result.add(transformMap((Map<String, Object>) obj, evaluationContext, summary));
       } else if (obj instanceof List) {
-        result.addAll(transformList((List) obj, evaluationContext, summary, additionalContext));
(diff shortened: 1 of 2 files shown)


apurvchandra:
Hi @ezimanyi can you give an idea when is this going to be released.

maggieneterval:
@apurvchandra this fix will be released with Spinnaker 1.22, which is currently scheduled for mid-August. You can subscribe to the [Spinnaker release calendar](https://spinnaker.io/community/releases/release-cadence/#upcoming-releases) for an exact date.

If you'd like to test this change before the 1.22 release, you can run the nightly builds using [these instructions](https://spinnaker.io/community/contributing/releasing/#run-the-nightly-builds-not-recommended), or if you need further customization, you can deploy a custom BOM using [these instructions](https://spinnaker.io/guides/operator/custom-boms/#deploy-custom-spinnaker-builds).

PR #721, opened 2020-07-13
fix(autoconfiguration): fix discovery/eureka configuration ordering

cfieber:
Moves some `Configuration` classess that were previously being `Import`ed into an
`AutoConfiguration` class into their own dedicated `AutoConfiguration` referenced
in `spring.factories`. This enables applying ordering annotations to configure
relationships between these `AutoConfiguration`s.

This enables reliable use of `ConditionalOnBean` and `ConditionalOnMissingBean`
as well as solves some implicit configuration ordering issues between Archaius
and Eureka more cleanly than previously.


cfieber:
Also removes `ClasspathPropertySource` which was a bad version of using a `PropertySource` annotation on a `Configuration` class.

the `kork-eureka` module is not intended to have a public compile time API so there is no reason to keep this around.

PR #723, opened 2020-07-13
feat(plugins): added the SpringLoaderPlugin

claymccoy:
It allows the use of package scanning and Spring Configurations.
The ExposeToApp annotation promotes beans to the application context.
This is a plugin written using SpringLoaderPlugin: https://github.com/spinnaker-plugin-examples/springExamplePlugin/pull/4

### DIFF
--- a/kork-plugins-spring-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/spring/SpringLoaderPlugin.java
+++ b/kork-plugins-spring-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/spring/SpringLoaderPlugin.java
@@ -17,6 +17,8 @@
 package com.netflix.spinnaker.kork.plugins.api.spring;
 
 import com.netflix.spinnaker.kork.annotations.Alpha;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import org.pf4j.PluginWrapper;
@@ -45,6 +47,8 @@ public abstract class SpringLoaderPlugin extends PrivilegedSpringPlugin {
 
   @Override
   public void registerBeanDefinitions(BeanDefinitionRegistry registry) {
+    final String springLoaderBeanName = wrapper.getPluginId() + "." + SpringLoader.class.getName();
+
     ClassLoader pluginClassLoader = getClass().getClassLoader();
     BeanDefinition springLoaderBeanDefinition =
         BeanDefinitionBuilder.genericBeanDefinition(SpringLoader.class)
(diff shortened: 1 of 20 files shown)


kork-plugins-spring-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/spring/SpringLoaderPlugin.java
```diff
@@ -0,0 +1,79 @@
...
+    try {
+      registerBean(springLoaderBeanDefinition, registry);
+    } catch (ClassNotFoundException e) {
+      throw new IllegalStateException(e);
+    }
+
+    // delay controller mapping until after the plugin has a chance to load its own controllers
+    registry
+        .getBeanDefinition("requestMappingHandlerMapping")
+        .setDependsOn("com.netflix.spinnaker.kork.plugins.api.spring.SpringLoader");
```
ncknt, line 67: Does this actually work with multiple plugins? Line 59, we register the `SpringLoader` bean but the next plugin will also try to do the same. Even in the event that it works, I imagine it wouldn't wait for the package scan of all loaders. Should we instead just register the bean with a unique name (e.g. `this.getClass().getName()`)

claymccoy: Good point. I'll test this. I should probably differientiat based on the plugin.

kork-plugins-spring-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/spring/SpringLoaderPlugin.java
```diff
@@ -0,0 +1,79 @@
...
+import org.pf4j.PluginWrapper;
+import org.springframework.beans.factory.config.AutowireCapableBeanFactory;
+import org.springframework.beans.factory.config.BeanDefinition;
+import org.springframework.beans.factory.support.BeanDefinitionBuilder;
+import org.springframework.beans.factory.support.BeanDefinitionRegistry;
+import org.springframework.context.annotation.AnnotationConfigApplicationContext;
+
+/** Allows a plugin to scan packages for beans and load Spring Configurations. */
+@Alpha
+public abstract class SpringLoaderPlugin extends PrivilegedSpringPlugin {
```
jonsie, line 31: Do you see a use for a standalone `PrivilegedSpringPlugin` or should this be consolidated into one class?

claymccoy: Yes, there are a few reasons that I did this.
* Backwards compatibility. Several of our early plugins use this and there are no changes to it.
* This is the core functionality that registers bean definitions. SpringLoader just decorates that with some niceties. There may be different and better ways to do this. I didn't want to conflate the core functionality with the icing that may not ultimately be ideal.

kork-plugins-spring-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/spring/SpringLoaderPlugin.java
claymccoy: @ncknt I added a new commit that fixes this and tested with mulitple plugins.

PR #724, opened 2020-07-14
Service client implementation

srekapalli:
-- Allows us to remove boiler plate code in every service which creates a retrofit client.
From :
```
new RestAdapter.Builder()
        .setEndpoint(orcaBaseUrl)
        .setClient(new Ok3Client(okHttpClientProvider.getClient(new DefaultServiceEndpoint("orca", orcaUrl))))
        .setConverter(new JacksonConverter(mapper))
        .setLog(new Slf4jRetrofitLogger(OrcaService.class))
        .setLogLevel(logLevel)
        .build()
        .create(OrcaService.class);
```

to:
```
retrofitServiceProvider.getService(OrcaService.class, new DefaultServiceEndpoint('orca','http://orca.net')`
```

-- Possibility of switching to other client/http engines if needed in the future.
-- Once every service moves to this impl, we can remove all the `retrofit` related dependencies from kork-web.


kork-serviceclient-retrofit/kork-serviceclient-retrofit.gradle
```diff
@@ -0,0 +1,18 @@
+apply plugin: "java-library"
+
+dependencies {
+  api(platform(project(":spinnaker-dependencies")))
+
+  implementation project(":kork-web")
+
+  api "com.squareup.retrofit:retrofit"
```
cfieber, line 8: these should all be implementation or runtimeOnly if this module doesn't expose a public interface

kork-web/src/main/java/com/netflix/spinnaker/kork/web/serviceclient/ServiceClientFactory.java
```diff
@@ -0,0 +1,50 @@
...
+   * Initializes and returns the underlying client builder implementation for further customizations
+   * as needed.
+   *
+   * @param type client type
+   * @param serviceEndpoint endpoint configuration
+   * @param <T> type of client , usually represents a interface with all the remote method
+   *     definitions.
+   * @return a client builder
+   */
+  public <T> B build(Class<T> type, ServiceEndpoint serviceEndpoint);
```
cfieber, line 49: It will be interesting to see how often this method is needed in client code.

I wonder if we could define something like ServiceClientCustomizer<B> here, and let customizations happen via beans that implement that as needed so this doesn't have to happen at the service client creation call site.

Just a thought though

srekapalli: Good point. Probably worth having specialized clients also built the same way as default one by implementing the same interface.  Will put in a different approach.

kork-web/src/main/java/com/netflix/spinnaker/kork/web/serviceclient/ServiceClientFactory.java
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * Decide if this factory can support the endpoint provided.
+   *
+   * @param type client type
+   * @param serviceEndpoint endpoint configuration
+   * @param <T> type of client , usually represents a interface with all the remote method
+   *     definitions.
+   * @return a client builder
+   */
+  public default <T> boolean supports(Class<T> type, ServiceEndpoint serviceEndpoint) {
```
cfieber, line 44:
```suggestion
  public default boolean supports(Class<?> type, ServiceEndpoint serviceEndpoint) {
```

I think the type isn't going to be known for implementors of this method (if it is, `<T>` could be moved to the class level)

cfieber:
are there dependencies declared as `api` in kork-web that can be removed with this change?

jonsie:
I think some of the terminology here is confusing me.  Perhaps we should call this `RetrofitServiceProvider` as opposed to `ClientServiceProvider`?  I think the use of `client` is odd since we already build up the underlying client (OkHttp) and typically the Retrofit bit is referred to as a `service`.  Also, the module name `kork-serviceclient-retrofit` seems strange - perhaps just `kork-retrofit`?

jonsie:
I think we'll also want to support Retrofit 2 with this module.

jonsie:
In general I would say replace instances of `ClientService` with `RetrofitService` and that will help clear up the purpose of the code a bit more.

kork-serviceclient-retrofit/src/main/java/com/netflix/spinnaker/kork/web/serviceclient/DefaultServiceClientFactory.java
```diff
@@ -0,0 +1,66 @@
...
+      RestAdapter.LogLevel retrofitLogLevel,
+      OkHttpClientProvider clientProvider,
+      RequestInterceptor spinnakerRequestInterceptor) {
+    this.retrofitLogLevel = retrofitLogLevel;
+    this.clientProvider = clientProvider;
+    this.spinnakerRequestInterceptor = spinnakerRequestInterceptor;
+  }
+
+  @Override
+  public <T> T getClient(Class<T> type, ServiceEndpoint serviceEndpoint) {
```
jonsie, line 50: I would also change this to `getService`.

kork-web/src/main/java/com/netflix/spinnaker/kork/web/serviceclient/ServiceClientFactory.java
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.kork.web.serviceclient;
+
+import com.netflix.spinnaker.config.ServiceEndpoint;
+
+/** Factory to build a client for a service. */
+public interface ServiceClientFactory {
```
jonsie, line 23: Probably change this to `RetrofitServiceFactory` to.

srekapalli: I will leave this as is (ServiceClientFactory), since it doesn't carry Retrofit knowledge.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/web/serviceclient/DefaultRetrofitServiceFactory.java
```diff
@@ -0,0 +1,67 @@
...
+import com.netflix.spinnaker.config.okhttp3.OkHttpClientProvider;
+import com.netflix.spinnaker.kork.annotations.NonnullByDefault;
+import com.netflix.spinnaker.retrofit.Slf4jRetrofitLogger;
+import retrofit.Endpoint;
+import retrofit.RequestInterceptor;
+import retrofit.RestAdapter;
+import retrofit.converter.JacksonConverter;
+
+@NonnullByDefault
+class DefaultRetrofitServiceFactory implements ServiceClientFactory {
```
jonsie, line 35: Maybe this can just be called the `RetrofitServiceFactory`?

jonsie:
Looks like `serviceclient` is still a package name here.  Maybe that should be changed (or relevant code moved to) the `retrofit` package?

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/RetrofitServiceFactoryAutoConfiguration.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.Ordered;
+import org.springframework.core.annotation.Order;
+import retrofit.RequestInterceptor;
+import retrofit.RestAdapter;
+
+@Configuration
+@ConditionalOnProperty(value = "retrofit.enabled", matchIfMissing = true)
```
jonsie, line 32: Under what context would we disable this?

srekapalli: Probably never if we have this as a dependency in the application , but during test phase sometimes It's useful to turn off things if needed and wire up the bare minimum instead of making custom configurations.

jonsie: Sounds good.

PR #726, opened 2020-07-14
feat(plugins): Allow metrics on private methods annotated with @Metered

jonsie:
I was reviewing a plugin that @dreynaud wrote and saw that he annotated a private method on an extension point with `@Metered`.  I then realized that our logic to skip private methods was basically pointless since we require the `@Metered` annotation (the logic to skip private methods was originally there when we automatically metered all extension methods - we no longer do that).


dreynaud:
LGTM, thanks!

PR #728, opened 2020-07-21
fix(plugins): Add missing @Bean annotation to serdeServiceSdkFactory

jonsie:


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/proxy/aspects/LogInvocationAspect.kt
```diff
@@ -50,12 +50,12 @@ class LogInvocationAspect : InvocationAspect<LogInvocationState> {
...
 
-    log.trace("Invoking method={} on extension={}", logInvocationState.methodName,
-      logInvocationState.extensionName)
-
     setOrRemoveMdc(Header.PLUGIN_ID.header, descriptor.pluginId)
     setOrRemoveMdc(Header.PLUGIN_EXTENSION.header, logInvocationState.extensionName)
 
+    log.trace("Invoking method={} on extension={}", logInvocationState.methodName,
+      logInvocationState.extensionName)
+
```
jonsie, line 58: Really small random change - moved the log to after adding the MDC keys.

PR #729, opened 2020-07-24
fix(exception): change ValidationException to be a UserException sinc…

marchello2000:
…e that's how it is used


marchello2000:
actually, going to close this and make another exception

PR #730, opened 2020-07-27
feat(annotations): Adding DeprecationInfo annotation

robzienert:
There's been discussion of how to deprecate things in `-api` modules. We have some general guidance, but I wanted something a little more standard that we can easily point people to that will help guide them in providing useful code deprecation context.

Idea here is that any time we use the `@Deprecated` annotation, this annotation would also come along for the ride.

Rather than include a bunch of information on usage and process in the annotation docs, I figured pointing to the spinnaker.io website would be a better place, and we can improve the docs there.

### DIFF
--- a/kork-annotations/src/main/java/com/netflix/spinnaker/kork/annotations/DeprecationInfo.java
+++ b/kork-annotations/src/main/java/com/netflix/spinnaker/kork/annotations/DeprecationInfo.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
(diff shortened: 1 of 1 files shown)


marchello2000:
👍  does this/can this play with https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-deprecated/replace-with.html (for kotlin only, obv)?

robzienert:
@marchello2000 I updated to make this annotation use `replaceWith` similar to the Kotlin `@Deprecated` annotation. This annotation can and should be used in combination with the Kotlin one.

ajordens:
LGTM

PR #732, opened 2020-08-05
fix(misc): Move actuator config from kork-web into a separate module

srekapalli:


settings.gradle
```diff
@@ -27,6 +27,7 @@ enableFeaturePreview("VERSION_ORDERING_V2")
 
 include(
   "spinnaker-dependencies",
+  "kork-actuator",
```
ajordens, line 30: Completely random question ... `kork-actuator` or `kork-web-actuator`?

kork-actuator/kork-actuator.gradle
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+apply plugin: "java-library"
+apply from: "$rootDir/gradle/kotlin-test.gradle"
+
+dependencies {
+  api(platform(project(":spinnaker-dependencies")))
```
cfieber, line 22: should be able to make these implementation instead of api since no one will depend on this

settings.gradle
cfieber: `-actuator` lines up with spring-boot naming (`spring-boot-starter-actuator`)

PR #734, opened 2020-08-05
feat(plugins): Add remote plugin info release cache for plugins that contain remote extensions, add kork-api module

jonsie:
The basic idea here is we have a cache of enabled plugins that contain remote extensions.  When a cache change is detected, a corresponding `RemotePluginExtensionsChanged` event (either `ENABLED`, `DISABLED`, or `UPDATED`) is emitted.  These events can be used by services to wire up the remote extension bits (HTTP client and any configuration necessary).

This also introduces `kork-api` which I feel is useful here since I'm copying the `RemoteExtensions` model I added in front50 (and we do that same basic type of model copying in various places).

### DIFF
--- a/kork-api/src/main/java/com/netflix/spinnaker/kork/api/plugins/remote/RemoteExtension.java
+++ b/kork-api/src/main/java/com/netflix/spinnaker/kork/api/plugins/remote/RemoteExtension.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.kork.api.plugins.remote;
+
+import java.util.HashMap;
+import java.util.Map;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+/**
+ * A Spinnaker plugin's remote extension configuration.
+ *
+ * <p>This model is used by Spinnaker to determine which extension points and services require
+ * remote extension point configuration.
+ *
+ * <p>The plugin release {@link SpinnakerPluginInfo.Release#requires} field is used to inform
+ * Spinnaker which service to use in configuring the extension point {@link #type} and additionally
+ * if the remote extension is compatible with the running version of the Spinnaker service.
+ */
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class RemoteExtension {
+
+  /**
+   * The remote extension type. The remote extension is configured in the service that implements
+   * this extension type.
+   */
+  @Nonnull private String type;
+
+  /** Identifier of the remote extension. Used for tracing. */
+  @Nonnull private String id;
+
+  /**
+   * Outbound transport configuration for the remote extension point; the protocol to address it
+   * with and the necessary configuration.
+   */
+  @Nonnull private RemoteExtensionTransport transport = new RemoteExtensionTransport();
+
+  /** Configures the remote extension point. */
+  @Nullable private Map<String, Object> config;
+
+  @Data
+  @NoArgsConstructor
+  @AllArgsConstructor
+  public static class RemoteExtensionTransport {
+
+    @Nonnull private Http http = new Http();
+
+    @Data
+    @NoArgsConstructor
+    @AllArgsConstructor
+    public static class Http {
+
+      /** URL for remote extension invocation. */
+      @Nonnull private String url;
+
+      /** A placeholder for misc. configuration for the underlying HTTP client. */
+      @Nonnull private Map<String, Object> config = new HashMap<>();
+    }
+  }
+}
--- a/kork-plugins/kork-plugins.gradle
+++ b/kork-plugins/kork-plugins.gradle
@@ -5,6 +5,8 @@ apply from: "$rootDir/gradle/lombok.gradle"
 
 dependencies {
   api(platform(project(":spinnaker-dependencies")))
+  api project(":kork-api")
+
   implementation project(":kork-core")
   implementation project(":kork-exceptions")
   implementation project(":kork-security")
--- a/kork-plugins/src/main/java/com/netflix/spinnaker/config/PluginsAutoConfiguration.java
+++ b/kork-plugins/src/main/java/com/netflix/spinnaker/config/PluginsAutoConfiguration.java
@@ -42,6 +42,8 @@ import com.netflix.spinnaker.kork.plugins.update.downloader.FileDownloaderProvid
 import com.netflix.spinnaker.kork.plugins.update.downloader.SupportingFileDownloader;
 import com.netflix.spinnaker.kork.plugins.update.release.provider.AggregatePluginInfoReleaseProvider;
 import com.netflix.spinnaker.kork.plugins.update.release.provider.PluginInfoReleaseProvider;
+import com.netflix.spinnaker.kork.plugins.update.release.remote.RemotePluginInfoReleaseCache;
+import com.netflix.spinnaker.kork.plugins.update.release.source.Front50PluginInfoReleaseSource;
 import com.netflix.spinnaker.kork.plugins.update.release.source.LatestPluginInfoReleaseSource;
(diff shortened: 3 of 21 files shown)


kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/update/release/remote/RemotePluginInfoReleaseCacheTest.kt
```diff
@@ -0,0 +1,100 @@
...
+import com.netflix.spinnaker.kork.plugins.update.release.plugin3
+import com.netflix.spinnaker.kork.plugins.update.release.pluginNoReleases
+import com.netflix.spinnaker.kork.plugins.update.release.provider.AggregatePluginInfoReleaseProvider
+import com.netflix.spinnaker.kork.plugins.update.release.source.PluginInfoReleaseSource
+import dev.minutest.junit.JUnit5Minutests
+import dev.minutest.rootContext
+import io.mockk.mockk
+import org.springframework.context.ApplicationEventPublisher
+
+class RemotePluginInfoReleaseCacheTest : JUnit5Minutests {
```
jonsie, line 39: Test is a WIP.

kork-api/src/main/java/com/netflix/spinnaker/kork/api/plugins/remote/RemoteExtension.java
```diff
@@ -0,0 +1,76 @@
...
+ * <p>This model is used by Spinnaker to determine which extension points and services require
+ * remote extension point configuration.
+ *
+ * <p>The plugin release {@link SpinnakerPluginInfo.Release#requires} field is used to inform
+ * Spinnaker which service to use in configuring the extension point {@link #type} and additionally
+ * if the remote extension is compatible with the running version of the Spinnaker service.
+ */
+@Data
+@NoArgsConstructor
+public class RemoteExtension {
```
jonsie, line 39: If we think it makes sense to move this here I will make the necessary front50 update.

kork-api/src/main/java/com/netflix/spinnaker/kork/api/plugins/remote/RemoteExtension.java
```diff
@@ -0,0 +1,76 @@
...
+  @Nonnull private String id;
+
+  /**
+   * Outbound transport configuration for the remote extension point; the protocol to address it
+   * with and the necessary configuration.
+   */
+  @Nonnull private RemoteExtensionTransport transport = new RemoteExtensionTransport();
+
+  /** Configures the remote extension point. */
+  @Nullable private Map<String, Object> config;
```
jonsie, line 57: After thinking it over and based on some feedback from @robzienert I think it may make more sense to leave this untyped and then allow the specific service (Orca, Keel, etc) to map it to a specific remote extension object.  This gives us more flexibility.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/events/RemotePluginExtensionsLoaded.kt
```diff
@@ -0,0 +1,35 @@
...
+  pluginId: String,
+  version: String,
+  remoteExtensions: List<RemoteExtension>
+) {
+  enum class Status {
+    ENABLED,
+    DISABLED,
+    UPDATED
+  }
+}
```
jonsie, line 35: I haven't quite settled on a pattern yet for how exactly these events will be consumed and wire things up for the remote extension (HTTP client, necessary configuration) and I am open to suggestion.

kork-plugins/src/main/java/com/netflix/spinnaker/config/PluginsAutoConfiguration.java
```diff
@@ -199,6 +200,29 @@ public static PluginInfoReleaseProvider pluginInfoReleaseProvider(
...
+      ApplicationEventPublisher applicationEventPublisher,
+      SpinnakerUpdateManager updateManager,
+      SpinnakerPluginManager pluginManager,
+      SpringPluginStatusProvider springPluginStatusProvider) {
+    return new RemotePluginInfoReleaseCache(
+        new AggregatePluginInfoReleaseProvider(
+            Arrays.asList(
+                latestPluginInfoReleaseSource,
+                springPluginInfoReleaseSource,
+                preferredPluginInfoReleaseSource),
```
jonsie, line 218: Since these plugin releases are dynamic (more like Deck plugin releases in that way) I decided not to include the front50 pinning release source here - I feel like that is only really necessary for the plugins that are loaded in-process.

kork-plugins/kork-plugins.gradle
```diff
@@ -5,6 +5,8 @@ apply from: "$rootDir/gradle/lombok.gradle"
 
 dependencies {
   api(platform(project(":spinnaker-dependencies")))
+  api project(":kork-api")
```
jonsie, line 8: `api` dependency because `RemoteExtensions` is used in `SpinnakerPluginInfo` (which I have not moved to `kork-api` yet) and that model is pulled into Gate.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/release/remote/RemotePluginInfoReleaseCache.kt
```diff
@@ -0,0 +1,142 @@
...
+) {
+
+  private val log by lazy { LoggerFactory.getLogger(javaClass) }
+
+  // unbounded cache because we need an in-memory reference to all enabled plugins with remote extensions
+  private val pluginCache: Cache<String, PluginInfoRelease> = Caffeine.newBuilder()
+    .build<String, PluginInfoRelease>()
+
+  @Scheduled(fixedDelayString = "\${spinnaker.extensibility.remote-extension-loader.frequency:PT60S}")
+  fun refresh() {
```
jonsie, line 56: I wonder if we also want to call this manually when the application is starting up so we force refresh the remote plugins before the schedule kicks in.

kork-plugins/src/main/java/com/netflix/spinnaker/config/PluginsAutoConfiguration.java
robzienert: What happens when a new `PluginInfoReleaseSource` is added somewhere down the line? I'd have expected to see an injection of`Collection<PluginInfoReleaseSource>` and filter out the types you know you don't want. Was this route considered?

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/events/RemotePluginExtensionsLoaded.kt
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * A Spring [ApplicationEvent] that is emitted when a remote extension is loaded/unloaded.
+ *
+ * @param source The source of the event
+ * @param status Whether the remote extension is [ENABLED], [DISABLED], or [UPDATED]
+ * @param pluginId The plugin ID for the remote extension
+ * @param version The plugin release version
+ * @param remoteExtensions The list of remote extensions associated with the plugin release
+ */
+class RemotePluginExtensionsLoaded(
```
robzienert, line 33: nit: `data class`? 

The event name is kinda odd when combined with `DISABLED` status. Perhaps `RemotePluginChanged`?

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/events/RemotePluginExtensionsLoaded.kt
robzienert: Seems like this is out of scope of the PR? There's definitely a few ways forward here, but we haven't really built up the part where we're actually creating these remote extension transports yet. Perhaps we use `ObjectProvider` for components that are reloadable.

kork-plugins/src/main/java/com/netflix/spinnaker/config/PluginsAutoConfiguration.java
jonsie: Yes, that's what I initially implemented. But then I reverted it because I didn't want someone inadvertently changing the behavior of this cache.  Now that you bring it up though - I agree.  I think it's reasonable to assume in kork-plugins land that any additional `PluginInfoReleaseSource` will be processed by the aggregate.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/events/RemotePluginExtensionsLoaded.kt
jonsie: Yes.  I wasn't a fan of the name of this event.  `RemotePluginChanged` is 💯 !

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/events/RemotePluginExtensionsLoaded.kt
jonsie: Yep - out of scope for this PR.  I think this is the next PR, basically.

PR #737, opened 2020-08-06
feat(exceptions): add generic exception handler for UserException

marchello2000:


kork-web/src/main/java/com/netflix/spinnaker/kork/web/exceptions/GenericExceptionHandlers.java
```diff
@@ -134,6 +133,7 @@ private void storeException(
       HttpServletRequest request, HttpServletResponse response, Exception ex) {
     // store exception as an attribute of HttpServletRequest such that it can be referenced by
     // GenericErrorController
+    logger.warn("Handled error in generic exception handler", ex);
```
marchello2000, line 136: added this because otherwise there often no logs of the exception that caused this handler to fire...

ezimanyi:
@marchello2000 @cfieber : This looks to be breaking application creation (and potentially some other workflows that interact with `front50`).

The issue is that `front50` defines a `NotFoundException` to extend `UserException` [here](https://github.com/spinnaker/front50/blob/ac2e2a72cbc0b6e9a525947f6f926b1cff085ebd/front50-core/src/main/java/com/netflix/spinnaker/front50/exception/NotFoundException.java#L8).  This means that with the new logic, `front50` is returning a 400 if you request an object that does not exist.

In some places (particularly orca) we have logic that ignores a 404 response and handles a missing object as a special case.  One example is application creation, where we first check if the application exists (expecting a 404) and create it if it's missing.

I'm not sure whether the fix is to:
(1) no longer have `NotFoundException` extend `UserException` (interestingly the generic kork one does not, but front50 makes its own)
(2) add a more specific handler for `NotFoundException`
(3) somehow reduce the scope of this change

Either way I think we should fix/revert this reasonably quickly as the integration tests are completely broken (and in this case I don't think it's a false positive...I'm unable to create an application in my test spinnaker either).

ajordens:
My vote is to jettison the front50-specific `NotFoundException` and favor the kork generic ones that should work correctly here.

cc: @marchello2000

marchello2000:
Ugh... @ezimanyi, thanks for catching this.
if i make the change in Front50 (cracking it open now) will we be ok? i will search who else extendes userexception

ezimanyi:
`UserException` is fairly new, so I suspect that there aren't too many things around that extend it.  (I think that it's actually fairly recently that the `front50` `NotFoundException` was changed to extend it.). So my feeling is that fixing this in front50 (and doing a quick check for similar cases in other services) should be fine.

marchello2000:
made this: https://github.com/spinnaker/front50/pull/922 looking at other services nothing jumps out, (keel uses `userexception` a bunch and i've let keel devs know)

PR #738, opened 2020-08-07
fix(redis): Support passwords / SSL for jedis cluster

robzienert:
Reported by @kskewes


kskewes:
Closes https://github.com/spinnaker/spinnaker/issues/5470#issuecomment-664755104

I believe.

kskewes:
@german-muzquiz - Is there anything holding up this one from merging for 1.22?

PR #741, opened 2020-08-10
fix(artifact): Add subpath matching on ExpectedArtifact

edgarulg:
This PR is about add an extra decision on matching artifacts using the `metadata` field `subPath`.

Spinnaker fails with "matches multiple artifacts" error when using multiple Git/repo artifacts on two pipelines, the pipeline parent triggers pipeline children.

We have a parent pipeline that defines two git artifacts as below:
```
{
    "type": "git/repo",
    "reference": "https://github.com/edgarulg/test-repo",
    "version": "master",
    "metadata" : {
        "subPath": "config"
    }
}
{
    "type": "git/repo",
    "reference": "https://github.com/edgarulg/test-repo",
    "version": "master",
    "metadata" : {
        "subPath": "parent"
    }
}
```
And a child pipeline that defines a git artifact as below:
```
{
    "type": "git/repo",
    "reference": "https://github.com/edgarulg/test-repo",
    "version": "master",
    "metadata" : {
        "subPath": "test"
    }
}
```
When executing the parent pipeline, it try to match artifacts with the above examples but it fails because only `type`, `version`, `name` and `reference` are used to determine if an artifact match. That's why it throws a `InvalidRequestException`.

I am adding an extra decision using the `metadata` field `subPath` in order to make a distinction between git artifact to avoid this problem.

The `subPath` property can be found in here: https://spinnaker.io/reference/artifacts-with-artifactsrewrite/types/git-repo/


kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/ExpectedArtifact.java
```diff
@@ -95,6 +95,13 @@ public boolean matches(Artifact other) {
       return false;
     }
 
+    // applicable only for git artifacts that uses same URL and branch but different subpath
```
danielpeach, line 98: I would enforce this in the code below (and demonstrate this in the tests).

I might also link to the documentation for this feature (which you helpfully sent me, thank you!). I didn't know this feature / special metadata key existed.

edgarulg: I added more tests to demonstrate it and added the doc's link of the git artifact

edgarulg:
@ezimanyi thanks for your feedback, I agree with you on avoid break the abstraction of the ExpectedArtifact, for this issue I moved the fix to this PR: https://github.com/spinnaker/orca/pull/3859

I am going to close this PR because this won't be applicable.

PR #742, opened 2020-08-11
bump springboot to 2.2.7

nimakaviani:
this fixes https://github.com/spinnaker/keel/issues/1409

/cc @luispollo

### DIFF
--- a/spinnaker-dependencies/spinnaker-dependencies.gradle
+++ b/spinnaker-dependencies/spinnaker-dependencies.gradle
@@ -19,7 +19,7 @@ ext {
     retrofit2        : "2.8.1",
     spectator        : "0.103.0",
     spek             : "1.2.1",
-    springBoot       : "2.2.5.RELEASE",
+    springBoot       : "2.2.7.RELEASE",
     springCloud      : "Greenwich.SR2",
     springfoxSwagger : "2.9.2",
     swagger          : "1.5.20" //this should stay in sync with what springfoxSwagger expects


luispollo:
@cfieber as part of the SpringBoot stuff you've been working on, have you had any issues with upgrading to later versions? We'd like to bump up the version slightly here, but this got me wondering if there's anything blocking us from upgrading to the next major.

luispollo:
Ah. Spoke too soon. Found the PR where we had previously had to revert the exact same upgrade, @nimakaviani: https://github.com/spinnaker/kork/pull/665

@robfletcher would have more context.

nimakaviani:
hmmm interesting. would be good to know how we can move forward and resolve the conflict with internal metrics.

cfieber:
@luispollo I feel like its worth a try again to see if this is in a happy place for us, a lot has changed since the previous attempt got rolled back.

I'd suggest applying this locally and seeing if everything starts up fine running our internal wrapper

luispollo:
> I'd suggest applying this locally and seeing if everything starts up fine running our internal wrapper

Thanks @cfieber, will do. I think my concern is that this affects all services, not just keel, but I can try with a couple.

nimakaviani:
@luispollo did you by any chance get to try this version bump to see how things work?

robzienert:
Ping.

luispollo:
Sorry I missed the ping @robzienert and @nimakaviani. Will give this a try next week, so reopening this to track, and linking it to https://github.com/spinnaker/keel/issues/1409 which is the one I was keeping an eye on.

luispollo:
OK... I gave this a shot today. Things were looking good until I tried our orca wrapper. It failed to start with this exception:
```
Caused by: java.lang.invoke.LambdaConversionException: Invalid receiver type class org.springframework.boot.actuate.autoconfigure.metrics.export.atlas.AtlasProperties; not a subtype of implementation type class org.springframework.boot.actuate.autoconfigure.metrics.export.properties.PushRegistryProperties
	at java.base/java.lang.invoke.AbstractValidatingLambdaMetafactory.validateMetafactoryArgs(AbstractValidatingLambdaMetafactory.java:254)
	at java.base/java.lang.invoke.LambdaMetafactory.metafactory(LambdaMetafactory.java:328)
	at java.base/java.lang.invoke.BootstrapMethodInvoker.invoke(BootstrapMethodInvoker.java:127)
	... 65 common frames omitted
```

There was a change to the `AtlasProperties` class hierarchy per https://github.com/spring-projects/spring-boot/issues/20843, however this breaks this app for us. I was unable to determine whether this is actually a bug in Spring Boot itself (since all other monitoring system properties classes seem to still extend `PushRegistryProperties` in 2.2.7) or a version incompatibility with internal Netflix libraries that setup Spectator. I don't have more bandwidth to look at this at this time, but wanted to let you know what I found.

In any case, @nimakaviani, based on our last conversation in the SIG, it didn't sound like you were blocked by this anymore, right?

PR #744, opened 2020-08-12
feat(retry): adds a configuration-binding friendly retry utility

cfieber:


cfieber:
I am going through and adding some retries through the fiat authorize call path around a few things like redis reads, and I ended up wanting to duplicate this pattern of tunable config around retries in a few different config properties so I stuck it in kork-exceptions

I didn’t necessarily want to go all the way to introducing a resilience4j dependency in here, this feels like a decent interim step around spots where RetrySupport isn’t the worst thing to use

cfieber:
I'm revisiting this and adding some resilience4j retry stuff in kork instead

PR #747, opened 2020-08-17
feat(plugins): Remote plugins configuration, cache, and remote transport with OkHttp3 client implementation

jonsie:
Downstream services can use the `RemotePluginsProvider` to get remote plugins by ID or by extension type.  This produces a `RemotePlugin` which includes the plugin ID, version, and a set of `RemoteExtension` objects.  `RemoteExtension` contains the extension ID, extension type, extension configuration (a bit of a TODO here - just a map - will need to iterate on that as this continues and maybe resolve it to a type), as well as the resolved remote extension transport mechanism - this PR features an OkHttp3 implementation.


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/remote/RemotePluginConfigChangedListener.kt
```diff
@@ -0,0 +1,92 @@
...
+    event.remoteExtensionConfigs.forEach { remoteExtensionConfig ->
+
+      // TODO(jonsie): Support enabling/disabling transports in the config.
+      // Configure HTTP if it is available since it is the only configurable transport right now.
+      val remoteExtensionTransport = if (remoteExtensionConfig.transport.http.url.isNotEmpty()) {
+        val client = okHttpClientProvider.getClient(
+          DefaultServiceEndpoint(
+            remoteExtensionConfig.id,
+            remoteExtensionConfig.transport.http.url,
+            remoteExtensionConfig.transport.http.config
```
jonsie, line 62: Previous change added the config here - this is resolved internally and used to be build up the client.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/remote/transport/OkHttpRemoteExtensionTransport.kt
```diff
@@ -0,0 +1,59 @@
...
+) : RemoteExtensionTransport {
+
+  override fun invoke(remoteExtensionPayload: RemoteExtensionPayload) {
+    AuthenticatedRequest.propagate {
+      val request = Request.Builder()
+        .url(url)
+        .post(
+          RequestBody.create(
+            MediaType.parse("application/json"),
+            objectMapper.writeValueAsString(remoteExtensionPayload)
```
jonsie, line 41: We'll need to register subtypes in downstream services for this to work.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/remote/transport/OkHttpRemoteExtensionTransport.kt
```diff
@@ -0,0 +1,59 @@
...
+          )
+        )
+        .build()
+
+      val response = client.newCall(request).execute()
```
jonsie, line 46: I would expect the response code to be a 202 but I didn't want to be strict about it - so long as the remote extension uses the provided callback (which will either be another PR or part of this one) we are 👍 .

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/events/RemotePluginCacheRefresh.kt
```diff
@@ -0,0 +1,19 @@
+package com.netflix.spinnaker.kork.plugins.events
+
+import com.netflix.spinnaker.kork.plugins.remote.RemotePluginsCache
+import org.springframework.context.ApplicationEvent
+
+/**
+ * A Spring [ApplicationEvent] that is emitted when the remote plugins cache is changed.
+ *
+ * The remote plugins cache is the cache of resolved remote plugins with remote transport
+ * clients.  This can optionally be used in downstream Spinnaker services to ensure a remote
```
robzienert, line 10: downstream of what?

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/remote/RemotePluginConfigChangedListener.kt
```diff
@@ -0,0 +1,92 @@
...
+          remoteExtensionConfig.type,
+          remoteExtensionConfig.config ?: mapOf(),
+          remoteExtensionTransport
+        )
+      )
+    }
+
+    val remotePlugin = RemotePlugin(event.pluginId, event.version, remoteExtensions)
+    remotePluginsCache.put(remotePlugin)
+    log.debug("Remote plugin '{}' added to cache due to '{}' event", event.pluginId, event.status.toString())
```
robzienert, line 86:
```suggestion
    log.debug("Remote plugin '{}' added to cache due to '{}' event", event.pluginId, event.status)
```

nit: `toString()` will be called for you.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/remote/transport/OkHttpRemoteExtensionTransport.kt
robzienert: Downstream of what?

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/remote/transport/RemoteExtensionTransport.kt
```diff
@@ -0,0 +1,41 @@
...
+   */
+  fun invoke(remoteExtensionPayload: RemoteExtensionPayload)
+}
+
+/**
+ * Marker interface for the payload for the remote extension - implemented at various extension
+ * points throughout the services.
+ */
+@Beta
+interface RemoteExtensionPayload
```
robzienert, line 41: nit: I'd move this out to a separate file.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/remote/transport/OkHttpRemoteExtensionTransport.kt
jonsie: Really just in services, I'm overusing the term downstream (since kork just runs in a service, it's not really downstream of anything).  🤷‍♂️

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/remote/RemotePluginsCache.kt
```diff
@@ -0,0 +1,62 @@
...
+import com.netflix.spinnaker.kork.plugins.events.RemotePluginCacheRefresh
+import org.slf4j.LoggerFactory
+import org.springframework.context.ApplicationEventPublisher
+
+/**
+ * Provides a read/write remote plugin cache used by either the [RemotePluginConfigChangedListener]
+ * to add plugins based on configuration changes or the [RemotePluginsProvider] to retrieve plugins
+ * for use in services.
+ */
+class RemotePluginsCache(
```
robzienert, line 31: Food for thought:

If a service already has `pluginId@0.1.0` and then gets a changed event for `pluginId@0.2.0`, we'll just overwrite the cache object and it gets dereferenced. There's nothing preventing a service developer from having assigned the plugin to some other class, which could lead to problems and they'd be kinda hard to trace.

Two ways I see this going:

1. We add remote plugin lifecycles, which allow us to `stop()` a remote plugin (basically set a flag that any access to it will cause `IllegalStateException`) when replacing it in the cache.
2. Make the remote plugin cache put operations more of an upsert, where the existing `RemotePlugin` object is updated with new values and refreshed when changes are made.

I kinda like the first option more at this moment.

jonsie: Yep, I've been mulling this over but haven't landed on a solution yet.  I think this can merge in for now and continue forward, but I agree - this is something I think I will need to address.

PR #752, opened 2020-08-19
feat(plugins): Spring-backed native plugin extensions

robzienert:
There's a lot here, all of which is gated on `spinnaker.extensibility.framework.version=V2`. Services will continue to use V1 without configuration, and all existing plugins will continue to work, including existing `kork-plugins-spring-api` based plugins. This V2 framework is not yet feature complete, but the diff was getting large enough that I didn't want to big-bang it.

This PR satisfies the happy-path of backing native extension by Spring, meaning plugins can define both arbitrary classes as well as extensions for autowiring handled by Spring. Service dependencies will be injected if they are part of the service API module, and `SpinnakerExtensionPoint` classes in the plugin will be auto-promoted to the service `ApplicationContext` for autowiring into core components.

This work shares a lot of the same workflows that `PrivilegedSpringPlugin` has to make it work. What's different is that Spring is not exposed as part of the API. This means it continues to be less powerful, but has the advantage of not being subject to as many `ClassLoader` and bytecode incompatibility errors. As time goes on, I imagine we'll be able to continue to close the gap between `kork-plugins-api` and the functionality offered by `kork-plugins-spring-api`.

One big benefit of this change is that the plugin framework, once it's all said and done, will not need its factory beans defined a `static`, which will open the doors to a lot of simplification across the codebase as we'll be able to lean more heavily on Spring. 🌈 

Major things not done in this PR (to be done in subsequent PRs):

- `@ExtensionConfiguration` is not yet created correctly.
- Extension method proxying (metrics, logs, etc). This will transition to Spring AOP.
- Expanded testing. I wasn't happy with the testing libs that we have today, so I started down the path of making some improvements, but more need to be done.
- Moving `v2` package into the rest of the codebase. I'm keeping it separate for my own sanity.


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/v2/ExtensionPromotionBeanPostProcessor.kt
```diff
@@ -0,0 +1,58 @@
...
+
+  override fun postProcessAfterInitialization(bean: Any, beanName: String): Any? {
+    val definition = pluginApplicationContext.getBeanDefinition(beanName)
+
+    Assert.notNull(definition.beanClassName, "Bean class name is null")
+
+    val beanClass = getBeanClass(definition.beanClassName!!)
+    if (SpinnakerExtensionPoint::class.java.isAssignableFrom(beanClass)) {
+      log.debug("Promoting plugin extension to service context (${pluginWrapper.pluginId}): $beanClass")
+      beanPromoter.promote("${pluginWrapper.pluginId}_${beanName}", bean, beanClass)
```
cfieber, line 48: I'm wondering what happens if someone creates in the plugin classes MyExtensionImpl that implements a defined extension point MyExtension, since the bean class is MyExtensionImpl and only known to the plugin classloader, is spring able to resolve that the bean is an instance of MyExtension?

I'm sort of wondering if this needs to register the bean class as the MyExtension class instead - but I have no idea whether this is an issue or not

kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/testplugin/GeneratedPluginFixture.kt
```diff
@@ -0,0 +1,82 @@
...
+import java.nio.file.Files
+import java.nio.file.Path
+
+/**
+ * Standard fixture for tests that need to generate a plugin.
+ *
+ * It is recommended to create a companion object to wrap this fixture so that a new plugin is
+ * not generated for each test case.
+ */
+class GeneratedPluginFixture(pluginName: String, builder: TestPluginBuilder.() -> Unit) {
```
jonsie, line 34: I wonder if there is an opportunity here to unify what we have in `kork-plugins-tck`.  We don't actually use `TestPluginBuilder` there, but instead use `PluginJar` and `PluginZip`.  Here's an example (and we also have examples in Orca, Echo, and Gate):

https://github.com/spinnaker/kork/blob/master/kork-plugins-tck/src/test/kotlin/com/spinnaker/netflix/kork/plugins/tck/PluginsTckFixtureTest.kt#L73

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/v2/ExtensionPromotionBeanPostProcessor.kt
robzienert: It seems to work fine. The `BeanPromoter` does register both the bean instance, as well as the class, with the parent bean factory.

kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/testplugin/GeneratedPluginFixture.kt
robzienert: Probably would be a good move. You're thinking of this for service owners, not plugin developers, right?

kork-plugins-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/PluginComponent.java
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.plugins.api;
+
+import com.netflix.spinnaker.kork.annotations.Beta;
+
+import java.lang.annotation.*;
```
luispollo, line 20: nit: single imports.

kork-plugins-api/src/main/java/com/netflix/spinnaker/kork/plugins/api/PluginComponent.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Marks a class within a plugin as one that is candidate for dependency injection.
+ *
+ * Classes defined within the plugin, as well as classes exposed via a service's API are valid candidates for
```
luispollo, line 25: The second part of this comment seems to be more encompassing than what's on line 23. Worth reconciling?

kork-plugins/src/main/java/com/netflix/spinnaker/config/PluginsAutoConfiguration.java
```diff
@@ -20,11 +20,7 @@
 import com.netflix.spinnaker.config.PluginsConfigurationProperties.PluginRepositoryProperties;
 import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService;
 import com.netflix.spinnaker.kork.dynamicconfig.SpringDynamicConfigService;
-import com.netflix.spinnaker.kork.plugins.ExtensionBeanDefinitionRegistryPostProcessor;
-import com.netflix.spinnaker.kork.plugins.SpinnakerPluginManager;
-import com.netflix.spinnaker.kork.plugins.SpinnakerServiceVersionManager;
-import com.netflix.spinnaker.kork.plugins.SpringPluginStatusProvider;
-import com.netflix.spinnaker.kork.plugins.SpringStrictPluginLoaderStatusProvider;
+import com.netflix.spinnaker.kork.plugins.*;
```
luispollo, line 23: Hmm, are we relaxing the style on star imports? Then I won't nitpick elsewhere. 🙂

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/v2/DefaultBeanPromoter.kt
```diff
@@ -0,0 +1,35 @@
...
+ * Promotes a given plugin bean to the provided [serviceApplicationContext].
+ */
+class DefaultBeanPromoter(
+  private val serviceApplicationContext: GenericApplicationContext
+) : BeanPromoter {
+
+  override fun promote(beanName: String, bean: Any, beanClass: Class<*>) {
+    serviceApplicationContext.registerBean(
+      beanName,
+      beanClass,
```
luispollo, line 30: (Just for my education) I'm curious how this works in the first place (even with `PrivilegedSpringPlugin`) since plugin classes are loaded in a separate class loader from the service's. Do the class loaders of the parent and the plugin interact in some way?

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/v2/PluginFrameworkInitializer.kt
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  override fun postProcessBeanFactory(beanFactory: ConfigurableListableBeanFactory) {
+  }
+
+  override fun setApplicationContext(applicationContext: ApplicationContext) {
+    ApplicationContextGraph.serviceApplicationContext = applicationContext
+  }
+
+  override fun getOrder(): Int = Ordered.LOWEST_PRECEDENCE
```
luispollo, line 48: Isn't this the other way around? `Ordered.HIGHEST_PRECEDENCE` is defined as `Integer.MIN_VALUE` (i.e. loads earlier).

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/v2/SpringPluginInitializer.kt
```diff
@@ -0,0 +1,79 @@
...
+    // which ones need to be promoted to the service ApplicationContext for autowiring into core functionality.
+    pluginApplicationContext
+      .beanFactory
+      .addBeanPostProcessor(ExtensionPromotionBeanPostProcessor(
+        pluginWrapper,
+        pluginApplicationContext,
+        beanPromoter
+      ))
+
+    pluginApplicationContext.classLoader = pluginWrapper.pluginClassLoader
```
luispollo, line 59: Is this where the service and the plugin "connect" as far as class loaders go?

kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/v2/fixtures/ParentServiceBean.kt
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.plugins.v2.fixtures
+
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.stereotype.Component
+
+@Component("parentServiceBean")
+class ParentServiceBean(@Value("\${someKey}") val someKey: String)
```
luispollo, line 22: Is this used anywhere in the tests?

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/v2/PluginFrameworkInitializer.kt
robzienert: Ha. Whoops, yes.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/config/V2PluginConfiguration.kt
```diff
@@ -0,0 +1,65 @@
...
+import org.pf4j.PluginFactory
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.context.ApplicationEventPublisher
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import org.springframework.context.support.GenericApplicationContext
+
+@Configuration
+@ConditionalOnProperty(value = ["spinnaker.extensibility.framework.version"], havingValue = FRAMEWORK_V2)
+class V2PluginConfiguration {
```
jonsie, line 37: Since all the other configuration classes are Java, and we're importing this one into `PluginsAutoConfiguration`, should we make this one Java just to be uniform?

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/v2/SpinnakerPluginService.kt
```diff
@@ -0,0 +1,99 @@
...
+      }
+    }
+  }
+
+  /**
+   * Start the plugins, attaching exported plugin extensions to the provided [registry].
+   */
+  fun startPlugins(registry: BeanDefinitionRegistry) {
+    // Start plugins. This should only be called once.
+    pluginManager.startPlugins()
```
jonsie, line 76: Wrap this in `withTiming`?  Might be useful especially if we start doing more with a plugin `start` hook.

kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/testplugin/GeneratedPluginFixture.kt
jonsie: Yeah, `kork-plugins-tck` is used in services.  Though, it could be used in the framework too.  I think the plugin developer testing stuff would be a different library.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/config/V2PluginConfiguration.kt
robzienert: I would really rather not; I'd prefer to get rid of the last remnants of Java in `kork-plugins` instead!

kork-plugins/src/main/java/com/netflix/spinnaker/config/PluginsAutoConfiguration.java
robzienert: I'm doing whatever Spotless is doing. 🤷

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/v2/DefaultBeanPromoter.kt
robzienert: This is more information than you asked for, but I think we could use a top-to-bottom.

[`ClassLoader`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ClassLoader.html) is for loading classes; that is, they load compiled Java classes into the JVM. Once that's done, there's not much for them to do.

In Spinnaker's context, the service uses the default system `ClassLoader` (call it `ServiceClassLoader`), and then each plugin gets its own `PluginClassLoader`. `PluginClassLoader` is a _child_ of the `ServiceClassLoader`, so whatever classes `ServiceClassLoader` has available to it, so does the `PluginClassLoader`. A child `ClassLoader` will first look to itself for a class it can load, and if it can't, it'll go to its parent and so-on. A parent `ClassLoader` cannot load classes that are in a child.

An object created by `ClassLoaderA` is going to be able to interact with objects created by `ClassLoaderB` and vice-versa: It isn't _object isolation_, it's _class isolation_. It's just bytecode in a VM, so once the `Class` is loaded, we're OK to use it wherever. The part where things lead to sadness is when a `Class` loaded from `ClassLoaderA` interacts with a `Class` from `ClassLoaderB`, both of which use the same dependency (Jackson, Spring, etc.) as part of their contract (an `interface` or some other public member), but are using different _versions_ and then different _bytecodes_. In this case, Java cannot reconcile how these two classes can interact because the bytecodes are incompatible.

Going back to parent-child relationships of `ClassLoader`, you can see why having a conflicting version of Spring would cause issues because the `PluginClassLoader` loads _its definition of Spring_ for its own context, even though the `ServiceClassLoader` has already loaded another version, and the service has created a bunch of objects using a conflicting version.

Now, for the Spring bit that you commented on.

We're registering a `BeanDefinition` into the parent application, which has a bunch of metadata that Spring needs to know how, in what order, etc. to create your bean. An `ApplicationContext` is tightly-coupled to one `ClassLoader`, which a `BeanFactory` will use to load the class -- this won't work for our case, obviously, since the service `ClassLoader` doesn't know about plugin classes. To get around this, I'm providing a `Supplier` as a factory method to the bean instance that was already created from the `PluginClassLoader` -- effectively making this plugin extension a singleton within the scope of the service. Furthermore, Spring is able to do all of its autowiring and what have you because I'm also providing that BeanFactory with the `Class` of the bean that we're promoting, so it'll be able to do whatever reflection magic it needs to do.

Hope that helps, I gotta run to a meeting, so that's the abrupt end to this comment.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/v2/SpringPluginInitializer.kt
robzienert: Yeah, kind of. They don't really connect (aside from parent/child relationships of the class loaders themselves).

The service has its own ClassLoader and each plugin has its own ClassLoader. Similarly, the service has one Spring `ApplicationContext` and each plugin has its own `ApplicationContext` in a parent/child relationship. Just classes are isolated from each other with class loaders, the different `ApplicationContext`s isolate dependency graphs from each other, with child `ApplicationContexts` being able to reach into the parent for bean definitions should one not be available.

robzienert:
> Ah, one thing I looked for in particular was a test replicating the real world example we're shooting for

@luispollo Now that #753 is merged, I added a test case that explicitly handles your use case: https://github.com/spinnaker/kork/pull/752/files#diff-0b92bf0c1274a71ee654bd7b9c783ad0

kork-plugins/src/main/kotlin/com/netflix/spinnaker/config/V2PluginConfiguration.kt
jonsie: Would be nice, but haven't we seen issues with loading Kotlin ConfigurationProperties or Configuration classes though?  This is a bit vague, but I seem to recall an issue with that.

PR #759, opened 2020-08-25
feat(plugins): Support for remote extension definitions and resolving remote extension configuration type for remote plugins/extensions

jonsie:
Required for: https://github.com/spinnaker/orca/pull/3875

As I was implementing the remote extension/plugin in Orca I realized it would be very useful to be able to declare a bean that defines the remote extension so we can resolve the remote extension configuration type (and anything else that may need to be resolved in the future).


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/remote/extension/RemoteExtension.kt
```diff
@@ -46,11 +45,16 @@ class RemoteExtension(
...
    * prior to the remote extension invocation.
    */
-  val config: Map<String, Any>,
+  val config: RemoteExtensionConfigType,
 
   private val transport: RemoteExtensionTransport
 ) {
 
+  @Suppress("UNCHECKED_CAST")
+  fun <T: RemoteExtensionConfigType> getTypedConfig(): T {
```
jonsie, line 54: This works, but I feel like there may be a better way.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/remote/RemotePlugin.kt
```diff
@@ -18,6 +18,7 @@
 package com.netflix.spinnaker.kork.plugins.remote
 
 import com.netflix.spinnaker.kork.annotations.Beta
+import com.netflix.spinnaker.kork.plugins.remote.extension.RemoteExtension
```
robzienert, line 21: nit: Looks like we can clean this import up.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/remote/extension/RemoteExtensionDefinition.kt
```diff
@@ -0,0 +1,18 @@
+package com.netflix.spinnaker.kork.plugins.remote.extension
+
+/**
+ * Define properties of the remote extension. Spinnaker services must implement this and provide it
+ * as a Spring bean in order for remote plugin extensions to be resolved from the configuration.
+ */
+interface RemoteExtensionDefinition {
```
robzienert, line 7: If the service is implementing this, would it not be a `RemoteExtensionPointDefinition`?

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/remote/RemotePlugin.kt
jonsie: There's a small package refactor in this PR - this is the result of that.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/remote/extension/RemoteExtensionDefinition.kt
jonsie: Dig it.

PR #760, opened 2020-08-25
chore(detekt): Warn on undocumented public members; ignore SpreadOperator, ReturnCount

robzienert:


robzienert:
It's just warning at this point. I'd like to get the warnings cleaned up and then start failing the builds once we're fixed up.

PR #761, opened 2020-08-25
fix(plugins): Refactor PluginSdk to also allow using plugin class

robzienert:
I thought I could be lazy and just pass `Any::class.java` to the `SdkFactory` instead of the extension class, which is not available. This didn't work so hot because the resource loader SDK uses the value to load resources from the class path.

With this fix in, it looks like all of our plugins load as-is without modification under V2.


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/v2/context/PluginSdksRegisteringCustomizer.kt
```diff
@@ -34,7 +34,7 @@ class PluginSdksRegisteringCustomizer(
   override fun accept(plugin: Plugin, context: ConfigurableApplicationContext) {
     val sdk = PluginSdksImpl(
       serviceApplicationContext.getBeansOfType(SdkFactory::class.java).values
-        .map { it.create(Any::class.java, plugin.wrapper) }
+        .map { it.create(plugin.javaClass, plugin.wrapper) }
```
robzienert, line 37: This is the actual fix. Literally everything else is extra.

PR #764, opened 2020-08-29
feat(tooling): Enable detekt failures; related refactors to conform 

robzienert:
There wasn't a lot here. The docs can still be improved for sure, but this was the first step.


kork-plugins/README.md
```diff
@@ -0,0 +1,20 @@
+# kork-plugins
```
robzienert, line 1: I realize this README is pretty awful, but getting a good readme together for this module will take some time, so I just threw some things todo in here.

robzienert:
Hmm... strangely, detekt's gradle plugin doesn't allow scoping the `ignoreFailures` flag to a single module. 😓  I guess I'll work on getting all of the other modules sorted out as well.

robzienert:
Ah, parallel detekt execution is incompat with submodule-specific `ignoreFailures` settings.

PR #755, opened 2020-08-21
feat(secrets): Azure support for secrets

bpowell:

PR #765, opened 2020-08-31
fix(secrets): revert Azure support for secrets until orca can be fixed

kevinawoo:
Reverts spinnaker/kork#755 until https://github.com/spinnaker/orca/pull/3882 can be fixed.


kevinawoo:
cc @bpowell

PR #766, opened 2020-09-01
fix(plugins): Register PluginSdks to service application context

robzienert:
We are observing issues of promoting beans that have `PluginSdks` in the constructor not being able to autowire the dependency. This actually happens after the extension has been initialized and successfully autowired with `PluginSdks` - it's occurring when promoting the extension to the service application context. For some reason, the parent service needs to know about the `PluginSdks` instance that has already been autowired.

This side-steps the issue by registering the `PluginSdks` implementation into the service application context, so the child looks up to the parent while resolving.


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/v2/context/PluginSdksRegisteringCustomizer.kt
```diff
@@ -36,6 +36,6 @@ class PluginSdksRegisteringCustomizer(
       serviceApplicationContext.getBeansOfType(SdkFactory::class.java).values
         .map { it.create(plugin.javaClass, plugin.wrapper) }
     )
-    context.beanFactory.registerSingleton("pluginSdks", sdk)
+    serviceApplicationContext.beanFactory.registerSingleton("${plugin.wrapper.pluginId}-pluginSdks", sdk)
```
luispollo, line 39: 😲

PR #767, opened 2020-09-01
fix(client): add provider for insecure okhttp client

srekapalli:
- Adding a new provider impl to create insecure client. 
- This will allow us to move to ok3 clients in all the places where we create ok2 clients that call external endpoints.


kork-web/src/main/java/com/netflix/spinnaker/config/DefaultServiceEndpoint.java
```diff
@@ -31,12 +31,24 @@
...
+    this.isSecure = true;
+    this.config = new HashMap<>();
+  }
+
+  public DefaultServiceEndpoint(
+      @Nonnull String name, @Nonnull String baseUrl, @Nonnull Boolean isSecure) {
+    this.name = Objects.requireNonNull(name);
+    this.baseUrl = Objects.requireNonNull(baseUrl);
+    this.isSecure = Objects.requireNonNull(isSecure);
     this.config = new HashMap<>();
```
jonsie, line 52: Let's set the defaults `isSecure` and `config` on the class fields instead of in the constructors.

kork-web/src/main/java/com/netflix/spinnaker/config/okhttp3/OkHttpClientBuilderProvider.java
```diff
@@ -33,8 +33,9 @@
    * @return true if supports the url given
    */
   default Boolean supports(ServiceEndpoint service) {
-    return service.getBaseUrl().startsWith("http://")
-        || service.getBaseUrl().startsWith("https://");
+    return ((service.getBaseUrl().startsWith("http://")
+            || service.getBaseUrl().startsWith("https://"))
+        && service.isSecure());
```
jonsie, line 38: Seems like the operators should be lined up?

kork-web/src/main/java/com/netflix/spinnaker/config/DefaultServiceEndpoint.java
```diff
@@ -31,12 +31,24 @@
...
     this.baseUrl = Objects.requireNonNull(baseUrl);
+    this.isSecure = true;
+    this.config = new HashMap<>();
+  }
+
+  public DefaultServiceEndpoint(
+      @Nonnull String name, @Nonnull String baseUrl, @Nonnull Boolean isSecure) {
+    this.name = Objects.requireNonNull(name);
+    this.baseUrl = Objects.requireNonNull(baseUrl);
+    this.isSecure = Objects.requireNonNull(isSecure);
```
robzienert, line 51: Just make `isSecure` an un-boxed primitive and then you can avoid `@Nonnull` and this `Objects.requireNonNull` check.

kork-web/src/main/java/com/netflix/spinnaker/config/DefaultServiceEndpoint.java
```diff
@@ -31,12 +31,24 @@
...
     this.name = Objects.requireNonNull(name);
     this.baseUrl = Objects.requireNonNull(baseUrl);
+    this.isSecure = true;
+    this.config = new HashMap<>();
+  }
+
+  public DefaultServiceEndpoint(
+      @Nonnull String name, @Nonnull String baseUrl, @Nonnull Boolean isSecure) {
+    this.name = Objects.requireNonNull(name);
+    this.baseUrl = Objects.requireNonNull(baseUrl);
```
robzienert, line 50: Don't believe you need to have `Objects.requireNonNull` if you're defining the fields as `@Nonnull`.

kork-web/src/main/java/com/netflix/spinnaker/config/okhttp3/OkHttpClientBuilderProvider.java
robzienert: Spotless handles formatting.

kork-web/src/main/java/com/netflix/spinnaker/config/DefaultServiceEndpoint.java
srekapalli: we won't able to set them via constructor if we set the values on the class fields and we want those to be immutable so.

kork-web/src/main/java/com/netflix/spinnaker/config/DefaultServiceEndpoint.java
srekapalli: Are we guarded during Runtime ..?  `@Nonnull` will do the IDE level checks only I thought ..

robzienert: Ah, true. nvm then!

robzienert (approved):

PR #770, opened 2020-09-02
refactor(test): Migrate Spek to Spek2

robzienert:
Could've removed Spek from kork entirely, but I wanted to see how Spek 2.x works. 

Supposedly it has better IDE performance than Spek 1.x, but I haven't seen evidence of this yet via kork: Orca has some really large, phenomenally slow tests that could be upgraded that should give better results. Spek 1.x and Spek 2.x can live side-by-side, so I'll try converting a single test to compare, but kork has been entirely migrated.

### DIFF
--- a/kork-sql/src/test/kotlin/com/netflix/spinnaker/kork/sql/PagedIteratorSpec.kt
+++ b/kork-sql/src/test/kotlin/com/netflix/spinnaker/kork/sql/PagedIteratorSpec.kt
@@ -28,42 +28,38 @@ import com.nhaarman.mockito_kotlin.whenever
 import org.assertj.core.api.AbstractAssert
 import org.assertj.core.api.Assertions.assertThat
 import org.assertj.core.api.Assertions.assertThatThrownBy
-import org.jetbrains.spek.api.Spek
-import org.jetbrains.spek.api.dsl.describe
-import org.jetbrains.spek.api.dsl.given
-import org.jetbrains.spek.api.dsl.it
-import org.jetbrains.spek.api.dsl.on
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.style.gherkin.Feature
 
 internal object PagedIteratorSpec : Spek({
 
-  describe("fetching paged data") {
-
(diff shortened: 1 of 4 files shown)


kork-sql/src/test/kotlin/com/netflix/spinnaker/kork/sql/PagedIteratorSpec.kt
```diff
@@ -28,42 +28,38 @@ import com.nhaarman.mockito_kotlin.whenever
...
-      beforeGroup {
+      beforeScenario {
         whenever(nextPage(eq(pageSize), anyOrNull())) doReturn emptyList<String>()
       }
 
-      afterGroup { reset(nextPage) }
+      afterScenario { reset(nextPage) }
 
-      it("has no elements") {
+      When("has no elements") {
```
robfletcher, line 49: Shouldn't that be a `Then` rather than a `When`?

kork-sql/src/test/kotlin/com/netflix/spinnaker/kork/sql/PagedIteratorSpec.kt
```diff
@@ -107,32 +103,32 @@ internal object PagedIteratorSpec : Spek({
...
+      afterScenario { reset(nextPage) }
 
-      it("has some elements") {
+      Given("has some elements") {
         assertThat(subject.hasNext()).isTrue()
       }
 
-      on("draining the iterator") {
-        val results = mutableListOf<String>()
+      val results = mutableListOf<String>()
```
robfletcher, line 112: Seems weird you can have a declaration outside the DSL scope like this

robzienert: Maybe weird, but if you don't, the class won't compile - if you wanted to keep all variable definitions inside of the DSL scopes, you'd need to assign it into some holder object that is then passed implicitly from one function to the next. It's really no different than defining variables outside `Scenario` in this case?

PR #774, opened 2020-09-03
feat(web): Add optional Zipkin integration to RawOkHttpClientFactory

robzienert:
Doing the bare minimum for OSS here; getting Zipkin wired up into okhttp, but not configuring Sleuth. I suspect it's a slim chance that Zipkin is going to be utilized outside of Netflix, so I figured I would spare the effort of wiring that up for OSS.

Does this seem reasonable? Sleuth can always be added later for everyone else if there's demand.

### DIFF
--- a/kork-retrofit/kork-retrofit.gradle
+++ b/kork-retrofit/kork-retrofit.gradle
@@ -9,6 +9,7 @@ dependencies {
   implementation "com.squareup.retrofit:retrofit"
   implementation "com.jakewharton.retrofit:retrofit1-okhttp3-client"
   implementation "com.squareup.retrofit:converter-jackson"
+  implementation "io.zipkin.brave:brave-instrumentation-okhttp3"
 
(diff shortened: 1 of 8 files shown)


jonsie:
I think this looks more or less correct from an OkHttp3 client instrumentation perspective.  However, decorating the HttpTracing object with tags, server URL, sample rate, etc - I'm not sure if we get that for free from Sleuth/SBN-tracing or if we need to do that ourselves?

robzienert:
The bean is annotated as `@ConditionalOnMissingBean`, I was planning that we would wire that up in our own stuff and pull stuff into OSS if it makes sense.

jonsie (approved):
Oh right.  Yeah I think this should be good to merge then.

celiawa:
Hi @robzienert , could I know is there any guide on how to config  Zipkin integration?

PR #775, opened 2020-09-04
feat(credentials): Add credentials abstraction

ncknt:
This is the first part of the work related to https://github.com/spinnaker/governance/blob/master/rfc/account-management.md.

It adds the concept of `Credentials`, `CredentialsRepository` that stores credentials of a given type, as well as bricks to populate and manage these credentials:
- `CredentialsDefinition` is the data with which we can build `Credentials`. It's optional but most implementation have them deserialized from yaml/json/something else.
- `CredentialsDefinitionSource` represents a source of definitions. By default, it's a getter on a Spring configuration object. It is pluggable and could make external calls - or just read a local file.
- `CredentialsParser` take a definition and transforms it into `Credentials`. It typically wires Spinnaker internals to do so (e.g. using Spring managed beans)
- `AbstractCredentialsLoader` keeps a source in sync with a repository. `BaseCredentialsLoader` does it by getting the full credentials list and comparing what is in the repository. Another possible implementation could be event based (receiving changes from an external system).

Finally, the `poller` package provides a way to frequently poll a credentials loader using Spring scheduled tasks.


ncknt:
@robzienert would you mind taking a look?

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/CompositeCredentialsRepository.java
```diff
@@ -0,0 +1,83 @@
...
+    allRepositories = new HashMap<>();
+    repositories.forEach(this::registerRepository);
+  }
+
+  public void registerRepository(CredentialsRepository<? extends T> repository) {
+    allRepositories.put(repository.getType(), repository);
+  }
+
+  public T getCredentials(String accountName, String type) {
+    if (accountName == null || accountName.equals("")) {
```
robzienert, line 40: Could be `Strings.isEmptyOrNull()`?

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/CompositeCredentialsRepository.java
```diff
@@ -0,0 +1,83 @@
...
+      throw new IllegalArgumentException("An account name must be supplied");
+    }
+
+    CredentialsRepository<? extends T> repository = allRepositories.get(type);
+    if (repository == null) {
+      throw new IllegalArgumentException("No credentials of type '" + type + "' found");
```
robzienert, line 46: Please make use of `SpinnakerException` and its child exception classes, making subclasses where it makes sense.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/Credentials.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package com.netflix.spinnaker.credentials;
+
+/** Credentials to an external system. Each credentials has a unique name for its type. */
+public interface Credentials {
+
+  String getName();
+
+  String getType();
```
robzienert, line 24: Need docs. What's the difference between these two, what are they used for?

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/CredentialsLifecycleHandler.java
```diff
@@ -0,0 +1,51 @@
...
+ * After {@link Credentials} have been parsed, they can be activated, refreshed, or retired - e.g.
+ * adding agents. This happens before credentials are added or updated in the {@link
+ * CredentialsRepository} and after credentials are removed from the {@link CredentialsRepository}.
+ *
+ * @param <T>
+ */
+public interface CredentialsLifecycleHandler<T extends Credentials> {
+
+  /**
+   * Credentials have been added. This is called before credentials are available in @{@link
```
robzienert, line 29:
```suggestion
   * Credentials have been added. This is called before credentials are available in {@link
```

robzienert: This change should be applied to the other areas this happens, too.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/CredentialsRepository.java
```diff
@@ -0,0 +1,38 @@
...
+
+  boolean has(String name);
+
+  Set<T> getAll();
+
+  T save(T credentials);
+
+  void delete(String name);
+
+  String getType();
```
robzienert, line 37: Docs on interfaces and their methods, as well as any classes that do not implement an interface and cannot inherit the docs from a parent.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/MapBackedCredentialsRepository.java
```diff
@@ -0,0 +1,69 @@
...
+
+package com.netflix.spinnaker.credentials;
+
+import java.util.*;
+import javax.annotation.Nonnull;
+import lombok.Getter;
+
+public class MapBackedCredentialsRepository<T extends Credentials>
+    implements CredentialsRepository<T> {
+  protected Map<String, T> credentials = new HashMap<>();
```
robzienert, line 25: Are we at risk of concurrent modification issues by having this just be a `HashMap`?

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/definition/AbstractCredentialsLoader.java
```diff
@@ -0,0 +1,35 @@
...
+@NonnullByDefault
+public abstract class AbstractCredentialsLoader<T extends Credentials> {
+  @Getter protected final CredentialsRepository<T> credentialsRepository;
+
+  public AbstractCredentialsLoader(CredentialsRepository<T> credentialsRepository) {
+    this.credentialsRepository = credentialsRepository;
+  }
+
+  @PostConstruct
+  public abstract void load();
```
robzienert, line 34: Docs would be useful here. Since the return value is `void`, I'm having to assume each implementor would be responsible for adding the final loaded credentials into the `CredentialsRepository` themselves?

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/definition/BasicCredentialsLoader.java
```diff
@@ -0,0 +1,94 @@
...
+  protected final CredentialsParser<T, U> parser;
+  protected final CredentialsDefinitionSource<T> definitionSource;
+  /**
+   * When parallel is true, the loader may apply changes in parallel. See {@link
+   * java.util.concurrent.ForkJoinPool} for limitations. This can be useful when adding or updating
+   * credentials is expected to take some time, as for instance when making a network call.
+   */
+  @Setter @Getter protected boolean parallel;
+  // Definition is kept so we can quickly check for changes before parsing
+  protected final Map<String, T> loadedDefinitions = new HashMap<>();
```
robzienert, line 47: Also curious about concurrency here.

kork-credentials/kork-credentials.gradle
```diff
@@ -0,0 +1,38 @@
...
+apply plugin: "java-library"
+apply from: "$rootDir/gradle/lombok.gradle"
+
+dependencies {
+  implementation(platform(project(":spinnaker-dependencies")))
+  api project(":kork-core")
+  api "com.fasterxml.jackson.core:jackson-annotations"
+
+  testImplementation "org.spockframework:spock-core"
+  testImplementation "org.spockframework:spock-spring"
```
robzienert, line 26: You're not using Spock (yay!), don't need these deps.

kork-credentials/kork-credentials.gradle
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+apply plugin: "java-library"
+apply from: "$rootDir/gradle/lombok.gradle"
+
+dependencies {
+  implementation(platform(project(":spinnaker-dependencies")))
+  api project(":kork-core")
+  api "com.fasterxml.jackson.core:jackson-annotations"
```
robzienert, line 23: Avoid `api` scope when possible. We're trying to be more mindful of transient dependencies.

Also, Jackson isn't even used in this PR, so it can be removed.

robzienert (changes requested):
Changes:

- Add `package-info.java` files that set up `@NonnullByDefault`, and add `@Nullable` where needed.

cfieber:
I'm curious why we are adding a kork-credentials module instead of adding this in clouddriver. What other services do you expect to use this library in?

ncknt:
@cfieber Igor and Kayenta mainly.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/CompositeCredentialsRepository.java
ncknt: I had copied the `IllegalArgumentException` from https://github.com/spinnaker/clouddriver/blob/master/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/ArtifactCredentialsRepository.java#L49 and tried to keep the same unchecked exception for backward compatibility. Adding a new exception for that.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/definition/AbstractCredentialsLoader.java
robzienert: I think it may be an improvement, actually, to do `public <? extends Credentials> load()` and just handle the wiring in `Poller`, this would remove some boilerplate from implementations of `CredentialsLoader` (and wouldn't require an abstract class, which is a big win), and would improve the testability of `CredentialsLoader` implementations: Just assert on the response instead of mocking / verifying interactions with the `CredentialsRepository`.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/MapBackedCredentialsRepository.java
ncknt: Yes, thanks!

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/definition/AbstractCredentialsLoader.java
ncknt: `load()` loads all credentials of a given type (e.g. all AWS accounts) in the repository. The default (`BasicCredentialsLoader`) implementation takes an account source (by default a getter on a Spring config object) -> passes it in a parser -> (we get a collection of `Credentials`) -> applies differences to the repository. It's simple to use for services using it, e.g. for k8s:

```
  public AbstractCredentialsLoader<KubernetesNamedAccountCredentials> kubernetesCredentialsLoader(
      KubernetesConfigurationProperties configurationProperties,
      KubernetesCredentials.Factory credentialFactory,
      CredentialsRepository<KubernetesNamedAccountCredentials> kubernetesCredentialsRepository) {
    return new BasicCredentialsLoader<>(
        configurationProperties::getAccounts,
        a -> new KubernetesNamedAccountCredentials(a, credentialFactory),
        kubernetesCredentialsRepository);
  }
```

The abstract class provides room for other implementations. For example, an event based implementation can query an external system and get responses such as `account X removed` or `new account Y added`. We have one implementation like that but it's hard to make generic enough to be included here.

The advantage for the developer of such an implementation is that the lifecycle (e.g. `add agent A and B`)  and the Spinnaker internal considerations (wire these beans to create Credentials) are separate and can evolve with each release without requiring maintenance of the associated plugin.

ncknt:
Updated the PR. I also added a check to make sure we only store credentials of the right type in a repository.

claymccoy (approved):
LGTM, it looks like you have resolved or responded to the other feedback and this seems like a useful approach.

ncknt:
I missed the comment about `NonnullByDefault`, updated now

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/CompositeCredentialsRepository.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    CredentialsRepository<? extends T> repository = allRepositories.get(type);
+    if (repository == null) {
+      throw new UnknownCredentialsTypeException("No credentials of type '" + type + "' found");
+    }
+
+    T creds = repository.getOne(credentialsName);
+    if (creds == null) {
+      throw new IllegalArgumentException(
```
robzienert, line 54: Would be useful to move to `SpinnakerException` family for all exceptions. There's a number of places in the PR that need to change beyond this one point. Also happy to explain this (or doc if I haven't) why it's beneficial to do so.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/definition/AbstractCredentialsLoader.java
robzienert: This is helpful, thanks. I think it would be beneficial to capture this reasoning in the abstract class docs.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/poller/Poller.java
```diff
@@ -0,0 +1,41 @@
...
+
+package com.netflix.spinnaker.credentials.poller;
+
+import com.netflix.spinnaker.credentials.Credentials;
+import com.netflix.spinnaker.credentials.definition.AbstractCredentialsLoader;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * A poller attempts to reload credentials from its source at a regular interval.
```
robzienert, line 25: How is the interval configured? For sake of docs, I see it's done via `PollerConfiguration` here - even just `{@link PollerConfiguration}` would be valuable.

robzienert (approved):
Just some small stuff I'd like to see changed before it goes in - looks marvelous aside.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/CompositeCredentialsRepository.java
ncknt: Oh I thought you had +1'd the backward compatibility of raising `IllegalArgumentException`. I like the typed exceptions.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/poller/Poller.java
ncknt: Added comments (`credentials.poller.<type of creds>.reloadFrequencyMs` -> `credentials.poller.default.reloadFrequencyMs`) + `credentials.poller.enabled`.

PR #778, opened 2020-09-09
fix(redis): fix parsing parameters from uri

bluestealth:
Fixes parsing of `password` when using redis clusters, previously this included the basic authentication separator `:`  as part of the password.
Fixes parsing of `database` when using redis node pools, previously the default database `0` was always returned.
Moves these functions into `RedisClientConnectionProperties` so that they can be shared between redis types.

### DIFF
--- a/kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/RedisClientConnectionProperties.java
+++ b/kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/RedisClientConnectionProperties.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/RedisClientConnectionProperties.java
```diff
@@ -0,0 +1,55 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.jedis;
+
+import java.net.URI;
+import redis.clients.jedis.Protocol;
+
+public class RedisClientConnectionProperties {
```
cfieber, line 21: could you add some unit tests around this class

cfieber (approved):

PR #782, opened 2020-09-15
feat(plugins): Add remote extension synchronous read and write capabilities

jonsie:
Tried to implement this in a transport agnostic way - ideally we should be able to swap the transport config out with a gRPC config and the interaction with `RemoteExtension` will stay the same.

- Also includes a small breaking change to remove usage of M<S, O> in ServiceEndpoint
- Remove usage of Provider in favor of ObjectProvider - I find the errors much easier to grok as they are native Spring errors, and the API is more robust.
- Small remote plugin package refactor


kork-web/src/main/java/com/netflix/spinnaker/config/ServiceEndpoint.java
```diff
@@ -33,7 +33,7 @@
 
   /** Misc. config necessary for the service client. */
   @Nonnull
-  public Map<String, Object> getConfig();
+  public Map<String, String> getConfig();
```
jonsie, line 36: I'll follow up where necessary (Gate and internal libraries) with this change.

kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/remote/extension/RemoteExtensionTest.kt
```diff
@@ -0,0 +1,65 @@
...
+      expectThat(result).isA<ReadResponse>()
+    }
+  }
+
+  private class Fixture {
+    val writeResponse = WriteResponse()
+    val readResponse = ReadResponse()
+    val remoteExtensionPayload = Payload()
+    val remoteExtensionQuery = Query()
+    val transport: RemoteExtensionTransport = mockk(relaxed = true)
```
robzienert, line 49: Interesting thought: We should support an in-process transport. I think it could be beneficial to converge on "remote-first" extension points, but still allowing in-process plugins to implement them. What do you think?

robzienert (approved):

kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/remote/extension/RemoteExtensionTest.kt
jonsie: Yes.  This goes in-line with what I was thinking WRT making the OkHttp client provider an in-process extension point too.  I think building remote extension points on-top of an extensible in-process transport is going to provide the flexibility we'll need for this to have wider adoption.

robzienert (approved):

PR #783, opened 2020-09-18
fix(misc): Move all the config server related implementation into sep…

srekapalli:
-- Moving config server related functionality into separate module so that we can optionally exclude.
-- Will add this as a new dependency in `Igor` and `Clouddriver`.


kork-cloud-config-server/kork-cloud-config-server.gradle
```diff
@@ -0,0 +1,23 @@
+apply plugin: "java-library"
+apply from: "$rootDir/gradle/lombok.gradle"
+
+dependencies {
+  api(platform(project(":spinnaker-dependencies")))
+
+  implementation "commons-io:commons-io"
+  implementation "org.apache.commons:commons-lang3"
+
+  api project(':kork-secrets')
```
cfieber, line 10: I expect this becomes a runtimeOnly dependency where it is used and these can become implementation instead of api

cfieber (approved):
just the api -> implementation in the gradle scopes but otherwise LGTM

kork-cloud-config-server/kork-cloud-config-server.gradle
```diff
@@ -0,0 +1,22 @@
...
+
+dependencies {
+  api(platform(project(":spinnaker-dependencies")))
+
+  implementation "commons-io:commons-io"
+  implementation "org.apache.commons:commons-lang3"
+  implementation project(':kork-secrets')
+  api project(':kork-config')
+  api "org.springframework.cloud:spring-cloud-context"
+  api "org.springframework.cloud:spring-cloud-config-server"
```
cfieber, line 12: these should be implementation as well

PR #785, opened 2020-09-22
feat(authz): Add AccessDeniedDecorator to kork-api to customize authz error messages

jonsie:


kork-api/src/main/java/com/netflix/spinnaker/kork/api/authz/AccessDeniedDecorator.java
```diff
@@ -0,0 +1,13 @@
...
+import com.netflix.spinnaker.kork.plugins.api.internal.SpinnakerExtensionPoint;
+
+/**
+ * Decorate the various authz access denied error messages that Spinnaker throws with custom
+ * additional information.
+ */
+public interface AccessDeniedDecorator extends SpinnakerExtensionPoint {
+
+  /** Decorates an access denied error based on the resource type and resource name. */
+  String decorate(String resourceType, String resourceName);
```
jonsie, line 12: We need to do some `fiat` refactoring to pull `fiat-api` over to kork and also move some things out of `fiat-core` (like `ResourceType` for example) -- so this method just takes strings.  We can add another method down the road that takes `ResourceType`.

jonsie:
Converted this to a draft as I'm now looking into how to make this more generic such that it supports specific exception types.

jonsie:
Closing this since I've made so many changes it doesn't make sense to keep this under this PR.

PR #787, opened 2020-09-23
feat(credentials): split kork-credentials and kork-credentials-api

ncknt:
Follow up to https://github.com/spinnaker/kork/pull/775 splitting b/w API classes and a (for now) small implementation of pollers.


robzienert (approved):
💥

PR #792, opened 2020-09-29
fix(dependencies): Remove explicit version for logging interceptor an…

srekapalli:
- Removing obsolete provider class
- Remove explicit version from the dep.


vigneshm (approved):
Thanks for the quick fix! 👍

jonsie (approved):

PR #793, opened 2020-09-30
feat(test): test that downloading a plugin emits an event

danielpeach:
We listen to these events and don't want them to get dropped in a future refactor...

Would it make sense to put these application events into the api package?


robzienert:
> Would it make sense to put these application events into the api package?

We certainly could do, but it'd need to be refactored to not have a Spring dependency. I presume you're using these in plugins themselves? @jonsie and I have been talking a little bit about making events a new extension point model, but haven't gotten around to it just yet.

robzienert (approved):

danielpeach:
No, in our extension projects, though there's nothing that would prevent us from moving that code into a plugin. Do you mean that the extension point would be a Spring-free wrapper around `ApplicationListener`? That sounds nice.

robzienert:
> Do you mean that the extension point would be a Spring-free wrapper around ApplicationListener?

Yeah, we'd need to get events that aren't necessarily an `ApplicationEvent` type (we'd probably just make a new `SpinnakerEvent` or something like that?) and then we'd be looking at an API around the more generic Spring `EventListener` and `EventPublisher`. We'd need to consider whether or not we want to expose the API as sync, async, or both for event processing, though. Orca supports both, so we could pretty easily bring that in?

PR #795, opened 2020-10-02
chore(dependencies): bump okhttp from 2.7.0 to 2.7.5 due to vulnerabilities

zachsmith1:
https://nvd.nist.gov/vuln/detail/CVE-2016-2402


kevinawoo (approved):
LGTM, initial look at the [diff between versions](https://github.com/square/okhttp/compare/parent-2.7.0...parent-2.7.5) looks like there's not a big impact on the stuff we'll be using.

PR #798, opened 2020-10-05
chore(dependencies): bring in specific spring version

cfieber:
this allows us to bump Spring to a newer version independent of bumping Spring-Boot


zachsmith1 (approved):
LGTM!

PR #799, opened 2020-10-06
fix(error): Errors rendered as markdown so we need two newlines characters for a newline to render correctly

jonsie:


robzienert (approved):
Are these errors always rendered as markdown?

jonsie:
> Are these errors always rendered as markdown?

Yes.

jonsie:
I should probably add that to the `ExceptionMessage` API doc.

cfieber (approved):

PR #800, opened 2020-10-06
feat(plugins): add test that demonstrates that secrets can be used in…

danielpeach:
… the v2 framework.

Secrets didn't work in the v1 framework.


jonsie (approved):
Nice.

PR #804, opened 2020-10-07
chore(moniker): merge moniker into kork

plumpy:
Autobumps won't pass since the artifact ID has changed, but I fixed up the dependencies in all 6 of those projects and verified they all still build.


cfieber (approved):

robzienert (approved):
Aw yeah, nice.

PR #806, opened 2020-10-08
feat(plugins): Include spinnaker-official and spinnaker-community repositories by default

robzienert:
I'd like to get these behind our own DNS and CDN eventually, but I think this may suffice for now?


claymccoy (approved):
LGTM

jonsie (approved):

PR #812, opened 2020-10-14
v2 plugin framework: allow service-side extension injection without special handling

danielpeach:


danielpeach:
Wanted to get some feedback on this idea. It would be very helpful for us to be able to support extensions that do not have [special handling at the injection point](https://github.com/spinnaker/orca/blob/master/orca-core/src/main/java/com/netflix/spinnaker/orca/DefaultStageResolver.java).

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/v2/SpinnakerPluginService.kt
```diff
@@ -84,7 +99,55 @@ class SpinnakerPluginService(
         }
 
         if (p is PluginContainer) {
-          p.registerInitializer(registry)
+          val pluginContext = p.pluginContext
+          val pluginId = p.wrapper.pluginId
+          val initializerBeanName = p.registerInitializer(registry)
+
+          // Find the plugin's SpinnakerExtensionPoints.
+          ClassPathScanningCandidateComponentProvider(false).apply {
```
danielpeach, line 107: The main idea is here. It creates lazy implementations for extension beans that will be created later on inside the plugin application context.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/v2/SpringPluginFactory.kt
```diff
@@ -39,16 +38,10 @@ class SpringPluginFactory(
...
   private val serviceApplicationContext: GenericApplicationContext,
-  private val invocationAspects: List<InvocationAspect<*>>
 ) : PluginFactory {
 
   private val log = LoggerFactory.getLogger(javaClass)
 
-  /**
-   * [DefaultBeanPromoter] is the One True [BeanPromoter] for non-testing scenarios.
-   */
-  private val beanPromoter = DefaultBeanPromoter(serviceApplicationContext)
```
danielpeach, line 50: There's no need for the bean promoter, since the proxies are injected into the parent context up front.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/v2/SpinnakerPluginService.kt
jonsie: Have you run the service tests (like orca-plugin-tests, echo-plugin-tests, gate-plugin-tests) against this change?

danielpeach: Yeah! All of the service plugin tests pass.

jonsie: Nice.

danielpeach:
Realizing this might be unclear / seem unmotivated...

A bunch of plugins at armory depend on interfaces that haven't been marked with `SpinnakerExtensionPoint`. This allows people to try out new ideas without having to figure out how to isolate interfaces into `<service>-api` packages and it's safer than depending on the Spring loader. 

Those plugins aren't going to work in the v2 framework partly because a). it doesn't promote random interfaces into the service application context and b). service-level classes that need to inject extension beans need to handle them in a special way (like using an `ObjectProvider`). The `ServiceInjectionScenarioTest` is a good example of this (a service bean trying to inject an extension bean - the test doesn't pass without the changes in this PR). 

Trying to solve the second problem here. The idea is that you can add proxies for extension beans into the service application context. These proxies are added early on in the service application lifecycle, so they can be injected into service beans without an ObjectProvider (or whatever other pattern you're using). Once you actually need to use a method on an extension bean, the proxy delegates down to the plugin application context and finds the implementation.

The piece that seems questionable to me is that you might call a method on the proxy before the plugin application context and its extension beans have been initialized. I've solved that for the moment by forcing `SpringPluginInitializer` to be initialized (if it hasn't been already) when a proxy method is called. This solves this problem, but I'm not 100% sure that there aren't weird DI bugs that could happen because of this. I've tried to root those out with the test cases I've added.

robzienert (approved):
Cool solution, but I suspect this is fraught with some potentially annoying edge cases. Specifically, this strategy will not work if you have something like the following:

```kotlin
interface MyExtenisonPoint {
  fun doStuff()
}

@Component
class Something(
  val extensions: List<MyExtensionPoint>
) {

  @PostConstruct
  fun stuff() {
    extensions.forEach { it.doStuff() }
  }

  // or...

  @Scheduled(withDelay = 1_000)
  fun scheduledStuff() {
    extensions.forEach { it.doStuff() }
  }
}
```

In both cases, the `Something` service component will invoke the lazy-initialized extensions before they're actually initialized and will cause an exception and the service will crash. I don't believe many of the `{serivce}-plugin-tests` will catch these scenarios.

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/proxy/LazyExtensionInvocationProxy.kt
```diff
@@ -0,0 +1,58 @@
...
+ * */
+internal class LazyExtensionInvocationProxy(
+  target: Lazy<SpinnakerExtensionPoint>,
+  private val targetClass: Class<out SpinnakerExtensionPoint>,
+  private val invocationAspects: List<InvocationAspect<InvocationState>>,
+  private val descriptor: SpinnakerPluginDescriptor
+) : ExtensionInvocationHandler {
+
+  private val target by target
+  private val delegate by lazy { ExtensionInvocationProxy(this.target, invocationAspects, descriptor) }
```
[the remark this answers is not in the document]
danielpeach: This works - added a test for it.

danielpeach:
Thanks for the review @robzienert. Just cleaning this up now.

About the concern you outlined above: calling a method on an extension bean works as you would hope. I've updated the tests to make this (hopefully) more clear. It would be bad if you couldn't know when an extension bean had been initialized.

Here's a spot where I verify that calling an extension method works properly: https://github.com/spinnaker/kork/pull/812/files#diff-1f0807f933c59a81d7c3c4b0c12cfcdd5b7c5a9c71fc089c04e2487008519d3eR126-R132

That test will fail if I comment out these lines, which force the plugin to be initialized if it hasn't been already: https://github.com/spinnaker/kork/pull/812/files#diff-4378d0609e86a5587a26780df19e8297ca9a94b718d8043e4133427c9d85bc4dR137-R138

PR #813, opened 2020-10-23
feat(api): Move ExpressionFunctionProvider to API

robzienert:
Migrated from Kotlin -> Java. Same exact functionaliity. New package, though, so this PR will break built-in integrations in Orca and... Echo, requiring a small adjustment.


jonsie (approved):

kork-expressions/src/main/kotlin/com/netflix/spinnaker/kork/expressions/ExpressionFunctionProvider.kt
```diff
@@ -1,151 +0,0 @@
...
- *  }
- * ```
- *
- * The above function provider could then be called in a SpEL expression:
- *
- * ```
- * ${netflix.hello()}
- * ```
- */
-interface ExpressionFunctionProvider : ExtensionPoint {
```
cfieber, line 52: not necessary by any means but do you think it would be worthwhile to leave this here as a deprecated noop interface extending the new ExpressionFunctionProvider in `-api`?

cfieber (approved):
LGTM

kork-expressions/src/main/kotlin/com/netflix/spinnaker/kork/expressions/ExpressionFunctionProvider.kt
robzienert: Oh, good call - yes, I will do that.

PR #816, opened 2020-10-30
fix(client): adding support to provide a plain okhttp client with sys…

srekapalli:
- This change allows us to migrate from `okhttp2` and use `clientProvider` implementation for all non-spinnaker endpoints like `slack`, `pagerduty` etc.

### DIFF
--- a/kork-web/src/main/java/com/netflix/spinnaker/config/okhttp3/DefaultOkHttpClientBuilderProvider.java
+++ b/kork-web/src/main/java/com/netflix/spinnaker/config/okhttp3/DefaultOkHttpClientBuilderProvider.java
@@ -71,10 +71,11 @@ public class DefaultOkHttpClientBuilderProvider implements OkHttpClientBuilderPr
   }
 
   protected OkHttpClient.Builder setSSLSocketFactory(
-      OkHttpClient.Builder builder, ServiceEndpoint service) {
+      OkHttpClient.Builder builder, ServiceEndpoint serviceEndpoint) {
 
-    if (okHttpClientConfigurationProperties.getKeyStore() == null
-        && okHttpClientConfigurationProperties.getTrustStore() == null) {
+    if ((okHttpClientConfigurationProperties.getKeyStore() == null
+            && okHttpClientConfigurationProperties.getTrustStore() == null)
+        || serviceEndpoint.isUseDefaultSslSocketFactory()) {
       return builder;
     }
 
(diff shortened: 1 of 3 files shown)


kork-web/src/main/java/com/netflix/spinnaker/config/okhttp3/DefaultOkHttpClientBuilderProvider.java
```diff
@@ -71,10 +72,11 @@ public DefaultOkHttpClientBuilderProvider(
...
 
   protected OkHttpClient.Builder setSSLSocketFactory(
-      OkHttpClient.Builder builder, ServiceEndpoint service) {
+      OkHttpClient.Builder builder, ServiceEndpoint serviceEndpoint) {
 
-    if (okHttpClientConfigurationProperties.getKeyStore() == null
-        && okHttpClientConfigurationProperties.getTrustStore() == null) {
+    if ((okHttpClientConfigurationProperties.getKeyStore() == null
+            && okHttpClientConfigurationProperties.getTrustStore() == null)
+        || serviceEndpoint instanceof PlainServiceEndpoint) {
```
jonsie, line 79: Hm, I feel like I'm missing something here.  So is it assumed that a `PlainServiceEndpoint` will have a keystore and a truststore?

srekapalli: We will assign system defaults to both and we will not use the keystore and truststore provided in the app configuration.

kork-web/src/main/java/com/netflix/spinnaker/config/PlainServiceEndpoint.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.Map;
+import javax.annotation.Nonnull;
+
+/**
+ * Represents a service endpoint URL and name.
+ *
+ * <p>If a client is constructed using this type, system defaults will be used for socket factory
+ * and trust manager.
+ */
+public class PlainServiceEndpoint extends DefaultServiceEndpoint {
```
ajordens, line 29: Not sure about `Plain` but otherwise this seems reasonable to me.

Maybe Simple or Basic?

jonsie (approved):

kork-web/src/main/java/com/netflix/spinnaker/config/PlainServiceEndpoint.java
srekapalli: I ended up modifying to be another attribute on the service endpoint instead of a separate data type.

PR #819, opened 2020-11-10
fix(misc): Skip header check if the endpoint is not matching the prov…

srekapalli:
Skip header check if the endpoint is not matching the supplied pattern.


kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java
```diff
@@ -27,12 +29,16 @@
...
+  private final Pattern headerCheckPattern;
 
-  MetricsInterceptor(Provider<Registry> registry, boolean skipHeaderCheck) {
+  MetricsInterceptor(
+      Provider<Registry> registry,
+      OkHttpMetricsInterceptorProperties okHttpMetricsInterceptorProperties) {
     this.registry = registry;
-    this.skipHeaderCheck = skipHeaderCheck;
+    this.okHttpMetricsInterceptorProperties = okHttpMetricsInterceptorProperties;
+    this.headerCheckPattern = okHttpMetricsInterceptorProperties.getHeaderCheckRegExPattern();
```
jonsie, line 41: Nit:  This tripped me up.  Instead of creating a member variable `headerCheckPattern` and assigning the value from `okHttpMetricsInterceptorProperties` I think it's fine to just call `okHttpMetricsInterceptorProperties.getHeaderCheckRegExPattern()` where it used below.

srekapalli: was keeping that in a member variable to avoid compiling the regex pattern every time we intercept the request.

ajordens: I think the properties class could keep track of the compiled pattern.

kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttpMetricsInterceptorProperties.java
```diff
@@ -15,18 +15,42 @@
...
 public class OkHttpMetricsInterceptorProperties {
 
+  /** If set to true, will skip header check completely. */
   public boolean skipHeaderCheck = false;
 
+  /**
+   * If Skip header check is set to false and regex is set, header check will be skipped for all
+   * endpoints except for the ones that match the provided regex.
+   */
+  public String headerCheckRegEx = null;
```
ajordens, line 32: `String headerCheckRegex` or `String headerCheckPattern` imo (looking at ~/.spinnaker/* we have more instances of "*Pattern" for regexes).

Not sure why we have public instance variables and getters/setters?

srekapalli: yh.. Didn't notice those are public vars. Seems like a left over from groovy to java. Fixed now.

jonsie (approved):

PR #820, opened 2020-11-11
test(sql/postgres): bump testcontainers, truncate cascade for PostgreSQL

ncknt:
- Bump testcontainers to fix builds on MacOS (https://github.com/testcontainers/testcontainers-java/pull/3159)
- In Clouddriver, we (will) have foreign keys to deal with when and need to cascade.


ncknt:
@cfieber or @asher Could you PTAL?

cfieber (approved):

PR #822, opened 2020-12-03
Batch truncate commands in cleanupDb

robfletcher:
Keel recently started using foreign keys in its database and I've noticed intermittent CI failures caused by attempts to truncate tables that have foreign key references. Although `SqlTestUtil.cleanupDb` disables foreign key checks I think what we're seeing is Gradle running tests in `keel-sql` and `keel-web` concurrently. Both use testcontainers and will clean up after tests with `SqlTestUtil.cleanupDb`. I think what's happening is that one test finishes cleaning up while another has just started and re-enables foreign key checks. This PR tries to correct this by batching all the truncate commands and the commands that disable/re-enable foreign key checks. That way they should all execute atomically.


luispollo (approved):
LGTM! Thanks for catching and fixing this!

jonsie (approved):

robzienert (approved):

PR #826, opened 2020-12-07
fix(logging): only log a warning when the generic handler handles an …

dbyron0:
…exception

The other @ExceptionHandler methods call storeException too, so move the warning to
@ExceptionHandler(Exception.class)


robzienert (approved):
Much better 👍

PR #827, opened 2020-12-08
chore(swagger): Upgrade SpringFox and Swagger

luispollo:
Upgrades SpringFox, which brings with it a newer/better version of Swagger. Among other things, this fixes the forever-annoying issue with the Swagger UI not properly auto-scrolling and auto-expanding a linked API section.

I haven't been able to test this yet, hence the draft. Something in the updated dependency tree broke loading of one of keel's plugins. 😞   If anyone has some spare cycles and could try this with another service, it would be greatly appreciated.

Heads-up: the new version changes the Swagger UI path from `/swagger-ui.html` to `/swagger-ui/index.html`. There's no way to customize that as far as I've seen. Not sure how big of a deal that would be.

### DIFF
--- a/kork-swagger/src/main/java/com/netflix/spinnaker/config/SwaggerConfig.java
+++ b/kork-swagger/src/main/java/com/netflix/spinnaker/config/SwaggerConfig.java
@@ -16,13 +16,10 @@
 
 package com.netflix.spinnaker.config;
 
-import static com.google.common.base.Predicates.or;
-
-import com.google.common.base.Predicate;
 import com.google.common.collect.ImmutableList;
 import java.util.List;
 import java.util.Objects;
-import java.util.stream.Collectors;
+import java.util.function.Predicate;
 import javax.annotation.Nullable;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.ConfigurationProperties;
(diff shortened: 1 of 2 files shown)


kork-swagger/src/main/java/com/netflix/spinnaker/config/SwaggerConfig.java
```diff
@@ -133,13 +129,13 @@ private BasePathProvider(String basePath, String documentationPath) {
     }
 
     @Override
-    protected String applicationPath() {
```
luispollo, line 136: They deprecated `AbstractPathProvider` in this version, and the docs are not great around path customization, so this is my current guess as to what needs to change, but needs testing.

kork-swagger/src/main/java/com/netflix/spinnaker/config/SwaggerConfig.java
```diff
@@ -32,11 +29,9 @@
 import springfox.documentation.builders.RequestHandlerSelectors;
 import springfox.documentation.service.ApiInfo;
 import springfox.documentation.spi.DocumentationType;
-import springfox.documentation.spring.web.paths.AbstractPathProvider;
+import springfox.documentation.spring.web.paths.DefaultPathProvider;
 import springfox.documentation.spring.web.plugins.Docket;
-import springfox.documentation.swagger2.annotations.EnableSwagger2;
 
-@EnableSwagger2
```
luispollo, line 39: Following the (very light on detail) migration instructions here: https://github.com/springfox/springfox#migrating-from-existing-2x-version

spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -127,7 +127,7 @@ dependencies {
     api("io.mockk:mockk:1.10.0")
     api("io.springfox:springfox-swagger-ui:${versions.springfoxSwagger}")
     api("io.springfox:springfox-swagger2:${versions.springfoxSwagger}")
-    api("io.swagger:swagger-annotations:${versions.swagger}")
+    api("io.swagger.core.v3:swagger-annotations:${versions.swagger}")
```
luispollo, line 130: This may break services using annotations, I'm not sure yet. Second thing to verify. SpringFox uses different deps for v2 and v3: https://github.com/springfox/springfox/blob/9722b7e3a41208bc3e7c0409aa3e63fc27a66f52/gradle/dependencies.gradle#L61-L68

robzienert:
This will need to be tested inside of Gate, too.

PR #828, opened 2020-12-08
feat(credentials): create repository one type at a time

ncknt:
Allows for more control for Spring dependencies, also avoid double loading and support parallel loading.


german-muzquiz (commented):
Seems that generics definition in signature of other methods like `registerCredentialsLoader` and `registerCredentialsRepository` could be removed, since they are already included at class level.

german-muzquiz (approved):

PR #833, opened 2020-12-14
feat(plugins): Get plugin ID from proxied spinnaker extension point

jonsie:
So a service can have some understanding of plugin IDs associated with extension point instances, if need be.


robzienert (approved):
Neat. Good one.

PR #835, opened 2020-12-17
feat(plugins): Support application event listeners

robzienert:
Initial support for allowing plugins to listen on application events.

This comes from a community request to listen in on [BeforeInitialExecutionPersist](https://github.com/spinnaker/orca/blob/master/orca-core/src/main/java/com/netflix/spinnaker/orca/events/BeforeInitialExecutionPersist.java). As with all things, I had to create an adapter to convert from Spring types to non-Spring types.

I'll be making a followup PR to adapt Orca accordingly.


link108:
This is great! I really like how SpinnakerApplicationEvent is typed and not just a map 💯

link108 (approved):

PR #837, opened 2020-12-18
fix(retrofit): Explicitly mark BAD_REQUEST and NOT_FOUND errors as not retryable

dbyron0:
cc: @ajordens .. an attempt to address https://github.com/spinnaker/spinnaker/issues/6237

### DIFF
--- a/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerRetrofitErrorHandler.java
+++ b/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerRetrofitErrorHandler.java
@@ -55,8 +55,7 @@ public final class SpinnakerRetrofitErrorHandler implements ErrorHandler {
           return new NotFoundException(e).setRetryable(false);
         }
         SpinnakerHttpException retval = new SpinnakerHttpException(e);
-        if (HttpStatus.Series.resolve(e.getResponse().getStatus())
-            == HttpStatus.Series.CLIENT_ERROR) {
+        if (e.getResponse().getStatus() == HttpStatus.BAD_REQUEST.value()) {
           retval.setRetryable(false);
         }
         return retval;
--- a/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerRetrofitErrorHandler.java
+++ b/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerRetrofitErrorHandler.java
@@ -52,9 +52,14 @@ public final class SpinnakerRetrofitErrorHandler implements ErrorHandler {
     switch (e.getKind()) {
       case HTTP:
         if (e.getResponse().getStatus() == HttpStatus.NOT_FOUND.value()) {
-          return new NotFoundException(e);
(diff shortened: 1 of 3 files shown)


kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerRetrofitErrorHandler.java
```diff
@@ -52,9 +52,14 @@ public Throwable handleError(RetrofitError e) {
     switch (e.getKind()) {
       case HTTP:
         if (e.getResponse().getStatus() == HttpStatus.NOT_FOUND.value()) {
-          return new NotFoundException(e);
+          return new NotFoundException(e).setRetryable(false);
```
dbyron0, line 55: Better to change NotFoundException to always set retryable to false?

link108 (approved):
LGTM

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerRetrofitErrorHandler.java
```diff
@@ -52,9 +52,14 @@ public Throwable handleError(RetrofitError e) {
...
         }
-        return new SpinnakerHttpException(e);
+        SpinnakerHttpException retval = new SpinnakerHttpException(e);
+        if (HttpStatus.Series.resolve(e.getResponse().getStatus())
+            == HttpStatus.Series.CLIENT_ERROR) {
+          retval.setRetryable(false);
```
ajordens, line 60: How much testing has been done with this outside of `kork`?  At a glance `Series.CLIENT_ERROR` is any 4xx, including 429 which could be retryable.

This does seem like a long drawn out fix for something that could have just been done in rosco? Sorry the holiday period made the cycle time on this quite long.

```
Response response =
        retrySupport.retry(
            () ->
                AuthenticatedRequest.allowAnonymous(
                    () -> clouddriverService.fetchArtifact(artifact)),
            5,
            1000,
            true);
```

Why can't we just wrap `clouddriverService.fetchArtifact(artifact))` in a try/catch that checks the status code?

dbyron0: I haven't tested this outside of kork, and you're right about not handling 429 properly.

You're also right that adding the try/catch block to rosco is the most direct/simplest way forward.  It does seem like the kind of thing to handle centrally though.  I haven't come across other places that would benefit from it yet, but I bet they exist.

What do you think of keeping the change here, but limiting it to 400/bad request?

dbyron0: Like this: https://github.com/spinnaker/kork/pull/837/commits/11a5ff9363824d427ff99efc9ab68f0d76815f80

ajordens: That looks more reasonable to me. I'll set a reminder to take one last look before merging tomorrow.

PR #839, opened 2021-01-07
feat(web): General purpose 'User-Agent'-injecting okhttp3 interceptor

ajordens:


ajordens:
Turns out I don't actually need this.

PR #848, opened 2021-02-23
chore(spring): Fixes error when loading empty properties

rvazquezglez:
chore(spring): Fixes error when loading empty properties

While working on [upgrading spring version](https://github.com/spinnaker/kork/pull/847/files) found a bug where the properties being loaded can be empty. This causes an exception when mapping the properties to a JSON object.

image

```
Failed reading extension config: Could not map provided config to expected shape
com.netflix.spinnaker.kork.exceptions.IntegrationException: Failed reading extension config: Could not map provided config to expected shape
	at com.netflix.spinnaker.kork.plugins.config.SpringEnvironmentConfigResolver.resolveInternal(SpringEnvironmentConfigResolver.kt:108)
	at com.netflix.spinnaker.kork.plugins.config.SpringEnvironmentConfigResolver.resolve(SpringEnvironmentConfigResolver.kt:69)
	at com.netflix.spinnaker.kork.plugins.config.SpringEnvironmentConfigResolverTest$tests$1$7.invoke(SpringEnvironmentConfigResolverTest.kt:130)
	at com.netflix.spinnaker.kork.plugins.config.SpringEnvironmentConfigResolverTest$tests$1$7.invoke(SpringEnvironmentConfigResolverTest.kt:38)
```

The root cause for this might be in `org.springframework.core.env.PropertySource#getProperty` from the newer Spring version, however, it's easily reproducible on current version unit tests by adding an empty map to one of the properties (without upgrading Spring).

### DIFF
--- a/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/config/SpringEnvironmentConfigResolver.kt
+++ b/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/config/SpringEnvironmentConfigResolver.kt
@@ -129,6 +129,10 @@ class SpringEnvironmentConfigResolver(
     propertyNames
       .filter { it.startsWith("spinnaker.extensibility") }
       .map { it to getProperty(it) }
+      .filter {
+        val second = it.second
+        if (second is Map<*, *>) second.isNotEmpty() else true
+      }
       .toMap()
 
   private inner class SystemConfigException(message: String) : SystemException(message)
--- a/kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/config/SpringEnvironmentConfigResolverTest.kt
+++ b/kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/config/SpringEnvironmentConfigResolverTest.kt
@@ -178,6 +178,7 @@ class SpringEnvironmentConfigResolverTest : JUnit5Minutests {
     "spinnaker.extensibility.plugins.netflix.very-important.extensions.orca.stage.config.optional" to "some new value",
     "spinnaker.extensibility.extensions.netflix.bar.config.somelist[0].hello" to "one",
     "spinnaker.extensibility.extensions.netflix.bar.config.somelist[1].hello" to "two",
+    "spinnaker.extensibility.repositories" to mapOf<String, String>(),
     "spinnaker.extensibility.repositories.foo.url" to "http://localhost:9000",
     "spinnaker.extensibility.plugins.netflix.sweet-plugin.config.somestring" to "overridden default"
   )


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/config/SpringEnvironmentConfigResolver.kt
```diff
@@ -129,6 +129,10 @@ class SpringEnvironmentConfigResolver(
     propertyNames
       .filter { it.startsWith("spinnaker.extensibility") }
       .map { it to getProperty(it) }
+      .filter {
```
zachsmith1, line 132: Would it make more sense to filter out empty maps before the `map { it to getProperty(it) }` function here? Otherwise this LGTM!

zachsmith1 (commented):
Added a note but otherwise LGTM

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/config/SpringEnvironmentConfigResolver.kt
rvazquezglez: The value is set by the `getProperty(it)` invocation, it's not possible to know if it will be empty before calling it.

zachsmith1 (approved):
LGTM

zachsmith1 (approved):

zachsmith1 (approved):

kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/config/SpringEnvironmentConfigResolverTest.kt
```diff
@@ -178,6 +178,7 @@ class SpringEnvironmentConfigResolverTest : JUnit5Minutests {
     "spinnaker.extensibility.plugins.netflix.very-important.extensions.orca.stage.config.optional" to "some new value",
     "spinnaker.extensibility.extensions.netflix.bar.config.somelist[0].hello" to "one",
     "spinnaker.extensibility.extensions.netflix.bar.config.somelist[1].hello" to "two",
+    "spinnaker.extensibility.repositories" to mapOf<String, String>(),
     "spinnaker.extensibility.repositories.foo.url" to "http://localhost:9000",
```
german-muzquiz, line 182: Is there an example of a yaml file that causes the error you're seeing? According to this test, the equivalent yaml file would be something like this:

```
spinnaker:
  extensibility:
    repositories: {}
    repositories:
      foo: 
        url: http://localhost:9000
```

Which is not valid yaml because there are two different values for the key `repositories`. I think you would need to add a new test with a different `properties` object to exercise the error.

PR #851, opened 2021-03-10
feat(metrics): Add Criticality annotation and use in MetricsInterceptor

jonsie:
When we add this annotation to controller methods we will be able to set the criticality of the controller method - this is useful to better setup alerts on failures of only controller methods tagged with a criticality of "high".


ajordens (approved):
This was basically what I was thinking, thanks for putting it together.

PR #853, opened 2021-03-18
feat(plugins): Add config property to enable/disable extension proxying

jonsie:
Configure via:

```
spinnaker:
  extensibility:
    proxyExtensions: true | false
```


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/v2/SpinnakerPluginService.kt
```diff
@@ -46,13 +46,15 @@ import java.util.function.Supplier
  * [SpinnakerUpdateManager] as the primary touch points for the plugin framework, decoupling
  * Spinnaker-specific plugin framework logic from PF4J wherever possible.
  */
+@Suppress("LongParameterList")
```
luispollo, line 49: 😆

luispollo (approved):
Thank you for adding this. Looks good to me!

kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/v2/SpinnakerPluginService.kt
jonsie: Not fixing that right now.  Heh.

PR #858, opened 2021-03-19
feat(x509/crl): reimplement the ability to revoke certificates

dmart:
This change effectively reverts https://github.com/spinnaker/kork/pull/684
as the removal of this functionality could have disastrous consequences
for operators.  The shameful choice in naming has been addressed.

\#684 was merged without any type of deprecation warning and there
doesn't seem to be an alternative approach for operators relying on the
ability to revoke certificates.

This reverts commit c6b94918d6fd26d1c84af3a4cf53f8cf2f5082e9.

### DIFF
--- a/kork-tomcat/kork-tomcat.gradle
+++ b/kork-tomcat/kork-tomcat.gradle
@@ -5,5 +5,8 @@ dependencies {
 
   api(platform(project(":spinnaker-dependencies")))
   implementation("org.springframework.boot:spring-boot-starter-web")
+  implementation("com.netflix.spectator:spectator-api")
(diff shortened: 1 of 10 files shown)


jonsie (approved):
I'm in favor of bringing this back, thanks for updating the terminology.

ajordens:
My take on naming is to align on [deny|allow] list but thanks for the attention on that. 

Tagged @cfieber as the original reverter almost a year ago.

sathishavm:
Can we merge this, We need this functionality to revoke our old certificates.

jvz (approved):

dbyron-sf (approved):

PR #860, opened 2021-04-08
fix(plugins): fixes enable-default-repositories flag

danielpeach:
The `spinnaker.extensibility.enable-default-repositories` flag didn't work, which was breaking users who have air-gapped environments and needed to turn the default repos off.

I tried to unify the configuration to be more consistent by just mapping `PluginsConfigurationProperties` with a `Binder` and removed a spot where we were manually mapping a single property (`pluginsRootPath`). I'm not sure why the repositories are handled differently: https://github.com/spinnaker/kork/blob/master/kork-plugins/src/main/java/com/netflix/spinnaker/config/PluginsAutoConfiguration.java#L136-L142. I can clean that up too if that makes sense.

I initially used the `ConfigResolver` instead of a `Binder`, but decided against it since it doesn't map properties in a way that people will expect. It doesn't do the snake case mapping (eg `enable-default-properties` doesn't work).

### DIFF
--- a/kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/PluginsConfigurationPropertiesTest.kt
+++ b/kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/PluginsConfigurationPropertiesTest.kt
@@ -1,0 +1,126 @@
+/*
+ * Copyright 2021 Armory, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.kork.plugins
+
+import com.netflix.spinnaker.config.PluginsAutoConfiguration
+import com.netflix.spinnaker.config.PluginsConfigurationProperties
+import com.netflix.spinnaker.kork.plugins.update.SpinnakerUpdateManager
+import dev.minutest.junit.JUnit5Minutests
+import dev.minutest.rootContext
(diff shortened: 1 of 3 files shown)


kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/PluginsConfigurationPropertiesTest.kt
```diff
@@ -0,0 +1,126 @@
...
+          val updateManager = ctx.getBean(SpinnakerUpdateManager::class.java)
+          expectThat(updateManager.repositories.map { it.id }).contains(
+            PluginsConfigurationProperties.SPINNAKER_OFFICIAL_REPOSITORY,
+            PluginsConfigurationProperties.SPINNAKER_COMMUNITY_REPOSITORY,
+          )
+        }
+      }
+    }
+
+    derivedContext<ApplicationContextRunner>("when default plugin repositories are disabled") {
```
danielpeach, line 58: This test fails without the other changes in this PR.

kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/PluginsConfigurationPropertiesTest.kt
```diff
@@ -0,0 +1,126 @@
...
+          val updateManager = ctx.getBean(SpinnakerUpdateManager::class.java)
+          expectThat(updateManager.repositories.map { it.id }).doesNotContain(
+            PluginsConfigurationProperties.SPINNAKER_OFFICIAL_REPOSITORY,
+            PluginsConfigurationProperties.SPINNAKER_COMMUNITY_REPOSITORY,
+          )
+        }
+      }
+    }
+
+    derivedContext<ApplicationContextRunner>("when no plugins root path is configured") {
```
danielpeach, line 83: The next two tests pass with or without the other changes in this PR.

cfieber (approved):

jonsie (approved):
Thanks for cleaning this up.

robzienert (approved):

darkrasid:
@danielpeach 

Is there any plan to apply kork 7.109.0 to spinnaker 1.25.x ?

I have this problem and I'm not able to apply 1.26.x because of k8s version dependency.

ishallbethat:
How to use the enable-default-repositories this flag ?

PR #864, opened 2021-05-07
fix(plugins): Use getExtensionClass() instead of getClass() for Expre…

mtweten:
…ssionFunctionProviders

ExpressionFunctionProvider extends SpinnakerExtensionPoint, enabling SpEL expressions to be contributed by plugins. However, when a function is being registered, the class of the ExpressionFunctionProvider is used - for ExpressionFunctionProviders coming from a plugin, this is a proxy class. In order for this to work correctly, the getExtensionClass() method should be used instead to resolve the appropriate class (proxy class or not).


cbfshr (approved):

st8f:
Hi! Got the same problem with expression extension. This fix will help a lot. Let's merge it already, can we?

st8f:
Checks are green! :tada:

dbyron-sf (approved):
Approving since @ajordens marked it ready to merge...

PR #866, opened 2021-05-27
fix(plugins): deck plugin versions should not be compared to the down…

danielpeach:
…loading service's version

### DIFF
--- a/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/SpinnakerUpdateManager.kt
+++ b/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/SpinnakerUpdateManager.kt
@@ -123,20 +123,15 @@ class SpinnakerUpdateManager(
       return null
     }
 
-    val systemVersion = pluginManager.systemVersion
-    val versionManager = SpinnakerServiceVersionManager(serviceName)
+    val versionManager = pluginManager.versionManager
     val lastPluginRelease: MutableMap<String, PluginRelease> = mutableMapOf()
 
-    // If service version cannot be determined(i.e 0.0.0),
-    // just check if the release requirement mentions the service to see if the plugin is applicable. This case should
-    // never happen, we should always know the system version but if it happens we can at least give a matching plugin
-    // to the service name in question.
-    //
-    // If version is determined, check the version constraint explicitly to decide if the plugin is applicable or not.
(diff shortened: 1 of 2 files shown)


kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/update/SpinnakerUpdateManager.kt
```diff
@@ -123,20 +123,15 @@ class SpinnakerUpdateManager(
...
-    // to the service name in question.
-    //
-    // If version is determined, check the version constraint explicitly to decide if the plugin is applicable or not.
+    // We're loading a plugin for a different service and we don't know the other service's version
+    // (e.g., Gate doesn't know Deck's version).
+    // We check if the release requirement mentions the service to see if the plugin is applicable,
+    // then pick the latest version.
     for (release in pluginInfo.releases) {
-      if ((systemVersion == ServiceVersion.DEFAULT_VERSION && release.requires.contains(serviceName, ignoreCase = true)) ||
-        versionManager.checkVersionConstraint(systemVersion, release.requires)
```
danielpeach, line 138: `systemVersion` here was Gate's version. 

`release.requires` was a string like `deck>=x.y.z`

kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/update/SpinnakerUpdateManagerTest.kt
```diff
@@ -138,13 +141,46 @@ class SpinnakerUpdateManagerTest : JUnit5Minutests {
...
+        provider = "Spinnaker"
+        releases = listOf(
+          SpinnakerPluginInfo.SpinnakerPluginRelease(false).apply {
+            requires = "deck>=0.0.2"
+            version = "0.0.1"
+            date = Date.from(Instant.now())
+            url = "/path/to/plugin"
+          },
+          SpinnakerPluginInfo.SpinnakerPluginRelease(false).apply {
+            requires = "deck>=0.0.2"
```
danielpeach, line 159: This test fails without the other changes in this PR. The plugin wants deck >= 0.0.2, but the service it's being loaded from is only 0.0.1. Since the update manager was comparing the service version to the deck constraint, it thought that there weren't any compatible plugins.

link108 (approved):

link108:
Thank you for fixing this! 🙏

PR #868, opened 2021-07-10
Support additional partitions

dbyron-sf:


kork-aws/src/main/java/com/netflix/spinnaker/kork/aws/ARN.java
```diff
@@ -25,7 +25,20 @@
...
+  // or more specifically:
+  //
+  // $ cat ./aws-java-sdk-core/src/main/resources/com/amazonaws/partitions/endpoints.json | jq -r
+  // '.partitions[] | .partition'
+  // aws
+  // aws-cn
+  // aws-us-gov
+  // aws-iso
+  // aws-iso-b
+  static final Pattern PATTERN =
```
xibz, line 40: Why do we need to parse the ARN? The AWS SDK for Java has `Arn.fromString(someArn)`. The Java SDK doesn't do as deep of a validation especially on partition, but I think that makes sense and we should loosen this a little bit and just the `fromString`.

xibz (changes requested):

kork-aws/src/main/java/com/netflix/spinnaker/kork/aws/ARN.java
dbyron-sf: Makes sense to me: https://github.com/spinnaker/kork/pull/868/commits/a4ea6472c0eabd4cff468df76e43af0d5e9ad5d0

xibz (approved):
Cool! I like these changes a lot more as this allows AWS to add more partitions and not have us constantly update this file.

dbyron-sf:
Yeah, was thinking the same.  Now I think we need an official approver to give a thumbs up.  Any suggestions?

allisaurus (approved):
Hooray use of `com.amazonaws.arn.Arn`! These changes LGTM re: arn/partition format and testing, but I haven't messed with KORK myself aside from the occasional rote AWS SDK version bump, so would like like to deter final approval to someone who's been more active here...

@danielpeach, do you see any reason these changes might not be good to go?

link108 (approved):

dbyron-sf:
Thanks for the review @allisaurus!

PR #872, opened 2021-07-21
fix(tomcat): upgrades tomcat to 9.0.48

xibz:
This upgrades tomcat from 9.0.31, which enabled header value validation,
to 9.0.48. The header value validation introduced 3 regressions that
9.0.48 fixes. Upgrading to 9.0.48 also has fixes for the CVEs listed
below

See links for more details:
https://bz.apache.org/bugzilla/show_bug.cgi?id=65272
https://bz.apache.org/bugzilla/show_bug.cgi?id=64210
https://bz.apache.org/bugzilla/show_bug.cgi?id=64192

Associated CVEs:
CVE-2020-1935
CVE-2021-33037

Also pretty sure this [issue](https://github.com/spinnaker/spinnaker/issues/4684) was caused by this.

Signed-off-by: benjamin_j_powell <benjamin_j_powell@apple.com>

### DIFF
--- a/kork-tomcat/src/test/java/com/netflix/spinnaker/kork/tomcat/CRLFHeaderTest.java
+++ b/kork-tomcat/src/test/java/com/netflix/spinnaker/kork/tomcat/CRLFHeaderTest.java
@@ -1,0 +1,106 @@
+package com.netflix.spinnaker.kork.tomcat;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.OutputStream;
+import java.net.Socket;
+import java.nio.charset.StandardCharsets;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.web.server.LocalServerPort;
+import org.springframework.context.annotation.Bean;
+import org.springframework.test.context.TestPropertySource;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@SpringBootTest(
+    webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
+    classes = {CRLFHeaderTest.HeaderTestControllerConfiguration.class})
+@TestPropertySource(
+    properties = {"logging.level.org.apache.coyote.http11.Http11InputBuffer = DEBUG"})
+@org.springframework.boot.context.properties.EnableConfigurationProperties({
+  TomcatConfigurationProperties.class
+})
+public class CRLFHeaderTest {
+  private static final String CR = "\r";
+  private static final String LF = "\n";
+
+  private static final String HOST = "127.0.0.1";
+  public static final String HTTP_1_1_PREFIX = "HTTP/1.1 ";
+
+  @SpringBootApplication
+  public static class HeaderTestControllerConfiguration {
+    @Bean
+    HeaderTestController headerTestController() {
+      return new HeaderTestController();
+    }
+  }
+
+  @RestController
+  @RequestMapping("/header/test")
+  static class HeaderTestController {
+    @GetMapping
+    public void onGetRequest() {}
+  }
+
+  @LocalServerPort private int port;
+
+  /*
+   * clientTest tests the TCP segmentation bug introduced in tomcat v9.0.31.
+   * See https://bz.apache.org/bugzilla/show_bug.cgi?id=64210
+   *
+   * We reproduce this test by forcing a TCP segmentation in between a header that ends with a CR
+   * and begins the next header with a LF. This causes tomcat to throw an exception and results in a
+   * 400 status code due to invalid header parsing
+   */
+  @Test
+  public void clientTest() throws IOException {
+    Socket socket = new Socket("127.0.0.1", port);
+    try {
+      // set the buffer size to 1 to force a TCP segment at 1 byte
+      socket.setSendBufferSize(1);
+      // set TCP_NODELAY to true which prevents tcp from buffering requests
+      socket.setTcpNoDelay(true);
+      InputStream in = socket.getInputStream();
+      OutputStream out = socket.getOutputStream();
+
+      String requestAndHeaders =
+          "GET /header/test HTTP/1.1"
+              + CR
+              + LF
+              + "Host: "
+              + HOST
+              + ":"
+              + port
+              + CR // Writes the CR to the socket and is received by the server
+              + LF // This LF causes a 400
+              + "User-Agent: spinnaker-test/1.0"
+              + CR
+              + LF
+              + CR
+              + LF;
+
+      // write the first header
+      out.write(requestAndHeaders.getBytes(StandardCharsets.UTF_8));
+      // force a write
+      out.flush();
+      BufferedReader lineReader = new BufferedReader(new InputStreamReader(in));
+
+      String line = lineReader.readLine();
+      while (line != null) {
+        if (line.startsWith(HTTP_1_1_PREFIX)) {
+          int status = Integer.parseInt(line.substring(HTTP_1_1_PREFIX.length()).trim());
+          Assertions.assertEquals(2, status / 100, "non-2xx status code returned: " + status);
+        }
+        line = lineReader.readLine();
+      }
+    } finally {
+      socket.close();
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


dbyron-sf:
Do you mind holding off on this until we sort out the other build stuff resulting in jcenter's disappearance?

xibz:
@dbyron-sf no rush on my end. Just ping this issue and I'll rebase.

mattgogerly:
I'm curious how this potentially fixes #4684 - we've run into this a couple of times and so we've just set the max header size to something ridiculous as a workaround.

Maybe I'm missing something obvious, but I can't see anything in the linked threads about header size?

spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -27,7 +27,12 @@ ext {
     springBoot       : "2.2.5.RELEASE",
     springCloud      : "Hoxton.SR4",
     springfoxSwagger : "2.9.2",
-    swagger          : "1.5.20" //this should stay in sync with what springfoxSwagger expects
+    swagger          : "1.5.20", //this should stay in sync with what springfoxSwagger expects
+    // tomcat (v9.0.31) was included via spring boot and upgrading from 9.0.31
+    // to 9.0.48 makes a few breaking changes, requires adding new
```
dbyron-sf, line 32: Not quite sure I understand this comment.  I read it as saying 9.0.48 makes a few breaking changes and requires changing code, as if to say we don't actually want 9.0.48.  But that's what we're getting here...and I don't see any code changes.

xibz: Ah good point. I meant that `9.0.31` to `9.0.46` makes breaking changes! Will update!

xibz:
@mattgogerly Yea, so I did A LOT of digging into this. So sometime's you'll see a `header is too large` exception but a lot of the time it'll be a 4xx with an something stating that it doesn't conform to the RFC. But it happens REALLY infrequently because you need a TCP packet to segment between a `CR` and `LF`. Meaning spinnaker will randomly encounter 400s.

spinnaker-dependencies/spinnaker-dependencies.gradle
xibz: @dbyron-sf Alright went ahead and fixed that. Was suppose to be in relation to spring boot's versions that makes breaking changes and thus making it harder to upgrade to. Hope the new commit clears that up!

dbyron-sf:
> @mattgogerly Yea, so I did A LOT of digging into this. So sometime's you'll see a `header is too large` exception if the header is actually too large, but another case is when 4xx with an exception something stating that it doesn't conform to the RFC. But it happens REALLY infrequently because you need a TCP packet to segment between a `CR` and `LF` in a header. Meaning spinnaker will randomly encounter 400s. So wasn't sure if that issue was being caused by headers being too large or the random tcp segment bug

@xibz We've been bitten by this too.  Would you mind leading me by the hand a bit and explaining how 9.0.48 resolves these problems?

dbyron-sf:
Sorry, I'm reading the bugs you linked and they fill in the details nicely.

spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -27,7 +27,13 @@ ext {
     springBoot       : "2.2.5.RELEASE",
     springCloud      : "Hoxton.SR4",
     springfoxSwagger : "2.9.2",
-    swagger          : "1.5.20" //this should stay in sync with what springfoxSwagger expects
+    swagger          : "1.5.20", //this should stay in sync with what springfoxSwagger expects
+    // tomcat (v9.0.31) was included via spring boot and upgrading spring boot
+    // from 2.2.5.RELEASE (tomcat v9.0.31 used here) to 2.5.0 (tomcat v9.0.48)
+    // makes a few breaking changes, requires adding new dependencies, and
+    // changing code. So instead updating tomcat to v9.0.48 as a around until
```
dbyron-sf, line 34:
```suggestion
    // changing code. So instead updating tomcat to v9.0.48 as a workaround until
```

xibz:
@dbyron0 - Yea, np. I forgot to actually push up the test that actually repros this issue. But yea, reading through the bugs should highlight why this happens and why we get random 4xx

spinnaker-dependencies/spinnaker-dependencies.gradle
dbyron-sf: Thanks!

kork-tomcat/src/test/java/com/netflix/spinnaker/kork/tomcat/CRLFHeaderTest.java
```diff
@@ -0,0 +1,108 @@
...
+  @Test
+  public void clientTest() throws IOException {
+    Socket socket = new Socket("127.0.0.1", port);
+    try {
+      socket.setSendBufferSize(1);
+      socket.setTcpNoDelay(true);
+      InputStream in = socket.getInputStream();
+      OutputStream out = socket.getOutputStream();
+
+      String requestAndFirstHeaders =
```
dbyron-sf, line 62: Thanks for adding a test!  I'm going to try this a bit on my local machine, but I get the feeling there are some subtle details here that make it fail with 9.0.31 but pass with 9.0.48.  If that's true, some comments would help.

xibz: Yea, definitely. I'll go ahead and add some comments to illustrate why this fails on v9.0.31

kork-tomcat/src/test/java/com/netflix/spinnaker/kork/tomcat/CRLFHeaderTest.java
```diff
@@ -0,0 +1,108 @@
...
+              + CR
+              + LF;
+
+      out.write(requestAndFirstHeaders.getBytes(StandardCharsets.UTF_8));
+      out.flush();
+      out.write(lastHeadersAndBlankLine.getBytes(StandardCharsets.UTF_8));
+      out.flush();
+
+      BufferedReader lineReader = new BufferedReader(new InputStreamReader(in));
+      StringBuffer output = new StringBuffer();
```
dbyron-sf, line 93: Am I right that this isn't really used?  I suppose it's helpful to have when running this in a debugger...Is there a way to print it if the test fails, or something?  And perhaps a nit to change the name, maybe to response?  There's already a variable named `out` ...

kork-tomcat/src/test/java/com/netflix/spinnaker/kork/tomcat/CRLFHeaderTest.java
dbyron-sf: Thank you!

kork-tomcat/src/test/java/com/netflix/spinnaker/kork/tomcat/CRLFHeaderTest.java
xibz: Went ahead and removed it. Was an artifact when I was trying to repro this issue.

kork-tomcat/src/test/java/com/netflix/spinnaker/kork/tomcat/CRLFHeaderTest.java
xibz: Added some comments. Let me know if that helps clear up what the test is doing!

kork-tomcat/src/test/java/com/netflix/spinnaker/kork/tomcat/CRLFHeaderTest.java
```diff
@@ -0,0 +1,106 @@
...
+   * We reproduce this test by forcing a TCP segmentation in between a header that ends with a CR
+   * and begins the next header with a LF. This causes tomcat to throw an exception and results in a
+   * 400 status code due to invalid header parsing
+   */
+  @Test
+  public void clientTest() throws IOException {
+    Socket socket = new Socket("127.0.0.1", port);
+    try {
+      // set the buffer size to 1 to force a TCP segment at 1 byte
+      socket.setSendBufferSize(1);
```
xibz, line 66: If you want this test to pass with 9.0.48 just set the buffer size to something large like 1000

dbyron-sf (approved):
LGTM, thanks for responding to all my questions so quickly!  This is a big enough change that I'm a little hesitant to merge it at a time that I hope is quite close to when we cut the 1.27 branches.  But, given that it fixes some CVEs, I'm ok to move forward with it (and potentially even backport it).  Would love a second opinion from e.g. @link108 though.

link108 (approved):

link108:
LGTM! I'm happy to pull this into the 1.27 release, we'll be able to backport any bug fixes that arise and patch the release from there

PR #873, opened 2021-07-21
chore(build): use dependencies that exist in maven central

dbyron-sf:
some dependencies (or transitive dependencies of those) only existed in bintray, not maven
central, and now that bintray is gone, we need to adjust.

### DIFF
--- a/spinnaker-dependencies/spinnaker-dependencies.gradle
+++ b/spinnaker-dependencies/spinnaker-dependencies.gradle
@@ -21,7 +21,7 @@ ext {
     retrofit         : "1.9.0",
     retrofit2        : "2.8.1",
     spectator        : "0.103.0",
-    spek             : "1.2.1",
+    spek             : "1.1.5",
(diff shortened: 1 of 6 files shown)


spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -21,7 +21,7 @@ ext {
     retrofit         : "1.9.0",
     retrofit2        : "2.8.1",
     spectator        : "0.103.0",
-    spek             : "1.2.1",
+    spek             : "1.1.5",
```
link108, line 24: was downgrading spek intentional?

dbyron-sf: It was.  1.1.5 is the last one available at maven central. See e.g. https://repo.maven.apache.org/maven2/org/jetbrains/spek/spek-junit-platform-engine/

link108 (approved):

PR #880, opened 2021-08-06
feat(build): bump dependencies for the given branch

link108:
Related to https://github.com/spinnaker/bumpdeps/pull/17
Pass the branch to bump dependencies so we can bump dependencies for a given branch

### DIFF
--- a/.github/workflows/bump_dependencies.yml
+++ b/.github/workflows/bump_dependencies.yml
@@ -11,7 +11,7 @@ jobs:
     - uses: spinnaker/bumpdeps@master
       with:
         ref: ${{ github.event.client_payload.ref }}
-        branch: ${{ github.event.client_payload.branch }}
+        baseBranch: ${{ github.event.client_payload.branch }}
         key: korkVersion
         repositories: clouddriver,echo,fiat,front50,gate,halyard,igor,keel,orca,rosco,swabbie
         mavenRepositoryUrl: https://repo.maven.apache.org/maven2
--- a/.github/workflows/bump_dependencies.yml
+++ b/.github/workflows/bump_dependencies.yml
@@ -11,6 +11,7 @@ jobs:
     - uses: spinnaker/bumpdeps@master
       with:
         ref: ${{ github.event.client_payload.ref }}
+        branch: ${{ github.event.client_payload.branch }}
         key: korkVersion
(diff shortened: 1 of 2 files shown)


.github/workflows/bump_dependencies.yml
```diff
@@ -11,6 +11,7 @@ jobs:
     - uses: spinnaker/bumpdeps@master
       with:
         ref: ${{ github.event.client_payload.ref }}
+        branch: ${{ github.event.client_payload.branch }}
```
dbyron-sf, line 14:
```suggestion
        baseBranch: ${{ github.event.client_payload.branch }}
```
?

.github/workflows/release.yml
```diff
@@ -15,6 +15,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - run: git fetch --prune --unshallow
+      - name: Get Branch
```
dbyron-sf, line 18: Perhaps a comment about why we're doing this / what it's doing...and perhaps some mention of what happens when there's more than one or no branches?

.github/workflows/release.yml
```diff
@@ -15,6 +15,16 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - run: git fetch --prune --unshallow
+      # Given a tag, determine what branch we are on
+      # Minor tags (eg v0.1.0) should be pushed to the master branch
```
dbyron-sf, line 19: I'm not sure this distinction between minor and patch is helpful here.  I mean, the code here doesn't have any logic to figure out minor/patch/etc.  It's figuring out the current branch (e.g. master or release-x) and targeting the autobump PR to that branch, assuming the branch naming convention is the same in the consumers of kork as it is in kork.

.github/workflows/release.yml
```diff
@@ -15,6 +15,16 @@ jobs:
...
+      # Patch tags (eg v0.1.1+) should be pushed to their respective release branches
+      # If the tag doesn't exist on a branch, bumpDeps will log and throw an error
+      # TODO: handle error case if tag exists on multiple branches
```
dbyron-sf, line 22: It'd be nice to handle the no-branch case here too, and potentially abort and not create the github release nor publish a jar in this case.

dbyron-sf (approved):

link108:
@mergifyio backport release-1.24.x

PR #886, opened 2021-08-13
chore(dependencies): Upgrade google-oauth-client to resolve CVE

j-sandy:
[CVE-2020-7692](https://nvd.nist.gov/vuln/detail/CVE-2020-7692)
Introduced by com.google.api-client:google-api-client in spinnaker-dependencies.
Upgrade com.google.guava:guava to support google-api-client upgrade.

### DIFF
--- a/spinnaker-dependencies/spinnaker-dependencies.gradle
+++ b/spinnaker-dependencies/spinnaker-dependencies.gradle
@@ -67,7 +67,7 @@ dependencies {
   constraints {
     api("cglib:cglib-nodep:3.3.0")
     api("com.amazonaws:aws-java-sdk:${versions.aws}")
-    api("com.google.api-client:google-api-client:1.28.0")
+    api("com.google.api-client:google-api-client:1.30.10")
     api("com.google.auth:google-auth-library-oauth2-http:0.18.0")
     api("com.google.apis:google-api-services-admin-directory:directory_v1-rev105-1.25.0")
     api("com.google.apis:google-api-services-cloudbuild:v1-rev836-1.25.0")
@@ -76,7 +76,7 @@ dependencies {
     api("com.google.apis:google-api-services-monitoring:v3-rev477-1.25.0")
     api("com.google.apis:google-api-services-storage:v1-rev141-1.25.0")
     api("com.google.code.findbugs:jsr305:3.0.2")
-    api("com.google.guava:guava:28.2-jre")
+    api("com.google.guava:guava:30.0-jre")
     // JinJava 2.5.3 has a bad bug: https://github.com/HubSpot/jinjava/issues/429
     api("com.hubspot.jinjava:jinjava:2.5.2")
     api("com.jakewharton.retrofit:retrofit1-okhttp3-client:1.1.0")


jkschneider (approved):
No significant breakages from Guava 28 to 30, so this should be safe.

j-sandy:
@jkschneider Pardon me re-requested by mistake. Please ignore it, if it can be ignored, or approve if its required to get this merged.

dbyron-sf:
https://nvd.nist.gov/vuln/detail/CVE-2020-7692 says it's been modified and now says:
```
This affects the package com.google.oauth-client:google-oauth-client before 1.31.0.
```
though it also says
```
It is awaiting reanalysis which may result in further changes to the information provided.
```
so perhaps it's better to wait until that's settled?

Note that when I bump `com.google.api-client:google-api-client` to 1.30.10, I see gradle bringing in version 29.0-android of `com.google.guava:guava`.  So I assume getting back to a -jre version of guava is reason for bumping its version.  Is that right?  Bumping guava to 29.0-jre is enough to do that.  Why the bigger bump to 30.0-jre?  I think it's to fix https://nvd.nist.gov/vuln/detail/CVE-2020-8908.  Is that right?

When I bump`com.google.api-client:google-api-client` to 1.31.0, I see gradle bringing in version 30.0-android of guava, so at that point bumping to 30.0-jre makes sense to me.  I could see opening a separate PR to bump guava to 30.0-jre to address https://nvd.nist.gov/vuln/detail/CVE-2020-8908.  With https://nvd.nist.gov/vuln/detail/CVE-2020-7692 still in flux, maybe that makes sense?

j-sandy:
google-oauth-client is transitive dependency for all google apis lib (google-api-services-storage, google-api-services-monitoring, etc.) constraint in spinnaker-dependencies. It is direct dependency for google-api-client. We had 2 options to fix the CVE in google-oauth-client 
(a) constrainting google-oauth-client:1.31.0 (being the nearest safe version) 
(b) upgrading google-api-client:1.30.10 (being nearest version having google-oauth-client:1.31.0). 
So, implemented option (b) as per existing design.

For guava API upgrade, I have raised separate PR [here](https://github.com/spinnaker/kork/pull/887), as recommended. However, I think we can opt for google-oauth-client:1.31.0 untill any further analysis being posted by NVD. If recommended, we can push TODO comment for future tracking.

dbyron-sf:
OK, the [guava PR](https://github.com/spinnaker/kork/pull/887) is in.  I agree, let's go with google-oauth-client:1.31.0 here.

I also agree that option (b) makes sense.  I'm still curious if your reasons for bumping guava here match what I wrote above.

j-sandy:
Initially, the decision to bump up guava was based on testing process, where we observed that few spinnaker components once constrained with google-oauth-client:1.31.0 required guava:30.0-jre for compilation, build and unit test.

j-sandy:
Based on discussions, raising an updated PR for google-oauth-client:1.31.0 with comment to track any updates by NVD for [CVE-2020-7692]( https://nvd.nist.gov/vuln/detail/CVE-2020-7692)

PR #893, opened 2021-09-02
feat(expressions): Allow TemporalAdjusters in expressions

gadams00:
When automatically updating base image once per week, we'd like to pin the version number to a particular day of the week: 

```java
LocalDate.now().with(TemporalAdjusters.previous(DayOfWeek.WEDNESDAY)).toString()
```


jkschneider (approved):
LGTM

PR #906, opened 2021-11-09
Spectator 1.0.6

dbyron-sf:


dbyron-sf:
May want to skip 1.0.5 and use a version with https://github.com/Netflix/spectator/pull/922 instead...

dbyron-sf:
> May want to skip 1.0.5 and use a version with [Netflix/spectator#922](https://github.com/Netflix/spectator/pull/922) instead...

And so now we're going straight to 1.0.6...

mattgogerly (approved):

PR #908, opened 2021-12-07
fix(plugins): update constructors for jackson mappers

dogonthehorizon:
When we moved to Jackson 2.12 there were some constructors that weren't
updated and weren't caught until runtime. I don't know enough about
Jackson's Kotlin integration to why this wasn't an issue. I built and
ran tests on master for this project and everything was fine, even after
making the constructor changes.

I tested this change by using composite builds to bring this into rosco
and see that it starts successfully now. I'm not sure if we'll need to
do similar kork bumps in other services.


link108 (approved):

link108:
@mergifyio backport release-1.27.x release-1.26.x

PR #912, opened 2021-12-14
fix(security): set platform restriction on log4j

jasonmcintosh:
Bumps log4j platform restriction.  Though the core lib is NOT included in dependencies, better safe than sorry...

### DIFF
--- a/spinnaker-dependencies/spinnaker-dependencies.gradle
+++ b/spinnaker-dependencies/spinnaker-dependencies.gradle
@@ -47,6 +47,11 @@ dependencies {
    * For example, `junit-bom` and `jackson-bom` will win out over the versions of JUnit and Jackson
    * specified by `spring-boot-dependencies`.
    */
+   // Log4shell safeguard.  Per analysis, log4j-core is not included in dependencies, but this would prevent transitive inclusion of it by extension
+   // platforms.   Doing 2.16.0 which completely removes message lookups AND sets jndi to disabled by default
+
+  api(platform("org.apache.logging.log4j:log4j-bom:2.16.0"))
+
   api(platform("org.jetbrains.kotlinx:kotlinx-coroutines-bom:1.4.3"))
   //kotlinVersion comes from gradle.properties since we have kotlin code in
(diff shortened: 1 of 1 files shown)


j-sandy:
This behavior is also disabled by default in 2.15.0 as mentioned [here](https://nvd.nist.gov/vuln/detail/CVE-2021-44228) . I hope we should have a valid reason to bump up for 2.16.0.

spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -47,6 +47,11 @@ dependencies {
    * For example, `junit-bom` and `jackson-bom` will win out over the versions of JUnit and Jackson
    * specified by `spring-boot-dependencies`.
    */
+   // Log4shell safeguard.  Per analysis, log4j-core is not included in dependencies, but this would prevent transitive inclusion of it by extension
+   // platforms.   Doing 2.16.0 which completely removes message lookups AND sets jndi to disabled by default
+
+  api(platform("org.apache.logging.log4j:log4j-bom:2.16.0"))
```
j-sandy, line 53: This behavior is also disabled by default in 2.15.0 as mentioned here . I hope we should have a valid reason to bump up for 2.16.0.

jasonmcintosh: 1) It's good to be on latest releases as much as possible or makes sense.  
2) 2.16 makes jdni lookups disabled by default entirely vs. just fixing the parsing in log messages.  As such it's MUCH more secure and recommended.

Either of the above reasons would be justification for bumping to 2.16.

dbyron-sf (approved):
Thanks for this @jasonmcintosh .  Planning to tag kork after this goes in.

spinnaker-dependencies/spinnaker-dependencies.gradle
jvz: There was another CVE published today (a medium one) for 2.15.0, so using 2.16.0 makes sense.

PR #918, opened 2021-12-15
fix(plugins): deck plugin versions should not be compared to the down…

dogonthehorizon:
…loading service's version (#866)


cristhian-castaneda (approved):
LGTM

PR #924, opened 2022-02-09
fix(dependencies): add com.amazonaws:aws-java-sdk-sts to kork-aws

chris-h-phillips:
Orca in particular (and other services likely) is unable to correctly assume
the IAM role associated with it's pod's service account when deployed to EKS.
It fails with a message like:

 Reason: Unable to load AWS credentials from any provider in the chain

This appears to be because orca is not pulling in aws-java-sdk-sts as a dependency
anywhere. Other services like front50 and clouddriver are pulling in this library
and are able to function correctly in this EKS / IRSA deployment scenario.

kork-aws seems to be a good place to put this dependency after discussion on Slack.

I noticed this when trying to configure orca using a secret in secretsmanager.

fixes: https://github.com/spinnaker/spinnaker/issues/6502

### DIFF
--- a/kork-aws/kork-aws.gradle
+++ b/kork-aws/kork-aws.gradle
@@ -37,6 +37,8 @@ dependencies {
     transitive = false
   }
 
+  runtimeOnly "com.amazonaws:aws-java-sdk-sts"
+
   testImplementation "org.junit.jupiter:junit-jupiter-api"
   testImplementation "org.junit.jupiter:junit-jupiter-params"
   testImplementation "org.junit.platform:junit-platform-runner"


dbyron-sf (approved):

sarab2009:
I have the same problem with echo to employ SQS services. It seems the echo pod on EKS ignores the IRSA.
`com.amazonaws.services.sqs.model.AmazonSQSException: Access to the resource https://sqs.eu-central-1.amazonaws.com/
is denied. (Service: AmazonSQS; Status Code: 403; Error Code: AccessDenied; Request ID: xxxxxxxxxxxxx; Proxy: null)`

Can this fix resolve this problem, as well?

PR #940, opened 2022-03-14
feat(sql): use sessionlevel locking for liquibase locks

apoorvmahajan:
## Purpose
This addresses https://github.com/spinnaker/spinnaker/issues/6625

## Background

* We observed that when clouddriver starts up, liquibase creates a table in the db and stores a lock, to ensure that only one of the pods can do the schema updates at any given point in time. If the pod is terminated before liquibase can release the lock, no new clouddriver pods can come up since in those pods, liquibase fails to acquire this lock as its held by a now dead pod. This lock has no timeout, so manual intervention is needed to clear this.

* Liquibase is aware of this issue, https://github.com/liquibase/liquibase/issues/1453 and has been working on it for a while. In the meantime, they propose using init containers https://www.liquibase.com/blog/using-liquibase-in-kubernetes  for running the liquibase logic or executing the `liquibase releaselocks` command to clear the locks manually, if needed.

* Based on this, we had three possible solutions that we could try:
1. Use the init container approach suggested by liquibase.
2. Have an external resource monitoring job, that checks to see if the pod holding the lock is alive or dead. If it is dead, it it will clear the lock.
3. Rework the locking logic in liquibase to not have transaction level locking, but instead have session level locking.
 

After discussing the options internally, it seems like the best way to solve this would be such that the system self-heals, and doesn't require any external solutions to fix the issue. With solution 1, it can so happen that for long running migrations, the init container can die before releasing the lock, so that doesn't completely make the problem go away. 

So we decided to go with option 3. We found an external library that already solves this problem for mysql db by implementing the session level locking provided by mysql. (https://github.com/blagerweij/liquibase-sessionlock) and have been using it for a while now without a repeat of this issue in our internal systems.

### DIFF
--- a/kork-sql/kork-sql.gradle
+++ b/kork-sql/kork-sql.gradle
@@ -12,6 +12,8 @@ dependencies {
   api "org.liquibase:liquibase-core"
   api "com.zaxxer:HikariCP"
 
+  implementation 'com.github.blagerweij:liquibase-sessionlock:1.4.0'
+
   testImplementation "org.springframework.boot:spring-boot-starter-actuator"
   testImplementation "org.springframework.boot:spring-boot-starter-web"
   testImplementation "org.springframework.boot:spring-boot-starter-test"
--- a/kork-sql/kork-sql.gradle
+++ b/kork-sql/kork-sql.gradle
@@ -12,7 +12,7 @@ dependencies {
   api "org.liquibase:liquibase-core"
   api "com.zaxxer:HikariCP"
 
-  implementation 'com.github.blagerweij:liquibase-sessionlock:1.4.0'
+  implementation 'com.github.blagerweij:liquibase-sessionlock:1.5.0'
 
   testImplementation "org.springframework.boot:spring-boot-starter-actuator"
   testImplementation "org.springframework.boot:spring-boot-starter-web"


apoorvmahajan:
## Test set up

* Run 2 clouddriver pods that use this library.
* Update the liquibase changeset in clouddriver to include the following sql statement: `SELECT SLEEP(120)` so that it would sleep for 2 mins after holding the lock.

 ## Expectations

* Only one of the 2 pods should acquire the lock before attempting to implement this changeset. The other pod should wait to acquire the lock in the meantime.
* On force killing the 1st pod midway through the changeset implementation (before the 2min sleep interval is up), we should see that it will not have released the lock. But even then, the 2nd pod should be able to acquire the lock and redo the changeset, since the first one did not complete the transaction.
* Then the 3rd pod (which came up in place of the forcefully killed 1st pod) should wait to acquire the lock that is now held by the 2nd pod.
* Eventually, the 2nd pod should complete the changeset implementation, release the lock and start clouddriver.
* Now the 3rd pod should acquire and release the lock and start clouddriver.

mattgogerly:
Looking at the commits since 1.4 it should probably be fine to go to 1.5, all that's changed is a few docs and a single error message for Postgre.

apoorvmahajan:
@mattgogerly I've bumped it up to 1.5.0.

german-muzquiz:
I'm curious if you need to configure something for using the library with liquibase < 4.x as indicated here: https://github.com/blagerweij/liquibase-sessionlock#usage

What version of liquibase were you using in your tests?

apoorvmahajan:
@juanpabloprado @german-muzquiz The version of liquibase that we use internally is 3.8.9 and it worked without making any further changes.

german-muzquiz (approved):

mattgogerly (approved):

mattgogerly:
@Mergifyio update

PR #949, opened 2022-04-20
chore(dependencies): Upgrade spring cloud dependencies to Hoxton.SR12

j-sandy:
With upgrade of spring cloud dependencies, transitive dependency of resilience4j bumped up to 1.7.0. Removed the constraints marked for resilience4j in spinnaker-dependencies.gradle.
Adding runtime dependency of hibernate-validator, which has been removed from resilience4j:1.7.0. It is required by front50, orca and echo.


mattgogerly (approved):

j-sandy:
As per spring boot 2.3 release [notes](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.3-Release-Notes#validation-starter-no-longer-included-in-web-starters), validation starter is no longer included in web starters. To provide validator for spinnaker components we can think of following approaches: 
1. To add spring-boot-starter-validation next to spring-boot-starter-web in kork-web. It bring extra dependencies along with validator dependencies.
2. To add hibernate-validator as runtimeOnly dependency to kork-web, because most of spinnaker services require reference implementation of hibernate-validator at runtime. Having explicit and minimal dependency would be good.

dbyron-sf (approved):

PR #967, opened 2022-06-23
feat(selectors): add account pattern and cloud provider selectors

mattgogerly:
Adds ServiceSelector implementations for selecting services by an account name pattern or cloud provider.

https://github.com/spinnaker/spinnaker/issues/6651

### DIFF
--- a/kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/ByAccountServiceSelector.java
+++ b/kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/ByAccountServiceSelector.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 Netflix, Inc.
+ * Copyright 2022 JPMorgan Chase & Co
  *
  * Licensed under the Apache License, Version 2.0 (the "License")
  * you may not use this file except in compliance with the License.
@@ -16,22 +16,18 @@
 
 package com.netflix.spinnaker.kork.web.selector;
 
-import java.util.HashSet;
 import java.util.Map;
-import java.util.Set;
+import java.util.regex.Pattern;
 
-public class ByExecutionTypeServiceSelector implements ServiceSelector {
+public class ByAccountServiceSelector implements ServiceSelector {
   private final Object service;
   private final int priority;
-  private final Set<String> executionTypes;
+  private final Pattern accountPattern;
 
-  @SuppressWarnings("unchecked")
-  public ByExecutionTypeServiceSelector(
-      Object service, Integer priority, Map<String, Object> config) {
+  public ByAccountServiceSelector(Object service, Integer priority, Map<String, Object> config) {
     this.service = service;
     this.priority = priority;
-    this.executionTypes =
-        new HashSet<>(((Map<String, String>) config.get("executionTypes")).values());
+    this.accountPattern = Pattern.compile((String) config.get("accountPattern"));
   }
 
   @Override
@@ -46,6 +42,10 @@ public class ByExecutionTypeServiceSelector implements ServiceSelector {
 
   @Override
   public boolean supports(SelectableService.Criteria criteria) {
-    return executionTypes.contains(criteria.getExecutionType());
+    if (criteria.getAccount() == null) {
+      return false;
+    }
+
+    return accountPattern.matcher(criteria.getAccount().toLowerCase()).matches();
   }
 }
--- a/kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/ByCloudProviderServiceSelector.java
+++ b/kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/ByCloudProviderServiceSelector.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 Netflix, Inc.
+ * Copyright 2022 JPMorgan Chase & Co
  *
  * Licensed under the Apache License, Version 2.0 (the "License")
  * you may not use this file except in compliance with the License.
@@ -20,18 +20,18 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 
-public class ByExecutionTypeServiceSelector implements ServiceSelector {
+public class ByCloudProviderServiceSelector implements ServiceSelector {
   private final Object service;
   private final int priority;
-  private final Set<String> executionTypes;
+  private final Set<String> cloudProviders;
 
   @SuppressWarnings("unchecked")
-  public ByExecutionTypeServiceSelector(
+  public ByCloudProviderServiceSelector(
       Object service, Integer priority, Map<String, Object> config) {
     this.service = service;
     this.priority = priority;
-    this.executionTypes =
-        new HashSet<>(((Map<String, String>) config.get("executionTypes")).values());
+    this.cloudProviders =
+        new HashSet<>(((Map<String, String>) config.get("cloudProviders")).values());
   }
 
   @Override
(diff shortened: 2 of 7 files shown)


kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/ByAccountServiceSelector.java
```diff
@@ -0,0 +1,51 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.kork.web.selector;
+
+import java.util.Map;
+import java.util.regex.Pattern;
+
+public class ByAccountServiceSelector implements ServiceSelector {
```
dbyron-sf, line 22: javadoc please

kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/ByAccountServiceSelector.java
```diff
@@ -0,0 +1,51 @@
...
+  private final Object service;
+  private final int priority;
+  private final Pattern accountPattern;
+
+  public ByAccountServiceSelector(Object service, Integer priority, Map<String, Object> config) {
```
dbyron-sf, line 27: Any chance we can see a PR that uses this code?  It probably has to be this way, and there's lots of other code that's like this, but is there any way to be more specific than Object for `service`?  Like a type param?  Also wishing for more specific than Map<String, Object> for config but that's almost certainly a bridge too far.

kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/ByAccountServiceSelector.java
```diff
@@ -0,0 +1,51 @@
...
+    this.service = service;
+    this.priority = priority;
+    this.accountPattern = Pattern.compile((String) config.get("accountPattern"));
```
dbyron-sf, line 30: Why not take `String accountPattern` as a constructor arg if that's all we need?

dbyron-sf: I realize there's a pattern to follow, but it seems pretty tempting to forge a new path.

kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/ByAccountServiceSelector.java
mattgogerly: I'll be opening a PR to Orca for this. The changes I wanted to make affect files that are currently in Groovy, so I have https://github.com/spinnaker/orca/pull/4269 open first to convert them to Java.

Unfortunately the whole selector setup is a bit of an untyped mess. [Here for example](https://github.com/spinnaker/orca/blob/f055f553c2e6f5f42131a8446f8daecc551c4747/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.groovy#L121), is using generics but the actual types are just interfaces that don't extend anything.

The config properties class is just a [Map](https://github.com/spinnaker/orca/blob/f055f553c2e6f5f42131a8446f8daecc551c4747/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfigurationProperties.groovy#L29)

It's all kinda loose in the implementations. Some of the selectors just want a list of strings and some want a single regex. Since the selector class isn't statically configured in configuration it's just naively using the first constructor, so without adding some hellish logic to first introspect what the selector wants I'm not sure it'll be easy to change it.

kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/ByAccountServiceSelector.java
mattgogerly: See above comment, its usage is wired into a lot of reflection and stuff that relies on it being a Map.

mattgogerly:
@Mergifyio update

mattgogerly:
@Mergifyio update

kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/ByAccountServiceSelector.java
dbyron-sf: Now that https://github.com/spinnaker/orca/pull/4269 is in, is there an orca PR to use this yet?

kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/ByAccountServiceSelector.java
dbyron-sf: Can you link to some of this reflection-using code please?

kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/ByAccountServiceSelector.java
mattgogerly: https://github.com/spinnaker/orca/pull/4287

kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/ByAccountServiceSelector.java
mattgogerly: https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfiguration.java#L151

dbyron-sf: Aaah, ok.

kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/ByCloudProviderServiceSelector.java
```diff
@@ -0,0 +1,71 @@
...
+  private final int priority;
+  private final Set<String> cloudProviders;
+
+  @SuppressWarnings("unchecked")
+  public ByCloudProviderServiceSelector(
+      Object service, Integer priority, Map<String, Object> config) {
+    this.service = service;
+    this.priority = priority;
+    this.cloudProviders =
+        new HashSet<>(((Map<String, String>) config.get("cloudProviders")).values());
```
dbyron-sf, line 54: From the docs above, isn't `config.get("cloudProviders")` a `List<String>`?
```
cloudProviders:
  - kubernetes
  - titus
```
but this code casts it to a `Map<String, String>`, right?  There's a groovy test [below](https://github.com/spinnaker/kork/pull/967/files#diff-b5a01344452d99da64995f6a21b5ced1eb15032af0d32baff4b7b00a4bdec3f9R46) with:
```
new ByCloudProviderServiceSelector(oortService, 10, ["cloudProviders": [0: "kubernetes"]]),
```
which is a map...so maybe the docs need updating?  Although a list is prettier if it works.

mattgogerly: Good spot, list works. My implementation was copied from the `ByExecutionTypeServiceSelector`. Who knows why it was done that way.

mattgogerly:
@Mergifyio update

dbyron-sf (approved):

PR #968, opened 2022-07-19
feat(tuning): Bump up default sizes

jasonmcintosh:
At any kind of scale, some of these settings are rather small.  Let's look at bumping some defaults to a TOUCH more conservative levels.  At least in ways that should NOT be 'negative" impact wise.  These are "max" vs. "min".  Open to discussion as well on the above!

ALSO tweaking certain other defaults including:
* Default disable TLS1.1 please!
* Default JDBC pool settings tuned a bit
* Default http request limits... since most services do calls to things like http://spin-clouddriver/ these are INSANELY small.  Let's fix!

### DIFF
--- a/kork-sql/src/main/kotlin/com/netflix/spinnaker/kork/sql/config/ConnectionPoolProperties.kt
+++ b/kork-sql/src/main/kotlin/com/netflix/spinnaker/kork/sql/config/ConnectionPoolProperties.kt
@@ -45,7 +45,7 @@ data class ConnectionPoolProperties(
   var connectionTimeoutMs: Long = TimeUnit.SECONDS.toMillis(5),
   var validationTimeoutMs: Long = TimeUnit.SECONDS.toMillis(5),
   var idleTimeoutMs: Long = TimeUnit.MINUTES.toMillis(3),
-  var maxLifetimeMs: Long = TimeUnit.SECONDS.toMillis(300),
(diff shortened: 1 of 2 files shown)


jvz (approved):

kork-sql/src/main/kotlin/com/netflix/spinnaker/kork/sql/config/ConnectionPoolProperties.kt
```diff
@@ -44,10 +44,10 @@ data class ConnectionPoolProperties(
   var password: String? = null,
   var connectionTimeoutMs: Long = TimeUnit.SECONDS.toMillis(5),
   var validationTimeoutMs: Long = TimeUnit.SECONDS.toMillis(5),
-  var idleTimeoutMs: Long = TimeUnit.MINUTES.toMillis(1),
-  var maxLifetimeMs: Long = TimeUnit.SECONDS.toMillis(30),
+  var idleTimeoutMs: Long = TimeUnit.MINUTES.toMillis(3),
+  var maxLifetimeMs: Long = TimeUnit.SECONDS.toMillis(300),
```
mattgogerly, line 48:
```suggestion
  var maxLifetimeMs: Long = TimeUnit.MINUTES.toMillis(5),
```

mattgogerly (approved):

PR #970, opened 2022-07-20
chore(dependencies): use version 4.13.0 of liquibase to address CVE-2022-0839

dbyron-sf:
introduce SqlMigrationProperty.duplicateFileMode to prevent test failures in SpringStartupTests.  See https://github.com/liquibase/liquibase/issues/2818 for background.

### DIFF
--- a/kork-sql/src/main/kotlin/com/netflix/spinnaker/kork/sql/config/SqlMigrationProperties.kt
+++ b/kork-sql/src/main/kotlin/com/netflix/spinnaker/kork/sql/config/SqlMigrationProperties.kt
@@ -15,6 +15,8 @@
  */
 package com.netflix.spinnaker.kork.sql.config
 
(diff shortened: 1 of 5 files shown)


mattgogerly:
Worth bumping MySQL to the latest while you're at it, or would you rather have that be a separate PR?

dbyron-sf:
> Worth bumping MySQL to the latest while you're at it, or would you rather have that be a separate PR?

I'd like to do it separately.

mattgogerly:
Any updates on this work @dbyron-sf? I think you mentioned there were some test issues in a couple of services?

dbyron-sf:
Not much to update here unfortunately.  https://github.com/liquibase/liquibase/issues/3091 has the latest, but we need someone to actually do the work...

dbyron-sf:
handled by https://github.com/spinnaker/kork/pull/1117

PR #971, opened 2022-07-26
feat(kork-secrets): Add AccessControlled and user secrets updates

jvz:
- Introduces AccessControlled interface to allow fiat to delegate hasPermission() checks to the AccessControlled object itself.
- Introduces UserSecretType annotation for identifying user secret type discriminators for Jackson data-binding.
- Splits up UserSecret into delegate classes UserSecretData and UserSecretMetadata.
- Renames UserSecretMapper to UserSecretSerde to split into a strategy interface along with addition of UserSecretSerdeFactory.
- Adds a "string" user secret type for plain inline secrets like kubeconfig files.
- Updates SecretsManagerSecretEngine to use updated user secrets API.


kork-secrets-aws/src/main/java/com/netflix/spinnaker/kork/secrets/engines/SecretsManagerSecretEngine.java
```diff
@@ -29,33 +32,50 @@
...
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.lang.NonNull;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import lombok.NonNull;
 import org.springframework.stereotype.Component;
 
+/**
+ * Secret engine using AWS Secrets Manager. Authentication is performed using the local AWS
+ * credentials and must have permission to perform {@code secretsmanager:DescribeSecret} and {@code
```
dbyron-sf, line 54: Would it be correct to say "managing account credentials" here (as opposed to credentials for a target account?

kork-secrets-aws/src/main/java/com/netflix/spinnaker/kork/secrets/engines/SecretsManagerSecretEngine.java
```diff
@@ -90,14 +110,29 @@ public UserSecret decrypt(@NonNull UserSecretReference reference) {
...
     String secretRegion = parameters.get(SECRET_REGION);
     String secretName = parameters.get(SECRET_NAME);
-    String encoding = parameters.get(StandardSecretParameter.ENCODING.getParameterName());
+    Map<String, String> tags =
+        getSecretDescription(secretRegion, secretName).getTags().stream()
+            .filter(tag -> tag.getKey().startsWith(UserSecretMetadataField.PREFIX))
+            .collect(Collectors.toMap(Tag::getKey, Tag::getValue));
+    String type = tags.get(UserSecretMetadataField.TYPE.getTagKey());
+    if (type == null) {
+      throw new InvalidSecretFormatException("No spinnaker:type tag found for " + reference);
```
dbyron-sf, line 119:
```suggestion
      throw new InvalidSecretFormatException("No " + UserSecretMetadataField.TYPE.getTagKey() + " tag found for " + reference);
```

kork-secrets-aws/src/main/java/com/netflix/spinnaker/kork/secrets/engines/SecretsManagerSecretEngine.java
```diff
@@ -135,6 +169,21 @@ public void clearCache() {
     cache.clear();
   }
 
+  protected DescribeSecretResult getSecretDescription(String secretRegion, String secretName) {
+    AWSSecretsManager client =
+        AWSSecretsManagerClientBuilder.standard().withRegion(secretRegion).build();
```
dbyron-sf, line 174: Did we talk about this before, or was this another place, that it would be more flexible to take an AWSSecretsManager as an argument to this method?

kork-secrets-aws/src/main/java/com/netflix/spinnaker/kork/secrets/engines/SecretsManagerSecretEngine.java
jvz: Not sure this comment is on the right lines?

dbyron-sf: It's for both the previous line and this one.  Instead of "local AWS credentials", perhaps "AWS managing account credentials" ?

kork-secrets-aws/src/main/java/com/netflix/spinnaker/kork/secrets/engines/SecretsManagerSecretEngine.java
jvz: That's right, we did talk about this, though I think I forgot about that. What were the attributes that were relevant for selecting a client here? The secret region and name?

kork-secrets-aws/src/main/java/com/netflix/spinnaker/kork/secrets/engines/SecretsManagerSecretEngine.java
jvz: Oh I see what you mean. Yeah, let me fix that.

kork-secrets-aws/src/main/java/com/netflix/spinnaker/kork/secrets/engines/SecretsManagerSecretEngine.java
dbyron-sf: clouddriver does a bunch of fancy caching of clients, so I figure it'll want to pass one in.  I bet a version of the method that doesn't take a client will be helpful for the other microservices.  I've added support for multiple managing accounts (still haven't gotten the PRs up yet), but that's also why I'd rather let someone pass in a client.

jvz: How about a request-scoped instance of the client?

dbyron-sf (approved):

PR #974, opened 2022-08-02
Add some missing credentials and secrets APIs

jvz:
- `@CredentialsType` type discriminator
- External secret APIs in `UserSecretManager` (based on `EncryptedSecret` to be analogous to `UserSecret`)

### DIFF
--- a/kork-secrets/src/main/java/com/netflix/spinnaker/kork/secrets/EncryptedSecret.java
+++ b/kork-secrets/src/main/java/com/netflix/spinnaker/kork/secrets/EncryptedSecret.java
@@ -24,6 +24,7 @@ import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.Setter;
+import lombok.extern.log4j.Log4j2;
 
 /**
  * EncryptedSecrets contain an engineIdentifier and named parameters. EncryptedSecrets can be used
@@ -40,6 +41,7 @@ import lombok.Setter;
  */
 @EqualsAndHashCode
 @NoArgsConstructor
+@Log4j2
 public class EncryptedSecret {
 
   public static final String ENCRYPTED_STRING_PREFIX = "encrypted:";
@@ -81,6 +83,7 @@ public class EncryptedSecret {
     try {
       return Optional.of(new EncryptedSecret((String) value));
     } catch (InvalidSecretFormatException e) {
+      log.warn("Tried to parse invalid encrypted secret URI '{}'", value, e);
       return Optional.empty();
     }
   }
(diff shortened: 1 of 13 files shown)


kork-secrets/src/main/java/com/netflix/spinnaker/kork/secrets/EncryptedSecret.java
```diff
@@ -68,6 +70,21 @@ public static EncryptedSecret parse(String secretConfig) {
...
+   * Optional otherwise.
+   */
+  public static Optional<EncryptedSecret> tryParse(@Nullable Object value) {
+    if (!(value instanceof String && isEncryptedSecret((String) value))) {
+      return Optional.empty();
+    }
+    try {
+      return Optional.of(new EncryptedSecret((String) value));
+    } catch (InvalidSecretFormatException e) {
+      return Optional.empty();
```
dbyron-sf, line 84: log something here?

kork-secrets/src/main/java/com/netflix/spinnaker/kork/secrets/EncryptedSecret.java
```diff
@@ -68,6 +70,21 @@ public static EncryptedSecret parse(String secretConfig) {
     return null;
   }
 
+  /**
+   * Tries to parse the provided value as an EncryptedSecret if possible or returns an empty
+   * Optional otherwise.
+   */
+  public static Optional<EncryptedSecret> tryParse(@Nullable Object value) {
```
dbyron-sf, line 77: add tests for tryParse?

kork-secrets/src/main/java/com/netflix/spinnaker/kork/secrets/EncryptedSecret.java
jvz: The context this would be called in would be fairly spammy I think.

dbyron-sf: How "normal" is it for this exception to happen?

dbyron-sf: Seems like if this happens it could cause e.g. clouddriver to not start, or an account to not work, and I bet I could extra help to find out why.  If it's not so frequent then an info or warn message wouldn't flood the logs.

jvz: Usually I'd expect either a CredentialsParser or the underlying BasicCredentialsLoader to end up logging that error, though I see what you mean. I can update that.

dbyron-sf (approved):

PR #977, opened 2022-09-21
fix(expression): toJson throws an exception for valid json.

armory-abedonik:
**What?**

`toJson` method throws an exception for valid JSON when `${}` expression is defined as a string value e.g.

```json
{
  "owner": "managed-by-${profile}"
}
```

We expect that the conversion of valid JSON will occur without any errors.


armory-abedonik:
This pull request is no longer relevant. Closed.

PR #978, opened 2022-09-23
feat(spel): add doNotEval method to restrict expressions evaluation for an object

armory-abedonik:
**What?**

Spinnaker supports [the expression functions](https://github.com/spinnaker/kork/blob/a4617ed929ca34a2216449425ab5b54b5ac271d5/kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/ExpressionsSupport.java#L55) e.g. `toJson`, `toBase64`, `fromBase64` and other. At the moment, `toJson` method [throws an exception](https://github.com/spinnaker/kork/blob/a4617ed929ca34a2216449425ab5b54b5ac271d5/kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/ExpressionsSupport.java#L198-L200) for valid JSON when `${}` expression is defined as a string value

```java
String converted = mapper.writeValueAsString(o);
if (converted != null && converted.contains("${")) {
  throw new SpelHelperFunctionException("result for toJson cannot contain an expression");
}
```

We have a use case where `toJson` method needs to convert such objects without any errors:

```json
{
  "owner": "managed-by-${team}"
}
```

**Solution**:
Introduce a new SpEL `doNotEval` method that wraps incoming JSON object with `NotEvaluableExpression` class. `toJson` method (and others in the future) will not do expressions evaluation or throw any exceptions for instances of `NotEvaluableExpression` class.

**How to use?**

```
${#toJson(#doNotEval(#stage('Terraform Show').outputs.status.outputs.show.planfile_json))}
```

### DIFF
--- a/kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/ExpressionsSupport.java
+++ b/kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/ExpressionsSupport.java
@@ -84,13 +84,17 @@ public class ExpressionsSupport {
                 HashMap.class,
                 LinkedHashMap.class,
                 TreeMap.class,
-                TreeSet.class));
+                TreeSet.class,
+                NotEvaluableExpression.class));
     Collections.addAll(allowedReturnTypes, extraAllowedReturnTypes);
 
     expressionFunctionProviders =
         new ArrayList<>(
             Arrays.asList(
-                new JsonExpressionFunctionProvider(), new StringExpressionFunctionProvider()));
+                new FlowExpressionFunctionProvider(),
+                new JsonExpressionFunctionProvider(),
+                new StringExpressionFunctionProvider()));
+
     if (extraExpressionFunctionProviders != null) {
       expressionFunctionProviders.addAll(extraExpressionFunctionProviders);
     }
@@ -187,6 +191,32 @@ public class ExpressionsSupport {
   }
 
   @SuppressWarnings("unused")
+  public static class FlowExpressionFunctionProvider implements ExpressionFunctionProvider {
+    /**
+     * @param o represents an object to restrict expressions evaluation
+     * @return not evaluable expression
+     */
+    public static NotEvaluableExpression doNotEval(Object o) {
+      return new NotEvaluableExpression(o);
+    }
+
+    @Override
+    public String getNamespace() {
+      return null;
+    }
+
+    @Override
+    public Functions getFunctions() {
+      return new Functions(
+          new FunctionDefinition(
+              "doNotEval",
+              "Restrict expressions evaluation for an object",
+              new FunctionParameter(
+                  Object.class, "value", "An object to restrict expressions evaluation")));
+    }
+  }
+
+  @SuppressWarnings("unused")
   public static class JsonExpressionFunctionProvider implements ExpressionFunctionProvider {
     /**
      * @param o represents an object to convert to json
@@ -194,6 +224,10 @@ public class ExpressionsSupport {
      */
     public static String toJson(Object o) {
       try {
+        if (o instanceof NotEvaluableExpression) {
+          return mapper.writeValueAsString(((NotEvaluableExpression) o).getExpression());
+        }
+
         String converted = mapper.writeValueAsString(o);
         if (converted != null && converted.contains("${")) {
           throw new SpelHelperFunctionException("result for toJson cannot contain an expression");
--- a/kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/NotEvaluableExpression.java
+++ b/kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/NotEvaluableExpression.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2022 Armory, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.kork.expressions;
+
+/**
+ * An instance of this class is used as the return value from the {@code doNotEval} SpEL helper
+ * which makes it possible to skip SpEL evaluation in other SpEL helpers e.g. {@code toJson}.
+ *
+ * <p>For example, in the evaluation context is defined only {@code fileMap} object:
+ *
+ * <pre>{@code
+ * Map<String, Object> fileMap = Collections.singletonMap("owner", "managed-by-${team}");
+ * }</pre>
+ *
+ * <p>An exception will be thrown in attempt to get JSON because of {@code fileMap} contains SpEL
+ * inside.
+ *
+ * <pre>{@code
+ * ${#toJson(fileMap)}
+ * }</pre>
+ *
+ * <p>In the given case {@code fileMap} contains SpEL for another tool e.g. Terraform. Use {@code
+ * doNotEval} to let Spinnaker know that this SpEL should be evaluated by a different tool. No
+ * exceptions are thrown this way.
+ *
+ * <pre>{@code
+ * ${#toJson(#doNotEval(fileMap))}
+ * }</pre>
+ */
+public class NotEvaluableExpression {
+
+  private final Object expression;
+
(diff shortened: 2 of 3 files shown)


kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/ExpressionsSupport.java
```diff
@@ -186,6 +189,32 @@ private void registerExpressionProviderFunctions(StandardEvaluationContext evalu
     }
   }
 
+  @SuppressWarnings("unused")
+  public static class FlowExpressionFunctionProvider implements ExpressionFunctionProvider {
+    /**
+     * @param o represents an object to restrict expressions evaluation
+     * @return not evaluable expression
+     */
+    public static NotEvaluableExpression doNotEval(Object o) {
```
xibz, line 198: This type needs to be added to the [allowed return list](https://github.com/spinnaker/kork/blob/master/kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/ExpressionsSupport.java#L73-L87)

kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/ExpressionsSupport.java
```diff
@@ -194,6 +223,10 @@ public static class JsonExpressionFunctionProvider implements ExpressionFunction
      */
     public static String toJson(Object o) {
       try {
+        if (o instanceof NotEvaluableExpression) {
```
xibz, line 226: I think this is still an issue with anything reprocessing the stage. If you have some stage that calls `toJson`, then any subsequent stage will cause this expression to be evaluated. We need to ensure even through reprocessing of this stage's data doesn't re-evaluate it's SpEL, otherwise it needs to remain the no eval type. Maybe some more testing there to ensure.

kork-expressions/src/test/java/com/netflix/spinnaker/kork/expressions/ExpressionsSupportTest.java
```diff
@@ -0,0 +1,48 @@
...
+package com.netflix.spinnaker.kork.expressions;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import java.util.Collections;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+public class ExpressionsSupportTest {
```
xibz, line 26: We need a test that evaluates the `doNotEval` expression directly. We need something like
```
Collections.singletonMap("owner", "managed-by-${#doNotEval('${team}')");
```

or something of that nature. I believe this test would have caught the allowed return types being missed

kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/NotEvaluableExpression.java
```diff
@@ -0,0 +1,30 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.kork.expressions;
+
+public class NotEvaluableExpression {
```
xibz, line 19: Can we get some documentation explaining the point of this class? More so why someone would see this in their spinnaker pipelines and how to go about evaluating it.

kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/ExpressionsSupport.java
```diff
@@ -90,7 +90,10 @@ public ExpressionsSupport(
     expressionFunctionProviders =
         new ArrayList<>(
             Arrays.asList(
-                new JsonExpressionFunctionProvider(), new StringExpressionFunctionProvider()));
+                new FlowExpressionFunctionProvider(),
```
xibz, line 93: `FlowExpressionFunctionProvider` is a really odd name. Maybe `SpELEvalSkipFunctionProvider`?

xibz (changes requested):
Thanks for the changes @armory-abedonik. I think this does most of what we talked about with a few changes being needed. :)

kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/ExpressionsSupport.java
armory-abedonik: If use `FlowExpressionFunctionProvider` then we can add other flow management expressions in the future e.g. `evalIfCondition`, `evalIfPositive` etc...

kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/ExpressionsSupport.java
armory-abedonik: I've tested the behaviour and it worked even without it... Maybe I messed something... Adderd.

kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/ExpressionsSupport.java
armory-abedonik: As you know, almost all spinnaker users don't use unresolved SpEL here. Only 1% of users do something crazy... In this way they can use `doNotEval` method, and they should take care that the next stages are processed correctly. I don't think that we can cover all possible use cases here.

kork-expressions/src/test/java/com/netflix/spinnaker/kork/expressions/ExpressionsSupportTest.java
armory-abedonik: A new unit test based on `ExpressionTransform` class has been added. Btw, we don't support such `"managed-by-${#doNotEval('${team}')"` use case because `doNotEval` returns an object (not a string).

kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/NotEvaluableExpression.java
armory-abedonik: sure, added

kork-expressions/src/test/java/com/netflix/spinnaker/kork/expressions/ExpressionsSupportTest.java
xibz: That was copy pasta, but yea, just adding a test for `#doNotEval` from the expression parser is what I meant.
```
Collections.singletonMap("owner", "${#doNotEval('${team}')}");
```
We just need to ensure that the object that is returned is returned properly, and then we can even pass that along to a `toJson` call.

kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/ExpressionsSupport.java
xibz: Yes, I think we have two use cases here. Users who want to remove the `doNotEval` when evaluating `toJson` and users who may want to keep `doNotEval` when calling `toJson`. It may be nice to put some metadata in `doNotEval` so it decides how it's evaluated rather than having it live in `toJson`. Like 
```
doNotEval('${someExpression}', 'toJsonRemoval')
```
maybe even having it as a boolean might work, but I can see other stages may needing this same kind of flexibility. We may need to think about this some. The one thing I dislike is having users know string specifics, like `toJsonRemove`, but if enums can be used in SpEL, that can be useful. I haven't explored that though.

What do you think?

armory-abedonik: Could you share an example when and how it should be used. I'm not sure that got you correctly...

xibz: My thinking is a user making a webhook call (maybe this webhook is outside metrics or something) with `toJson` and `doNotEval` which funnels back to Spinnaker. That seems like a valid use case, but the `toJson` that has the SpEL will be re-evaluated and thus returning the value. The main question is are we okay with that?

xibz (approved):

dogonthehorizon:
We'll want to add this to the [SpEL docs](https://spinnaker.io/docs/reference/pipeline/expressions/#helper-functions) but also need to wait until it's released. Noting this here for now and creating an issue in our team's backlog to address.

dogonthehorizon (approved):

dogonthehorizon:
Going to backport this first to 1.28/29 branches quickly followed by backports for https://github.com/spinnaker/kork/pull/981 so that they're appropriately feature flagged

dogonthehorizon:
@mergifyio backport release-1.28.x release-1.29.x

armory-abedonik:
@mergify backport release-1.27.x

PR #982, opened 2022-10-14
feat(build): make the modules that kork-runtime includes configurable

dbyron-sf:
To reduce code size and exposure to vulnerabilities when certain modules aren't being used.  Note that this doesn't actually change anything, but it makes it easier to do so.


jvz (approved):
Nice idea!

PR #991, opened 2022-11-09
fix(exceptions): Exception message should never be empty

xibz:
When socket timeouts occur the response body is null which results in an
empty message being sent back to deck which is not useful.

Signed-off-by: benjamin_j_powell <benjamin_j_powell@apple.com>


xibz:
@dbyron-sf - This is coming from socket timeouts which don't happen too often, and I dont have a stack trace currently as I fixed this a couple weeks ago. However, running a simple mock of a `SocketTimeoutException` should replicate this issue. Originally I thought it was from Rosco, and forgot to update the commit message. I'll update that.

dbyron-sf (approved):

PR #992, opened 2022-11-11
feat(secrets): Add credentials masking in JSON

jvz:
This adds a new `Sensitive` annotation to explicitly mark fields as sensitive. A sensitive field will not be serialized to JSON unless its value is a `secret://...` URI to avoid leaking credentials. This serializer also provides a configurable regular expression to match against `CredentialsDefinition` property names to mark them as likely sensitive which applies masking in addition to logging a warning so that the affected class can be updated.

### DIFF
--- a/kork-credentials/src/main/java/com/netflix/spinnaker/credentials/jackson/SensitiveSerializer.java
+++ b/kork-credentials/src/main/java/com/netflix/spinnaker/credentials/jackson/SensitiveSerializer.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2022 Apple Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.credentials.jackson;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.BeanProperty;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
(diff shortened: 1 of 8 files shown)


kork-credentials/src/main/java/com/netflix/spinnaker/credentials/jackson/SensitiveSerializer.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  @Override
+  public JsonSerializer<?> createContextual(SerializerProvider prov, BeanProperty property)
+      throws JsonMappingException {
+    if (property == null) {
+      return defaultStringSerializer;
+    }
+    Sensitive sensitive = property.getAnnotation(Sensitive.class);
+    if (sensitive != null) {
```
dbyron-sf, line 59: Is the logic backwards here? The idea is to get the "regular" serializer in the absence of a Sensitive annotation, right?
```suggestion
    if (sensitive == null) {
```

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/jackson/SensitiveSerializer.java
```diff
@@ -0,0 +1,88 @@
...
+      return this;
+    }
+    Class<?> declaringClass = property.getMember().getDeclaringClass();
+    if (CredentialsDefinition.class.isAssignableFrom(declaringClass)) {
+      String name = property.getName();
+      Matcher matcher = properties.getSensitivePropertyNamePattern().matcher(name);
+      if (matcher.matches()) {
+        log.warn(
```
dbyron-sf, line 67: I'd prefer info here since this feels like a pretty normal thing to have happen.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/jackson/SensitiveSerializer.java
dbyron-sf: Sorry, `defaultStringSerializer` is the regular serializer here, not `this`.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/jackson/SensitiveSerializer.java
dbyron-sf: Sorry, this check is totally a warning that shows a likely-sensitive property missing an annotation.

dbyron-sf (approved):

dbyron-sf:
Can you give us a peek into where you plan to use this?

jvz:
This is for the account management feature in Clouddriver. I'm working on a PR for that, too, but that's mostly just some unrelated features. Basically, this is something we came up with recently and thought would be useful to contribute.

PR #994, opened 2022-11-18
fix(artifact/gitrepo): resolve symlink clean error

tiewei:
Bump `commons-io` to `2.11.0` which fixed the cleaning
 directory behavior when there is symlink.
    
In version `2.7.0` the delete directory will first check
if the file exists with `Files.exists`, then issue a deleting with
`File.deleteIfExists`. However when it's a symlink, the default
behavior for `Files.exists` is to follow link, if the file/directory
the symlink pointing to has been deleted, the symlink itself will not
be deleted, this will causes error with `deleteDirectory` when there
is still symlink left in the directory.
    
In the newer version will add `NOFOLLOW_LINK` option on the `File.exists`
check, therefore it will work as expected.

### DIFF
--- a/spinnaker-dependencies/spinnaker-dependencies.gradle
+++ b/spinnaker-dependencies/spinnaker-dependencies.gradle
@@ -124,7 +124,7 @@ dependencies {
     api("com.sun.xml.bind:jaxb-impl:2.3.2")
     api("com.vdurmont:semver4j:3.1.0")
     api("commons-configuration:commons-configuration:1.8")
-    api("commons-io:commons-io:2.7")
+    api("commons-io:commons-io:2.11.0")
     // CVE's in 3.2.1
     api("commons-collections:commons-collections:[3.2.2,3.3)")
     api("de.danielbechler:java-object-diff:0.95")


dbyron-sf (approved):

ovidiupopa07 (approved):

dbyron-sf:
Instead of https://github.com/spinnaker/kork/pull/995 and https://github.com/spinnaker/kork/pull/996 let's ask mergify to make "proper" backport PRs.

I'm all for bumping the version...and potentially even backporting it.  I think the thing that's got me confused is [this comment](https://github.com/spinnaker/spinnaker/issues/6759#issuecomment-1321611165).  It makes it seem like the version bump here doesn't fix the issue...and maybe isn't even a prerequisite to fix the issue...that we need to change `FileUtils.deleteDirectory(stagingPath.toFile());` to `FileSystemUtils.deleteRecursively(stagingPath.toFile());`.

weitie-okta:
> Instead of https://github.com/spinnaker/kork/pull/995 and https://github.com/spinnaker/kork/pull/996 let's ask mergify to make "proper" backport PRs.
> 
> 
> 
> I'm all for bumping the version...and potentially even backporting it.  I think the thing that's got me confused is [this comment](https://github.com/spinnaker/spinnaker/issues/6759#issuecomment-1321611165).  It makes it seem like the version bump here doesn't fix the issue...and maybe isn't even a prerequisite to fix the issue...that we need to change `FileUtils.deleteDirectory(stagingPath.toFile());` to `FileSystemUtils.deleteRecursively(stagingPath.toFile());`.

I have tested my use case with 2.11, though the setup isn't the latest master (1.28.1 release with the change). Might need to bring up the latest masters for testing.

tiewei:
> Instead of #995 and #996 let's ask mergify to make "proper" backport PRs.
> 
> I'm all for bumping the version...and potentially even backporting it. I think the thing that's got me confused is [this comment](https://github.com/spinnaker/spinnaker/issues/6759#issuecomment-1321611165). It makes it seem like the version bump here doesn't fix the issue...and maybe isn't even a prerequisite to fix the issue...that we need to change `FileUtils.deleteDirectory(stagingPath.toFile());` to `FileSystemUtils.deleteRecursively(stagingPath.toFile());`.

Me too, I can't reproduce the issue after the patch with the 1.28.1 release BOM though, will need to get the latest setup to try it out.
However, since this issue isn't really a spinnaker bug (more like a commons-io bug) I could write a simple test code like below to test the `deleteDirectory` works with symlink, and the same test code passes in 2.11.0 but failed in 2.8.

```
import org.apache.commons.io.FileUtils;
import java.io.File;
import java.nio.file.Files;
import java.nio.file.Paths;

public class MyTestClass {
    public static void main(String args[]) throws Exception{
      new File("/tmp/testdir/foo/bar").mkdirs();
      // create symlink to non-exist target
      Files.createSymbolicLink(Paths.get("/tmp/testdir/foo/bar/barz"), Paths.get("../../fooz"));
      FileUtils.deleteDirectory(new File("/tmp/testdir"));
    }
}
```

ovidiupopa07:
hi @dbyron-sf. Can you please create a new tag from the master branch?  I need it to pick up the change and test it in our Spinnaker environment.  Thank you!

dogonthehorizon:
@mergifyio backport release-1.29.x release-1.28.x release-1.27.x

PR #1000, opened 2022-12-06
CVE:2021-0341  upgrading package from retrofit retrofit2

Luthan95:
here is the reference that i used.
https://github.com/square/retrofit/issues/1299 

for logging interceptor
https://howtodoinjava.com/retrofit2/logging-with-retrofit2/


kork-web/src/main/groovy/com/netflix/spinnaker/retrofit/RetrofitConfigurationProperties.java
```diff
@@ -16,18 +16,19 @@
 
 package com.netflix.spinnaker.retrofit;
 
+import okhttp3.logging.HttpLoggingInterceptor;
 import org.springframework.boot.context.properties.ConfigurationProperties;
-import retrofit.RestAdapter;
 
-@ConfigurationProperties("retrofit")
+@ConfigurationProperties("retrofit2")
```
dbyron-sf, line 22: This is going to break existing configuration.  Can we leave it as `retrofit`?

PR #1002, opened 2022-12-13
fix & refactor(cloud-config-server): Issue with non-existing file/folder location used in spring cloud config server properties while upgrading spring boot 2.4.x.

j-sandy:
With spring boot 2.4, spring cloud config server location properties will no longer fail silently if the file or folder does not exist. Now, requires prefix "optional:" for the location.

https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.4-Release-Notes#config-file-processing-application-properties-and-yaml-files

Added testcase in support of code change to introduce "optional:" prefix for location properties.

### DIFF
--- a/kork-cloud-config-server/kork-cloud-config-server.gradle
+++ b/kork-cloud-config-server/kork-cloud-config-server.gradle
@@ -11,6 +11,7 @@ dependencies {
   implementation "org.springframework.cloud:spring-cloud-config-server"
   implementation "io.awspring.cloud:spring-cloud-aws-context:2.3.5"
 
+  testImplementation project(':kork-config')
   testImplementation "org.springframework.boot:spring-boot-starter-test"
   testImplementation "org.assertj:assertj-core"
   testImplementation "org.junit.jupiter:junit-jupiter-api"
@@ -18,5 +19,8 @@ dependencies {
   testImplementation "org.junit.jupiter:junit-jupiter-params"
   testImplementation "org.mockito:mockito-core"
 
-  testRuntimeOnly "org.slf4j:slf4j-simple"
+}
+
+test {
+  useJUnitPlatform()
 }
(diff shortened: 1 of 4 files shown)


kork-cloud-config-server/kork-cloud-config-server.gradle
```diff
@@ -6,17 +6,22 @@ dependencies {
...
   implementation "commons-io:commons-io"
   implementation "org.apache.commons:commons-lang3"
+  implementation project(':kork-config')
   implementation project(':kork-secrets')
   implementation "org.springframework.cloud:spring-cloud-context"
   implementation "org.springframework.cloud:spring-cloud-config-server"
   implementation "io.awspring.cloud:spring-cloud-aws-context:2.3.5"
 
   testImplementation "org.springframework.boot:spring-boot-starter-test"
+  testImplementation "org.springframework.cloud:spring-cloud-starter-bootstrap"
```
dbyron-sf, line 16: Don't all users of kork-cloud-config-server need to specify `org.springframework.cloud:spring-cloud-starter-bootstrap` as a dependency?  We'd make their lives easier by making it a runtimeOnly dependency too, yes?  And I bet here, `testRuntimeOnly` is sufficient since nothing needs this to compile, right?

j-sandy: Update the dependency as testRuntimeOnly

dbyron-sf: What do you think about adding `runtimeOnly org.springframework.cloud:spring-cloud-starter-bootstrap` ?

j-sandy: Update the property.

kork-cloud-config-server/kork-cloud-config-server.gradle
```diff
@@ -6,6 +6,7 @@ dependencies {
 
   implementation "commons-io:commons-io"
   implementation "org.apache.commons:commons-lang3"
+  implementation project(':kork-config')
```
dbyron-sf, line 9: I tried this locally and it looks like this only needs to be 
```suggestion
  testImplementation project(':kork-config')
```
so DefaultPropertiesBuilder is available to ConfigServerTestApplication.

j-sandy: Updated the dependency as suggested.

dbyron-sf (approved):

PR #1004, opened 2023-01-09
feat(retrofit): add support for retrofit2 clients that handle errors by generating Spinnaker(Server|Http|Network)Exceptions

Luthan95:
Global Error handling approach for changing from retrofit 1.9 to retrofit 2.x version
Creating Custom callAdaptor to handle the failure and throw Custom exception using retrofit2 response
This will be used throw the exceptions SpinnakerHttpException, SpinnakerNetworkException and SpinnakerServerException based on response status. So this will help in handling exception/error globally
 
 References for ErrorHandler:
https://github.com/square/retrofit/blob/parent-2.8.1/samples/src/main/java/com/example/retrofit/ErrorHandlingAdapter.java
https://github.com/square/retrofit/blob/parent-2.8.1/retrofit/src/main/java/retrofit2/DefaultCallAdapterFactory.java

Retrofit2 client with error handler sample:-
```
Service svc = new Retrofit.Builder()
        .baseUrl("endpoint") // endpoint of service 
        .client(okHttpClient) // okhttp3 client with required interceptors and properties 
        .addCallAdapterFactory(ErrorHandlingExecutorCallAdapterFactory.getInstance())  // Call factory to handle error globally and to make use of custom spinnaker exceptions
        .addConverterFactory(JacksonConverterFactory.create(objectMapper))
        .build()
        .create(Service.class); 
```
Retrofit2 API sample:
 ```
 @GET("/fetch/")
  Call<ResponseBody> find(@Query("q") String name);

```
Use of Retrofit2 client call sample:

> By default  execute method of retrofit2 call throws IOException, so to handle globally make use of Retrofit2SyncCall.execute(Call<T> call) static method.

 `ResponseBody findResponse = Retrofit2SyncCall.execute(svc.find("name"))`
 
 **Note:**
New property which added to support the loglevel of retrofit2 client call, default to BASIC level

```
retrofit2:
  logLevel: BASIC // NONE,HEADERS,BODY
```

### CODE REVIEW

File: kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java [hunk 1]
```diff
@@ -0,0 +1,223 @@
+/*
+ * Copyright 2023 Netflix, Inc.
```
dbyron-sf, line 2: OpsMx?

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -28,13 +28,18 @@
 @Getter
 @NonnullByDefault
 public class SpinnakerHttpException extends SpinnakerServerException {
-  private final Response response;
+  private final Object response;
```
dbyron-sf, line 31: I haven't looked to see if anyone calls getResponse(), but this is going to break those calls.  Do we need a getResponse method?  If not, let's remove it in a separate PR, and then this change is safe.

dbyron-sf: OK, I see one call that you've updated here.  Are there more?

dbyron-sf: Seems like it would be cleaner to add a getResponseCode (or some other name) method here and remove getResponse...at least for the handleSpinnakerHttpException case.  Would be great to do that in a separate PR, and then adjust the method here.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -0,0 +1,223 @@
...
+import java.io.IOException;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.concurrent.Executor;
+import okhttp3.Request;
+import okio.Timeout;
+import org.jetbrains.annotations.Nullable;
+import org.springframework.http.HttpStatus;
+import retrofit2.*;
```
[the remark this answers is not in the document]
Luthan95: resolved

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
Luthan95: make changes

kork-retrofit/src/test/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerRetrofitErrorHandlerTest.java
```diff
@@ -56,6 +65,14 @@ public static void setupOnce() throws Exception {
             .setErrorHandler(SpinnakerRetrofitErrorHandler.getInstance())
             .build()
             .create(RetrofitService.class);
+
+    retrofit2Service =
```
dbyron-sf, line 69: Do you mind moving these to a separate test class?  I feel like it'll be easier to follow if the retrofit2 stuff has its own home.

Luthan95: Sure, I will make changes and create new class

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/Retrofit2SyncCall.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package com.netflix.spinnaker.kork.retrofit;
+
+import com.netflix.spinnaker.kork.retrofit.exceptions.RetrofitException;
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerNetworkException;
+import java.io.IOException;
+import retrofit2.Call;
+
+public class Retrofit2SyncCall<T> {
```
dbyron-sf, line 24: please add tests for this too.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -56,8 +56,16 @@ public SpinnakerHttpException(RetrofitException e) {
...
     super(message, cause);
     // Note that getRawMessage() is null in this case.
-    this.response = cause.getResponse();
-    this.responseCode = cause.getResponseCode();
+    this.response = cause.response;
+    this.retrofit2Response = cause.retrofit2Response;
+  }
+
+  public int getResponseCode() {
+    if (response != null) {
```
dbyron-sf, line 64: Worth a check to make sure only one of response or retrofit2Response is set?  If they're both set, we've got a mess.

Luthan95: Used netflix spectator Preconditions.checkState to check both response and retrofit2Response not set together

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -16,6 +16,7 @@
 
 package com.netflix.spinnaker.kork.retrofit.exceptions;
 
+import com.netflix.spectator.impl.Preconditions;
```
dbyron-sf, line 19: Let's use
```suggestion
import com.google.common.base.Preconditions;
```
instead.

Luthan95: com.google.common.base.Preconditions instead of com.netflix.spectator.impl.Preconditions for this added dependency com.google.guava:guava in kork-retrofit module

kork-retrofit/src/test/java/com/netflix/spinnaker/kork/retrofit/exceptions/Retrofit2SyncCallTest.java
```diff
@@ -0,0 +1,54 @@
...
+    String execute = Retrofit2SyncCall.execute(mockcall);
+    assertEquals("testing", execute);
+  }
+
+  @Test
+  public void testExecuteThrowException() throws IOException {
+    Call<String> mockcall = Mockito.mock(Call.class);
+    IOException ioException = new IOException("exception test");
+    when(mockcall.execute()).thenThrow(ioException);
+    assertThrows(
```
dbyron-sf, line 48: Can we also assert that the cause of the SpinnakerNetworkException is ioException?

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerRetrofitExceptionHandlers.java
```diff
@@ -63,7 +63,9 @@ public void handleSpinnakerHttpException(
     // We made an http request that failed and nothing else handled that
     // failure, so generate our response based on the response we received.
     storeException(request, response, e);
+
```
dbyron-sf, line 66: Please remove the whitespace-only changes from here since there aren't any other changes.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -16,21 +16,33 @@
...
+import com.google.common.base.Preconditions;
 import com.netflix.spinnaker.kork.annotations.NonnullByDefault;
 import retrofit.RetrofitError;
 import retrofit.client.Response;
 
 /**
- * An exception that exposes the {@link Response} of a given HTTP {@link RetrofitError} and a detail
- * message that extracts useful information from the {@link Response}.
+ * An exception that exposes the {@link Response} of a given HTTP {@link RetrofitError} or {@link
+ * okhttp3.Response} of given Http {@link RetrofitException} if retrofit 2.x used and a detail
```
dbyron-sf, line 26:
```suggestion
 * okhttp3.Response} of a {@link RetrofitException} if retrofit 2.x used and a detail
```

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -0,0 +1,250 @@
...
+      }
+
+      @Override
+      public Call<Object> adapt(Call<Object> call) {
+        return new ExecutorCallbackCall<>(callbackExecutor, call, retrofit);
+      }
+    };
+  }
+
+  static final class ExecutorCallbackCall<T> implements Call<T> {
```
dbyron-sf, line 78: Javadoc please.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -0,0 +1,250 @@
...
+      return delegate.request();
+    }
+
+    @Override
+    public Timeout timeout() {
+      return delegate.timeout();
+    }
+  }
+
+  static class MyExecutorCallback<T> implements Callback<T> {
```
dbyron-sf, line 155: javadoc please, and a more descriptive class name that reflects what's actually happening here.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -0,0 +1,250 @@
...
+          new Runnable() {
+            @Override
+            public void run() {
+              callback.onFailure(executorCallbackCall, finalException);
+            }
+          });
+    }
+  }
+
+  public static class MainThreadExecutor implements Executor {
```
dbyron-sf, line 228: I don't see this used anywhere.  Please remove it if that's really true.  If we do actually need it, it needs javadoc, and some non-trivial explanation of why we need it.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/Retrofit2SyncCall.java
```diff
@@ -0,0 +1,37 @@
...
+
+  /*
+   * This method used to handle IOException from {@link Call}.execute method centrally,
```
dbyron-sf, line 27:
```suggestion
   * Handle IOExceptions from {@link Call}.execute method centrally,
```

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/Retrofit2SyncCall.java
```diff
@@ -0,0 +1,37 @@
...
+   * instead of handling exception in all places of retrofit2 API call
```
dbyron-sf, line 28:
```suggestion
   * instead of all places that make retrofit2 API calls.
```

kork-retrofit/kork-retrofit.gradle
```diff
@@ -11,6 +11,10 @@ dependencies {
   implementation "com.jakewharton.retrofit:retrofit1-okhttp3-client"
   implementation "com.squareup.retrofit:converter-jackson"
   implementation "io.zipkin.brave:brave-instrumentation-okhttp3"
+  implementation "com.squareup.retrofit2:retrofit"
+  implementation "com.squareup.retrofit2:converter-jackson"
+  implementation "com.squareup.okhttp3:logging-interceptor"
+  implementation "com.google.guava:guava"
```
dbyron-sf, line 17: Are we using guava anywhere in this PR?

kork-retrofit/kork-retrofit.gradle
```diff
@@ -20,5 +24,5 @@ dependencies {
   testImplementation "com.squareup.okhttp3:mockwebserver"
   testRuntimeOnly "cglib:cglib-nodep"
   testRuntimeOnly "org.objenesis:objenesis"
-
+  testImplementation "org.mockito:mockito-inline"
```
dbyron-sf, line 27: Let's use mockito-extensions here too.  See https://github.com/spinnaker/rosco/pull/953#discussion_r1126981146.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -0,0 +1,266 @@
...
+import org.jetbrains.annotations.Nullable;
+import org.springframework.http.HttpStatus;
+import retrofit2.Call;
+import retrofit2.CallAdapter;
+import retrofit2.Callback;
+import retrofit2.Response;
+import retrofit2.Retrofit;
+
+/*
+ * As part retrofit2 {@link retrofit.RetrofitError} and {@link retrofit.ErrorHandler} are removed.
```
dbyron-sf, line 40:
```suggestion
 * {@link retrofit.RetrofitError} and {@link retrofit.ErrorHandler} are no longer present in retrofit2.
```

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -0,0 +1,266 @@
...
+  public static ErrorHandlingExecutorCallAdapterFactory getInstance(
+      @Nullable Executor callbackExecutor) {
+    return new ErrorHandlingExecutorCallAdapterFactory(callbackExecutor);
+  }
+
+  @Nullable
+  @Override
+  public CallAdapter<?, ?> get(Type returnType, Annotation[] annotations, Retrofit retrofit) {
+
+    if (getRawType(returnType) != Call.class) {
```
dbyron-sf, line 61: This could use a comment in the code.  Why do we need this check?  Better to throw an exception if the check fails?

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -0,0 +1,266 @@
...
+import retrofit2.Retrofit;
+
+/*
+ * As part retrofit2 {@link retrofit.RetrofitError} and {@link retrofit.ErrorHandler} are removed.
+ * So this class helps to achieve similar logic as retrofit and handle exceptions globally in retrofit2.
+ * This can be achieved by setting this class as CallAdapterFactory at the time of {@link Retrofit} client creation.
+ *  */
+public class ErrorHandlingExecutorCallAdapterFactory extends CallAdapter.Factory {
+
+  private final @Nullable Executor callbackExecutor;
```
dbyron-sf, line 46: Why does this need to be nullable?

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -0,0 +1,266 @@
...
+   * <p>Calls may be executed synchronously with {@link #execute}, or asynchronously with {@link
+   * #enqueue}. In either case Spinnaker(Http|Network|Server)Exception will be thrown if the
+   * response is not successful or an unexpected error occurs creating the request or decoding the
+   * response.
+   *
+   * @param <T> Successful response body type.
+   */
+  static final class ExecutorCallbackCall<T> implements Call<T> {
+
+    final Executor callbackExecutor;
```
dbyron-sf, line 91: private for these members as well?

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -0,0 +1,266 @@
...
+        Response<T> syncResp = delegate.execute();
+        if (syncResp.isSuccessful()) {
+          return syncResp;
+        }
+        throw createSpinnakerHttpException(syncResp);
+      } catch (SpinnakerHttpException e) {
+        throw e;
+      } catch (Exception e) {
+        if (e instanceof IOException) {
+          throw new SpinnakerNetworkException(RetrofitException.networkError((IOException) e));
```
dbyron-sf, line 119:
```suggestion
          throw new SpinnakerNetworkException(RetrofitException.networkError(e));
```

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -0,0 +1,266 @@
...
+      try {
+        Response<T> syncResp = delegate.execute();
+        if (syncResp.isSuccessful()) {
+          return syncResp;
+        }
+        throw createSpinnakerHttpException(syncResp);
+      } catch (SpinnakerHttpException e) {
+        throw e;
+      } catch (Exception e) {
+        if (e instanceof IOException) {
```
dbyron-sf, line 118: It's cleaner to have a separate catch for IOException.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
dbyron-sf: please add javadoc for the members too.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -0,0 +1,266 @@
...
+     *     creating the request or decoding the response
+     */
+    @Override
+    public Response<T> execute() {
+      try {
+        Response<T> syncResp = delegate.execute();
+        if (syncResp.isSuccessful()) {
+          return syncResp;
+        }
+        throw createSpinnakerHttpException(syncResp);
```
dbyron-sf, line 114: I suspect it's cleaner to move `throw createSpinnakerHttpException(syncResp);` outside this try/catch.  Then there's no need to catch and re-throw.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -0,0 +1,266 @@
...
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerServerException;
+import java.io.IOException;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.concurrent.Executor;
+import okhttp3.Request;
+import okio.Timeout;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
```
dbyron-sf, line 31:
```suggestion
import javax.annotation.Nullable;
```

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -0,0 +1,266 @@
...
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerNetworkException;
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerServerException;
+import java.io.IOException;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.concurrent.Executor;
+import okhttp3.Request;
+import okio.Timeout;
+import org.jetbrains.annotations.NotNull;
```
dbyron-sf, line 30:
```suggestion
import javax.annotation.Nonnull;
```

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
dbyron-sf: Aaah, I see [the javadoc](https://square.github.io/retrofit/2.x/retrofit/index.html?retrofit2/CallAdapter.Factory.html) says to return null if the factory can't handle it.  Still curious why we need the check.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -0,0 +1,266 @@
...
+
+  static Type getCallResponseType(Type returnType) {
+    if (!(returnType instanceof ParameterizedType)) {
+      throw new IllegalArgumentException(
+          "Call return type must be parameterized as Call<Foo> or Call<? extends Foo>");
+    }
+    return getParameterUpperBound(0, (ParameterizedType) returnType);
+  }
+
+  static <T> T checkNotNull(@Nullable T object, String message) {
```
dbyron-sf, line 260: replace with [Objects.requireNonNull](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Objects.html#requireNonNull(T,java.lang.String)).

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/RetrofitException.java
```diff
@@ -0,0 +1,114 @@
...
+    return url;
+  }
+
+  /** Response object containing status code, headers, body, etc. */
+  public Response getResponse() {
+    return response;
+  }
+
+  /** The event kind which triggered this error. */
+  public Kind getKind() {
```
dbyron-sf, line 89: I don't see any code that uses this method.  Until we need it, let's remove it (and the Kind enum).

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/RetrofitException.java
```diff
@@ -0,0 +1,114 @@
...
+    /** A non-200 HTTP status code was received from the server. */
+    HTTP,
+    /**
+     * An internal error occurred while attempting to execute a request. It is best practice to
+     * re-throw this exception so your application crashes.
+     */
+    UNEXPECTED
+  }
+
+  private final String url;
```
dbyron-sf, line 59: javadoc please

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/RetrofitException.java
```diff
@@ -0,0 +1,114 @@
...
+    return response;
+  }
+
+  /** The event kind which triggered this error. */
+  public Kind getKind() {
+    return kind;
+  }
+
+  /** The Retrofit this request was executed on */
+  public Retrofit getRetrofit() {
```
dbyron-sf, line 94: Same here.  Let's remove it until we need it.

kork-retrofit/src/test/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerRetrofit2ErrorHandleTest.java
```diff
@@ -0,0 +1,132 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.kork.retrofit.exceptions;
+
+import static org.junit.jupiter.api.Assertions.*;
```
dbyron-sf, line 19: please avoid star imports

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
Luthan95: It is required only if the async calls are used. In most of the services, only sync calls are used, we don't need to set it.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
Luthan95: I will throw SpinnakerServerException instead of returning null.

dbyron-sf: Returning null seems the correct behavior based on the javadoc. I'd still like a comment in the code explaining the check.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
dbyron-sf: Great.  Please add some javadoc explaining this.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -0,0 +1,256 @@
...
+    ExecutorCallbackCall(Executor callbackExecutor, Call<T> delegate, Retrofit retrofit) {
+      this.callbackExecutor = callbackExecutor;
+      this.delegate = delegate;
+      this.retrofit = retrofit;
+    }
+
+    /**
+     * Synchronously send the request and return its response.
+     *
+     * @throws SpinnakerServerException (and subclasses) if an unexpected error occurs while
```
dbyron-sf, line 114:
```suggestion
     * @throws SpinnakerServerException (and subclasses) if an error occurs while
```

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -0,0 +1,256 @@
...
+   * <p>Calls may be executed synchronously with {@link #execute}, or asynchronously with {@link
+   * #enqueue}. In either case Spinnaker(Http|Network|Server)Exception will be thrown if the
+   * response is not successful or an unexpected error occurs creating the request or decoding the
+   * response.
+   *
+   * @param <T> Successful response body type.
+   */
+  static final class ExecutorCallbackCall<T> implements Call<T> {
+
+    private final Executor callbackExecutor;
```
dbyron-sf, line 101: javadoc for these members please

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/RetrofitException.java
```diff
@@ -0,0 +1,79 @@
...
+
+  /** Response object containing status code, headers, body, etc. */
+  public Response getResponse() {
+    return response;
+  }
+
+  /**
+   * HTTP response body converted to specified {@code type}. {@code null} if there is no response.
+   *
+   * @throws IOException if unable to convert the body to the specified {@code type}.
```
dbyron-sf, line 66:
```suggestion
   * @throws RuntimeException wrapping the underlying IOException if unable to convert the body to the specified {@code type}.
```

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/RetrofitException.java
```diff
@@ -0,0 +1,79 @@
...
+   */
+  public <T> T getErrorBodyAs(Class<T> type) {
```
dbyron-sf, line 68: It's pretty tempting to add a Retrofit argument to this method and remove the class member since this is the only place it's used.

Luthan95: There is no other way that seems to be, if we use a common form of retrofit clients for all services, we can add Retrofit as an argument, but services using different conversion forms.

dbyron-sf: Well, it does nicely match the way RetrofitError works today.  An alternative would be to add a
```
String rawMessage;
```
member to RetrofitException and build that at construction time.  Given the dance that's already present in SpinnakerServerException for handling rawMessage vs. message, I'm fine to leave it as it is.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -0,0 +1,261 @@
...
+          new Runnable() {
+            @Override
+            public void run() {
+              callback.onFailure(executorCallbackCall, finalException);
+            }
+          });
+    }
+  }
+
+  static Type getCallResponseType(Type returnType) {
```
dbyron-sf, line 254: javadoc please.  It's not super obvious why the caller checks
```
    if (getRawType(returnType) != Call.class) {
```
and this code checks:
```
    if (!(returnType instanceof ParameterizedType)) {
```
especially because the exception message here doesn't mention ParameterizedType.

I could use an explanation.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -0,0 +1,261 @@
...
+
+    /*
+     * Expected the raw class type from returnType to be Call class
+     * otherwise return null as it cannot be handled by this factory
+     * */
+    if (getRawType(returnType) != Call.class) {
+      return null;
+    }
+
+    final Type responseType = getCallResponseType(returnType);
```
dbyron-sf, line 74: I could use some comments in the code explaining what's happening here.  The javadoc in the parent says:
> Returns a call adapter for interface methods that return returnType, or null if it cannot be handled by this factory.

But the details are a bit lost on me.  Why is this the right code to accomplish that?

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -0,0 +1,279 @@
...
+     * */
+    if (getRawType(returnType) != Call.class) {
+      return null;
+    }
+
+    /*
+     * check whether returnType is ParameterizedType, else throws IllegalArgumentException.
+     * */
+    if (!(returnType instanceof ParameterizedType)) {
+      throw new IllegalArgumentException(
```
dbyron-sf, line 90: Could the comment here explain why we're throwing an IllegalArgumentException instead of returning null since the above code returns null and the javadoc for the method we're implementing says to return null....or maybe switch to returning null here too?

If they're not too contrived...like they actually could happen because someone is trying to use retrofit2 properly, but doesn't understand it, I'd like to see tests for these two failure paths....or at least comments in the code that explain how these paths can happen, ideally with examples.

If the tests "artificially" call this method with specially-crafted params, as opposed to something close to a "real" usage, let's not do it.

kork-retrofit/src/test/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerRetrofit2ErrorHandleTest.java
```diff
@@ -146,8 +147,24 @@ public void testResponseHeadersInException() throws JsonProcessingException {
...
+        "Call return type must be parameterized as Call<Foo> or Call<? extends Foo>",
+        illegalArgumentException.getCause().getMessage());
+  }
+
   interface Retrofit2Service {
     @retrofit2.http.GET("/retrofit2")
     Call<String> getRetrofit2();
+
+    @retrofit2.http.GET("/retrofit2/para")
+    Call testNotParameterized();
```
dbyron-sf, line 168: For completeness, please add another method like
```
    @retrofit2.http.GET("/retrofit2/wrongReturnType")
    DummyWithExecute testWrongReturnType();
```
and
```
  interface DummyWithExecute {
    void execute();
  }
```
and a test to verify the other error handling in ErrorHandlingExecutorCallAdapterFactory.CallAdapter.get.

kork-retrofit/src/test/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerRetrofit2ErrorHandleTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+    mockWebServer.enqueue(
+        new MockResponse().setResponseCode(HttpStatus.GONE.value()).setBody(responseBodyString));
+    SpinnakerHttpException spinnakerHttpException =
+        assertThrows(SpinnakerHttpException.class, () -> retrofit2Service.getRetrofit2().execute());
+    assertNull(spinnakerHttpException.getRetryable());
+  }
+
+  @Test
+  public void testRetrofitSimpleSpinnakerNetworkException() {
```
dbyron-sf, line 115: This test takes 10 seconds to run on my machine.  Any way to speed it up?

Luthan95: yes, will override the default timeout and make it 1 sec

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
Luthan95: as per discussion, it is better to return null than throw an exception, so returning null

kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttp3ClientConfiguration.groovy
```diff
@@ -46,6 +48,9 @@ class OkHttp3ClientConfiguration {
   private final OkHttpClientConfigurationProperties okHttpClientConfigurationProperties
   private final OkHttp3MetricsInterceptor okHttp3MetricsInterceptor
 
+  @Autowired
+  private SpinnakerRequestHeaderInterceptor spinnakerRequestHeaderInterceptor;
```
dbyron-sf, line 52: Please use a constructor argument instead of autowiring this way.

dbyron-sf: Please add some javadoc as well to make it clear this is for retrofit2.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -0,0 +1,275 @@
...
+import okhttp3.Request;
+import okio.Timeout;
+import org.springframework.http.HttpStatus;
+import retrofit2.Call;
+import retrofit2.CallAdapter;
+import retrofit2.Callback;
+import retrofit2.Response;
+import retrofit2.Retrofit;
+
+/*
```
kirangodishala, line 40: @Luthan95 begin-comment delimiter(/**) is not properly added at several places, can you correct them?

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -0,0 +1,275 @@
...
+   * <p>Calls may be executed synchronously with {@link #execute}, or asynchronously with {@link
+   * #enqueue}. In either case Spinnaker(Http|Network|Server)Exception will be thrown if the
+   * response is not successful or an unexpected error occurs creating the request or decoding the
+   * response.
+   *
+   * @param <T> Successful response body type.
+   */
+  static final class ExecutorCallbackCall<T> implements Call<T> {
+
+    /*The executor used for Callback methods on a Call.*/
```
kirangodishala, line 123: Can you replace all
`/*coments like this */`
with
```
/**
* comments like this
*/
```

kork-web/src/test/groovy/com/netflix/spinnaker/kork/web/interceptors/SpinnakerRequestHeaderInterceptorTest.groovy
```diff
@@ -0,0 +1,53 @@
...
+import com.netflix.spinnaker.kork.common.Header
+import com.netflix.spinnaker.okhttp.OkHttpClientConfigurationProperties
+import com.netflix.spinnaker.okhttp.SpinnakerRequestHeaderInterceptor
+import com.netflix.spinnaker.security.AuthenticatedRequest
+import okhttp3.Interceptor
+import okhttp3.Request
+import org.mockito.Mockito
+import spock.lang.Specification
+
+class SpinnakerRequestHeaderInterceptorTest extends Specification {
```
dbyron-sf, line 28: It's probably a bit late to point this out, but ideally this would be java.

kork-web/src/main/groovy/com/netflix/spinnaker/okhttp/SpinnakerRequestHeaderInterceptor.groovy
```diff
@@ -0,0 +1,51 @@
...
+
+import com.netflix.spinnaker.security.AuthenticatedRequest
+import okhttp3.Interceptor
+import okhttp3.Response
+
+/**
+* As {@link retrofit.RequestInterceptor} no longer present in retrofit2,
+* we have to use {@link okhttp3.Interceptor} to add authenticated headers to requests.
+* */
+class SpinnakerRequestHeaderInterceptor implements Interceptor {
```
dbyron-sf, line 27: Ideally this would be java too.  Tests in groovy are potentially ok...but for new source code we really do need to avoid groovy.

kork-web/kork-web.gradle
```diff
@@ -47,4 +47,5 @@ dependencies {
   testImplementation "com.netflix.spectator:spectator-reg-micrometer"
   testRuntimeOnly "cglib:cglib-nodep"
   testRuntimeOnly "org.objenesis:objenesis"
+  testRuntimeOnly "org.mockito:mockito-inline"
```
dbyron-sf, line 50: Is this necessary with kork-retrofit/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker ?

kork-web/src/main/groovy/com/netflix/spinnaker/okhttp/SpinnakerRequestHeaderInterceptor.groovy
Luthan95: Sure I will change this to Java code

kork-web/kork-web.gradle
Luthan95: I will remove it.

kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttp3ClientConfiguration.groovy
```diff
@@ -46,7 +49,23 @@ class OkHttp3ClientConfiguration {
   private final OkHttpClientConfigurationProperties okHttpClientConfigurationProperties
   private final OkHttp3MetricsInterceptor okHttp3MetricsInterceptor
 
+  @Value("\${retrofit.log-level:BASIC}")
+  private final String retrofitLogLevel;
```
dbyron-sf, line 53: There's already a [RetrofitConfigurationProperties bean](https://github.com/spinnaker/kork/blob/2a9a71d506a60d4ec47c5d46b4871ee322fb61c1/kork-web/src/main/groovy/com/netflix/spinnaker/retrofit/RetrofitConfigurationProperties.java), so let's use it.

Luthan95: Yeah, but it is like a hardcoded one, in case the user wants to override to log level.

dbyron-sf: I don't understand.  Both with this mechanism and the one in RetrofitConfigurationProperties, there's a default value.  Also with both, there's a way for the user to override it by setting the `rettrofit.log-level` property, right?  As in, they're equivalent, aren't they?

dbyron-sf: Aaah, but RetrofitConfigurationProperties uses RestAdapter, but that's a retrofit1 type...that's the issue then, isn't it?

Luthan95: yeah...

dbyron-sf: It's pretty tempting to make a Retrofit2ConfigurationProperties that does the equivalent thing but with `HttpLoggingInterceptor.Level`.  If you have the stamina to do it in this PR, go for it.  Otherwise, let's do it in a followup.

kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttp3ClientConfiguration.groovy
```diff
@@ -83,6 +91,43 @@ class OkHttp3ClientConfiguration {
...
+    }
+
+    if (okHttp3MetricsInterceptor != null) {
+      okHttpClientBuilder.addInterceptor(okHttp3MetricsInterceptor)
+    }
+
+    /**
+     * The logging functionality was removed in Retrofit2, since the required HTTP layer is now completely based on OkHttp.
+     * Recommend to add logging as the last interceptor, because this will also log the information
+     * which you added with previous interceptors to your request.f
```
dbyron-sf, line 119:
```suggestion
     * which you added with previous interceptors to your request.
```

kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttp3ClientConfiguration.groovy
Luthan95: created new class Retrofit2ConfigurationProperties  with HttpLoggingInterceptor.Level

kork-web/src/main/groovy/com/netflix/spinnaker/retrofit/Retrofit2ConfigurationProperties.java
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright 2023 Netflix, Inc.
```
dbyron-sf, line 2:
```suggestion
 * Copyright 2023 OpsMx, Inc.
```

kork-web/src/main/groovy/com/netflix/spinnaker/retrofit/Retrofit2ConfigurationProperties.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.retrofit;
+
+import okhttp3.logging.HttpLoggingInterceptor;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties("retrofit2")
```
dbyron-sf, line 22: If it works to use `retrofit` I'd like to.  That way no one has to migrate their properties.

Luthan95: But RestdAdaptor.Level has HEADERS_AND_ARGS and FULL, which is not present in Retrofit2, if we use the same property for retrofit1.9 and retrofit2, it may cause issues...

dbyron-sf: Fair point.  If we do it this way (`retrofit2`), we need a release note once https://github.com/spinnaker/rosco/pull/953 goes in so rosco users know to update their config.  Please add a note to the PR description here mentioning the new config property.

PR #1006, opened 2023-01-13
fix(dependencies): include aws-java-sdk-sts

tiewei:
When running the service in an environment requires to assume into another role to fetch the required secrets, it will require the `aws-java-sdk-sts`. This happens when the service is running as EKS pod with service account associated with an IAM role.


jasonmcintosh (approved):

link108 (approved):

weitie-okta:
@link108 do you mind making it backport to `1.29`?

tiewei:
@link108 do you mind let the mergify to backport it?

PR #1010, opened 2023-01-23
refactor(SpinnakerHttpException): created a method which return status code

Luthan95:
Created a new method getResponseCode which will return status code of SpinnakerHttpException.
Removed to the getResponse method as all the services using the method to get the Status code only so getResponseCode method will get used instead of getResponse().getStatus()

### DIFF
--- a/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
+++ b/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
@@ -17,7 +17,6 @@
 package com.netflix.spinnaker.kork.retrofit.exceptions;
 
 import com.netflix.spinnaker.kork.annotations.NonnullByDefault;
-import lombok.Getter;
 import retrofit.RetrofitError;
 import retrofit.client.Response;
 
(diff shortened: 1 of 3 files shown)


kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -49,7 +47,11 @@ public SpinnakerHttpException(RetrofitError e) {
   public SpinnakerHttpException(String message, SpinnakerHttpException cause) {
     super(message, cause);
     // Note that getRawMessage() is null in this case.
-    this.response = cause.getResponse();
```
dbyron-sf, line 52: nit: We're in an internal method , so `cause.response` is OK.  In the spirit of minimizing diffs, I'd leave it as it was though.

dbyron-sf:
@Luthan95 what do you think of https://github.com/spinnaker/kork/pull/1010/commits/4674067a74cb72bc1e58799a7aeeee8a5028978c?  I was having a bit of trouble wrapping my head around the mock-with-real-response from makeSpinnakerHttpException, so I tried a "plain" mock, but ran into other struggles.  This seems like a good way forward.  It's a bit more involved to build a real RetrofitError, but it seems more straightforward than a hybrid mock, and less mocking == better test coverage.

dbyron-sf (approved):

PR #1012, opened 2023-01-27
fix(selectors): fix type cast error on cloud provider selector

mattgogerly:
Last comment in https://github.com/spinnaker/kork/pull/967 for background.

As it turns out, `List<String>` doesn't work, and Orca blows up in spectacular fashion. My guess is because the constructor arg is a `Map<String, Object>`, Jackson doesn't know how to handle the `Object` and defaults to `LinkedHashMap`.

I don't want to break existing configs, so just fixing this selector class to use a `Map`. Eurgh.

cc @dbyron-sf

### DIFF
--- a/kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/ByCloudProviderServiceSelector.java
+++ b/kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/ByCloudProviderServiceSelector.java
@@ -17,7 +17,6 @@
 package com.netflix.spinnaker.kork.web.selector;
 
 import java.util.HashSet;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
@@ -51,7 +50,7 @@ public class ByCloudProviderServiceSelector implements ServiceSelector {
       Object service, Integer priority, Map<String, Object> config) {
     this.service = service;
     this.priority = priority;
-    this.cloudProviders = new HashSet<>((List<String>) config.get("cloudProviders"));
+    this.cloudProviders = new HashSet<>(((Map<String, String>) config.get("cloudProviders")).values());
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


kork-web/src/main/java/com/netflix/spinnaker/kork/web/selector/ByCloudProviderServiceSelector.java
```diff
@@ -51,7 +50,8 @@ public ByCloudProviderServiceSelector(
       Object service, Integer priority, Map<String, Object> config) {
     this.service = service;
     this.priority = priority;
-    this.cloudProviders = new HashSet<>((List<String>) config.get("cloudProviders"));
+    this.cloudProviders =
+        new HashSet<>(((Map<String, String>) config.get("cloudProviders")).values());
```
dbyron-sf, line 54: Does the comment above need to change to match this?  Seems like it's showing a list now.

mattgogerly: It's always been a YAML list, Jackson just deserializes it into a Map for whatever reason. With the config in that comment I get a class cast exception from LinkedHashMap to List.

dbyron-sf: You get a class cast exception before this code change, right?  I don't suppose there's way to get a test to fail the same way?

mattgogerly: Yup. 

I don't think so without including Jackson deserialising as part of the test. If you want it I'll give it a go next week.

dbyron-sf: Yes please.  This seems subtle enough that I'd like some protection against future breakage.

mattgogerly: Added a test in https://github.com/spinnaker/orca/pull/4287

dbyron-sf (approved):

PR #1013, opened 2023-02-02
Migrate credentials storage API to Kork

jvz:
This brings Kork up to date with the great migration recently completed and will be one part of upstreaming. We'll still need a Fiat update to bridge over the migrated classes from there as well as a Clouddriver update to migrate to the new classes (and removal of the old ones which are now in Kork). There will also need to be a Gate update to make sure all the proper APIs are exported.

I intend to target these changes for 1.30. All the aforementioned services will also require some updates for that.


jvz:
> The following commits need their title changed:
> 
> * [0eeb9fc](https://github.com/spinnaker/kork/commit/0eeb9fc63fc0043a358a2a13dede4c13913a547d): update(secrets): Add UserSecretData.getSecretString() for scalar secrets
> 
> Please format your commit title into the form:
> 
> `<type>(<scope>): <subject>`, e.g. `fix(kubernetes): address NPE in status check`
> 
> This allows us to easily generate changelogs & determine semantic version numbers when cutting releases. You can read more about [commit conventions](https://www.spinnaker.io/community/contributing/submitting/#commit-message-conventions) here.

That link is a 404. Get real.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/service/CredentialsDefinitionCommand.java
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.credentials.service;
+
+/** Denotes a type of credential definition command. Useful for validation methods. */
+public enum CredentialsDefinitionCommand {
```
jasonmcintosh, line 20: Did I miss delete in here? (just normally a command operation?

dbyron-sf:
Sorry, I thought @link108 was working on getting a shiny new spinnakerbot running with up-to-date links.  See https://spinnaker.io/docs/community/contributing/code/submitting/#commit-and-pr-message-conventions.

kork-security/src/main/java/com/netflix/spinnaker/security/Authorization.java
```diff
@@ -0,0 +1,29 @@
...
+
+package com.netflix.spinnaker.security;
+
+import org.springframework.security.core.Authentication;
+
+/**
+ * Defines types of authorizations supported by {@link AccessControlled#isAuthorized(Authentication,
+ * Object)} and by extension, {@link org.springframework.security.access.PermissionEvaluator}.
+ */
+public enum Authorization {
```
jasonmcintosh, line 25: Missing Create?

jasonmcintosh: ASSUMING matching existing auth system....

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/service/CredentialsDefinitionService.java
```diff
@@ -0,0 +1,168 @@
...
+  @PreAuthorize("hasAnyAuthority('SPINNAKER_ADMIN', 'SPINNAKER_ACCOUNT_MANAGER')")
+  public void update(CredentialsDefinition definition, Authentication auth) {
+    validate(definition, auth, CredentialsDefinitionCommand.UPDATE);
+    repository.update(definition);
+  }
+
+  @PreAuthorize(
+      "hasAuthority('SPINNAKER_ADMIN') or hasAuthority('SPINNAKER_ACCOUNT_MANAGER') and hasPermission(#accountName, 'account', 'write')")
+  public void delete(String accountName) {
+    repository.delete(accountName);
```
jasonmcintosh, line 145: Missing validation intentionally on delete operations?  Wouldn't the same access rules apply here?

kork-credentials-api/src/main/java/com/netflix/spinnaker/credentials/definition/CredentialsDefinitionRepository.java
```diff
@@ -0,0 +1,152 @@
...
+   * @see #create(CredentialsDefinition)
+   */
+  void update(CredentialsDefinition definition);
+
+  /**
+   * Deletes an account by name.
+   *
+   * @param name name of account to delete
+   */
+  void delete(String name);
```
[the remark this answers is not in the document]
jvz: That seems like it complicates the API IMO.

kork-credentials-api/src/main/java/com/netflix/spinnaker/credentials/definition/CredentialsDefinitionRepository.java
```diff
@@ -0,0 +1,152 @@
...
+   * @return history of account updates for the given account name
+   */
+  List<Revision> revisionHistory(String name);
+
+  /**
+   * Provides metadata for an account definition revision when making updates to an account via
+   * {@link CredentialsDefinitionRepository} APIs.
+   */
+  class Revision {
+    private final int version;
```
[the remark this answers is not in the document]
jvz: This is automatically incremented on each save. It's there so that a future update can support reverting changes and whatnot. This is the same pattern as from Front50.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/service/CredentialsDefinitionCommand.java
jvz: We don't need to validate the credentials definition body in a delete call.

kork-security/src/main/java/com/netflix/spinnaker/security/Authorization.java
jvz: Intentionally dropped from here because it's only used in Front50 for applications.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/service/CredentialsDefinitionService.java
jvz: There's a permission check in the annotation; that's all the validation we need. If the account doesn't exist in the database, we'll get back a 404 once we do a database lookup to validate it.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/secrets/CredentialsDefinitionSecretManager.java
```diff
@@ -0,0 +1,103 @@
...
+      return true;
+    }
+    if (refsByAccountName.getOrDefault(accountName, Set.of()).stream()
+        .map(this::getUserSecret)
+        .anyMatch(
+            secret ->
+                !permissionEvaluator.hasPermission(authentication, secret, Authorization.READ))) {
+      return false;
+    }
+    return permissionEvaluator.hasPermission(
```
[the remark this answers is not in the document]
jvz: This method is for doing time-of-use checks against loaded credentials that used a user secret. Since credentials are always loaded from a background thread with no authentication context, we have to defer permission checks until later.

For user secrets, all `Authorization` types are treated equivalently as they're read-only anyways. The `WRITE` permission requirement here is because you can't _use_ an account to perform an action without said permission.

kork-credentials-api/src/main/java/com/netflix/spinnaker/credentials/CredentialsError.java
```diff
@@ -0,0 +1,35 @@
+/*
```
dbyron-sf, line 1: Sorry if this is obvious, but would you mind linking to the existing code in clouddriver that's moving here?  It will help me figure out what's new/changed vs. a straight move.

kork-secrets/src/main/java/com/netflix/spinnaker/kork/secrets/user/UserSecretManager.java
```diff
@@ -17,10 +17,7 @@
 package com.netflix.spinnaker.kork.secrets.user;
 
 import com.netflix.spinnaker.kork.annotations.NonnullByDefault;
-import com.netflix.spinnaker.kork.secrets.EncryptedSecret;
-import com.netflix.spinnaker.kork.secrets.SecretDecryptionException;
-import com.netflix.spinnaker.kork.secrets.SecretEngine;
-import com.netflix.spinnaker.kork.secrets.SecretEngineRegistry;
+import com.netflix.spinnaker.kork.secrets.*;
```
dbyron-sf, line 20: I think we're still trying to avoid '*' imports.  Bonus points if you can teach spotless to do this for us.

kork-credentials-api/src/main/java/com/netflix/spinnaker/credentials/CredentialsError.java
jvz: Anything new in this API module was taken from https://github.com/spinnaker/clouddriver/tree/master/clouddriver-api/src/main/java/com/netflix/spinnaker/clouddriver/security (`AccountDefinitionRepository` renamed to `CredentialsDefinitionRepository`, and `AccountDefinitionTypeProvider` renamed to `CredentialsDefinitionTypeProvider`. This particular error class was added for the new credentials inspection API as Spring's `BindException` and other validation-related classes like `ObjectError` do not support Jackson, so I needed a simpler DTO for that. The `AccessControlled...` class in there remains as it's specific to Clouddriver and relies on Fiat as a dependency besides just Kork.

Then, everything in https://github.com/spinnaker/clouddriver/tree/master/clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/security besides the two abstract classes at the top of the list are migrated (and apparently somewhat rewritten; I did not ever get around to submitting some of the changes to Clouddriver before I started migrating them to Kork). In particular, here's how they map to the updated code:

* `AuthorizedRolesExtractor` and implementation classes are no longer required as I've defined an `AccessControlled` interface in `kork-security` which provides the necessary functionality to perform permission checks via Fiat (though that currently relies on a feature flag in Fiat which I will be toggling on by default for this release as it's starting to become required for various things in this updated API).
* `AccountSecurityPolicy` is generally replaced by simpler usage of `PermissionEvaluator` from Spring Security in addition to a newly defined API, `CredentialsDefinitionValidator`, which provides security-related validation (amongst other validations) when saving an account.
* `AccountDefinitionTypes` renamed to `CredentialsDefinitionTypes`.
* `AccountDefinitionSource` renamed to `CompositeCredentialsDefinitionSource` (one of the only classes that couldn't be consistently renamed to match how every other class was renamed since `CredentialsDefinitionSource` is an existing interface).
* `AccountDefinitionService` renamed to `CredentialsDefinitionService`.
* `AccountDefinitionMapper` renamed to `CredentialsDefinitionMapper`.
* `AccountDefinitionSecretManager` renamed to `CredentialsDefinitionSecretManager`.

And then for the SQL stuff, that's migrated and renamed from https://github.com/spinnaker/clouddriver/tree/master/clouddriver-sql/src/main/kotlin/com/netflix/spinnaker/clouddriver/sql/security with some changes:
* I defined some `CustomRecord` and `CustomTable` classes similar to the pattern in Fiat for more readable jOOQ usage.

kork-security/src/main/java/com/netflix/spinnaker/security/Authorization.java
dbyron-sf: I clearly don't have enough of the different concepts here straight.  My first thought when seeing this was...what about creating accounts (or pipelines)?  But there's an Account Manger role (or [list of roles](https://github.com/spinnaker/fiat/blob/6fd331c044c96a885bb8fdc1373822b08086758d/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/AccountManagerConfig.java)) for that, right?  I have a feeling I'm requesting an essay (which I don't really mean to), or something that is maybe written up elsewhere, or could be?  What are the pros/cons of having create vs. a separate role?

kork-credentials-api/src/main/java/com/netflix/spinnaker/credentials/definition/CredentialsDefinitionRepository.java
```diff
@@ -0,0 +1,152 @@
...
+   * value in the {@link CredentialsType} or {@code com.fasterxml.jackson.annotation.JsonTypeName}
+   * annotation present on the corresponding {@link CredentialsDefinition} class.
+   *
+   * @param typeName credential type to search for
+   * @return list of all stored credentials definitions matching the given account type name
+   */
+  List<? extends CredentialsDefinition> listByType(String typeName);
+
+  /**
+   * Lists credentials views for a given credentials type accessible to the current user.
```
dbyron-sf, line 51: Is there really enough information to know what's accessible to the current user?  Doesn't seem like the sql implementation of this does any filtering like that.

kork-credentials-api/src/main/java/com/netflix/spinnaker/credentials/definition/CredentialsDefinitionRepository.java
```diff
@@ -0,0 +1,152 @@
...
+   */
+  List<Revision> revisionHistory(String name);
+
+  /**
+   * Provides metadata for an account definition revision when making updates to an account via
+   * {@link CredentialsDefinitionRepository} APIs.
+   */
+  class Revision {
+    private final int version;
+    private final long timestamp;
```
dbyron-sf, line 120: I realize this is the same as in [clouddriver](https://github.com/spinnaker/clouddriver/blob/8ddccd8c21d1c06b22a910f1e44ef9c826fd6526/clouddriver-api/src/main/java/com/netflix/spinnaker/clouddriver/security/AccountDefinitionRepository.java#L119), but...if this won't bust already-serialized info, how 'bout making this all nice and explicit like:
```suggestion
    private final long timestampMillis;
```

kork-credentials-sql/src/main/kotlin/com/netflix/spinnaker/credentials/sql/AccountHistoryRecordRevisionMapper.kt
```diff
@@ -0,0 +1,39 @@
...
+ */
+@Component
+class AccountHistoryRecordRevisionMapper(private val mapper: CredentialsDefinitionMapper) :
+  RecordMapper<AccountHistoryRecord, Revision> {
+  override fun map(record: AccountHistoryRecord): Revision =
+    record.body?.let {
+      try {
+        mapper.deserialize(it.data())
+      } catch (e: JsonProcessingException) {
+        null
```
dbyron-sf, line 36: Why not let this exception bubble up?  Or at least log something?

kork-credentials-sql/src/main/kotlin/com/netflix/spinnaker/credentials/sql/AccountHistoryRecordRevisionMapper.kt
```diff
@@ -0,0 +1,39 @@
...
+package com.netflix.spinnaker.credentials.sql
+
+import com.fasterxml.jackson.core.JsonProcessingException
+import com.netflix.spinnaker.credentials.definition.CredentialsDefinitionRepository.Revision
+import com.netflix.spinnaker.credentials.secrets.CredentialsDefinitionMapper
+import org.jooq.RecordMapper
+import org.springframework.stereotype.Component
+
+/**
+ * [RecordMapper] for historical account revisions.
```
dbyron-sf, line 26:
```suggestion
 * [RecordMapper] for loading historical account revisions.
```

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/secrets/CredentialsDefinitionMapper.java
```diff
@@ -0,0 +1,224 @@
...
+   * @param string the JSON string to parse
+   * @return the parsed account with loaded secrets
+   * @throws JsonProcessingException if the input string is invalid JSON or does not map to an
+   *     existing CredentialsDefinition class
+   */
+  public CredentialsDefinition deserializeWithSecrets(String string)
+      throws JsonProcessingException {
+    JsonNode jsonNode = objectMapper.readTree(string);
+    if (!jsonNode.isObject()) {
+      throw new SpinnakerException("Expected an object node but got " + jsonNode.getNodeType());
```
dbyron-sf, line 108: Can't tell the convention here, but it seems like other javadoc mentions throwing exceptions like this.  I'm waffling about whether there isn't some more specific exception type available from jackson (e.g. InvalidFormatException, though I somehow suspect that's for actual malformed json).  Also curious if there isn't more to include in the message to help track down the error (e.g. jsonNode.toString()).

dbyron-sf: Would it help to have a `deserializeWithSecretsWithErrors` method?

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/service/CredentialsInspector.java
```diff
@@ -0,0 +1,102 @@
...
+                Collectors.toMap(
+                    CompositeCredentialsDefinitionSource::getTypeName, Function.identity()));
+    this.permissionEvaluator = permissionEvaluator;
+  }
+
+  /**
+   * Enumerates and inspects all credential definitions the given authenticated user has access to.
+   * This combines all {@link CredentialsDefinitionRepository} and {@link
+   * CredentialsDefinitionSource} definitions. The resulting list of credentials should be checked
+   * for duplicate account names. Credentials with permissions defined should be filtered in or out
```
dbyron-sf, line 66: This method does seem to check for dups...so is this part of the comment really correct?

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/types/CredentialsDefinitionTypeProperties.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.List;
+import lombok.Data;
+
+@Data
+public class CredentialsDefinitionTypeProperties {
+  /**
+   * Additional packages to scan for {@link
+   * com.netflix.spinnaker.credentials.definition.CredentialsDefinition} implementation classes that
+   * may be annotated with {@link CredentialsType} to participate in the account management system.
+   * These packages are in addition to the default scan package from within Clouddriver. Note that
```
dbyron-sf, line 30: Still correct to mention Clouddriver here?

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/validator/AccessControlledCredentialsDefinitionValidator.java
```diff
@@ -0,0 +1,54 @@
...
+import com.netflix.spinnaker.security.AccessControlled;
+import com.netflix.spinnaker.security.Authorization;
+import lombok.RequiredArgsConstructor;
+import org.springframework.security.access.PermissionEvaluator;
+import org.springframework.security.core.Authentication;
+import org.springframework.stereotype.Component;
+import org.springframework.validation.Errors;
+
+/**
+ * Validates that an authenticated user has authorization to write with credentials. This helps to
```
dbyron-sf, line 30:
```suggestion
 * Validates that an authenticated user has authorization to write credentials. This helps to
```
?

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/validator/CredentialsDefinitionValidatorProperties.java
```diff
@@ -0,0 +1,39 @@
...
+
+package com.netflix.spinnaker.credentials.validator;
+
+import java.util.Map;
+import java.util.regex.Pattern;
+import lombok.Data;
+
+@Data
+public class CredentialsDefinitionValidatorProperties {
+  private static final Pattern DEFAULT_VALID_ACCOUNT_NAME_PATTERN =
```
dbyron-sf, line 25: Could use a comment on how this pattern came to be the chosen one.

kork-credentials-api/src/main/java/com/netflix/spinnaker/credentials/definition/CredentialsDefinitionRepository.java
jvz: There's a filter later on that checks the permissions of each object.

kork-credentials-sql/src/main/kotlin/com/netflix/spinnaker/credentials/sql/AccountHistoryRecordRevisionMapper.kt
jvz: So if someone ever manages to create a historical revision with invalid JSON or something, we should get spammed in the logs every time?

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/secrets/CredentialsDefinitionMapper.java
jvz: The point of this method is to throw an exception on error, not to return the error. The one with errors returned is the other API where we want to see _all_ the possible errors. The latter API is being used in a UI for this, and providing a user with a list of all the errors encountered is more useful than playing whack a mole with the errors. This method, however, can throw an exception as soon as it needs to.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/service/CredentialsInspector.java
jvz: Those are more so directions to the caller on what they have to do with these results.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/types/CredentialsDefinitionTypeProperties.java
jvz: Should be updated, yeah, even if Clouddriver is the only user of the API so far (I'm looking at adopting it in Kayenta, too, at the moment).

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/validator/CredentialsDefinitionValidatorProperties.java
jvz: Starts with a letter, ends with a letter or number, and in between may use underscores or dashes. Mostly a suggested regex if you don't already have some sort of naming scheme in use. You can also specify different regular expressions for each credentials type in the `valid-account-name-patterns` config map values.

kork-security/src/main/java/com/netflix/spinnaker/security/Authorization.java
jvz: `CREATE` is only supported by some config setting where you can restrict applications based on name prefixes for additional restrictions on who can create applications under what name. I'm hoping to replace the account manager virtual role one day with a `MANAGE` authorization type, but that would be part of the Abarth proposal. I didn't reuse `CREATE` for accounts as the more relevant permission here was also related to managing the account in general (sort of like an admin permission, not to be confused with an admin role).

kork-credentials-api/src/main/java/com/netflix/spinnaker/credentials/definition/CredentialsDefinitionRepository.java
dbyron-sf: OK...does it make sense to change the comment here though?

kork-credentials-sql/src/main/kotlin/com/netflix/spinnaker/credentials/sql/AccountHistoryRecordRevisionMapper.kt
dbyron-sf: No, I guess not.  Seems worth a comment though.  You're reminding me that I'm pretty sure it is important to catch these...otherwise revision history totally disappears if there's one busted historical revision.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/secrets/CredentialsDefinitionMapper.java
dbyron-sf: OK.  What do you think about including more info to help track down the offending credentials?

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/service/CredentialsInspector.java
dbyron-sf: Right, but...if we're checking here, do callers really need to do that?

kork-security/src/main/java/com/netflix/spinnaker/security/Authorization.java
dbyron-sf: OK, thanks for the explanation.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/validator/CredentialsDefinitionValidatorProperties.java
dbyron-sf: This doesn't really explain why, but it does seem like a reasonable default.  Thanks for pointing out that it's configurable.

kork-credentials-api/src/main/java/com/netflix/spinnaker/credentials/definition/CredentialsDefinitionRepository.java
jvz: Yeah I think so.

jasonmcintosh (approved):
I think overall looks good - will be interested in how it operates when we start using it, if find any weird serialization or other quirks, but ... I think it's a good migration we can use elsewhere.  Permissions pieces... we'll see but ... let's try see how goes

jasonmcintosh:
(I"m approving as a whole... ideally fix some of the grammar then merge ;) )

kork-credentials-api/src/main/java/com/netflix/spinnaker/credentials/definition/CredentialsDefinitionRepository.java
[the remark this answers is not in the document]
jvz: Ok, turns out something in between will be needed at the high level. In order to figure out which service to send the deletion request to at least.

jvz:
I'll mention here that I found a small problem in my proposed `CredentialsDefinitionService` related to permission enforcement. While it would work fine for cloud provider accounts, if you have any other account resource types defined in Fiat, then the permission check will always return `false`. I'll update this PR later with a fix for that in addition to feedback from comments.

link108:
@dbyron-sf ah I left that hanging, its fixed now 💯

kork-credentials-sql/src/main/resources/db/changelog/20210927-accounts.yml
```diff
@@ -0,0 +1,162 @@
...
+              nullable: false
+    - modifySql:
+        dbms: mysql
+        append:
+          value: " engine innodb DEFAULT CHARSET=utf8mb4 COLLATE utf8mb4_unicode_ci"
+    - modifySql:
+        dbms: postgresql
+        replace:
+          replace: json
+          with: jsonb
```
jvz, line 54: Just noticed that this doesn't actually work. Will have a second Liquibase file for this.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/validator/AccessControlledCredentialsDefinitionValidator.java
jvz: Should say "use" instead of "write with". Updating.

kork-credentials-api/src/main/java/com/netflix/spinnaker/credentials/definition/CredentialsDefinitionRepository.java
jvz: Kind of already being used in the UI.

kork-credentials-api/src/main/java/com/netflix/spinnaker/credentials/definition/CredentialsNavigator.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+/**
+ * Navigates the sea of credentials provided by a Spinnaker service.
+ *
+ * @param <T> type of credentials definition tracked
+ */
+@Alpha
+public interface CredentialsNavigator<T extends CredentialsDefinition>
```
jvz, line 31: This new API is being used in some Kayenta integration, though I'd like to adopt it in Clouddriver, too, at some point to try and avoid the need for polling from storage.

kork-credentials-sql/src/main/resources/db/changelog/20230217-accounts-body-typefix.yml
```diff
@@ -0,0 +1,17 @@
+databaseChangeLog:
+  - changeSet:
+      id: fix-accounts-body-column
+      author: msicker
+      preConditions:
+        - onFail: MARK_RAN
```
jvz, line 6: This seems to be required to make sure this change set doesn't fail when using non-Postgres databases (like MySQL).

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/secrets/CredentialsDefinitionMapper.java
jvz: In `AccountRecordCredentialsDefinitionMapper`, the account name is logged (at info or warning depending on the type of error), and there are two counter metrics for tracking the number of errors.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/secrets/CredentialsDefinitionSecretManager.java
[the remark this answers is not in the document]
jvz: I should note that this method is currently only invoked from `DescriptionAuthorizerService::authorize` in Clouddriver.

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/service/CredentialsDefinitionService.java
jvz: Updated with a more direct way for checking permissions.

jvz:
Updated with some finishing touches, docs, and feedback addressed.

jvz:
And hopefully the bot will be happy now about the commit history.

kyledmitchell:
@jvz What work is required to complete the original scope laid out in this PR and close it out.  What has been completed so far?

From original description it sounds like

- Initial work to migrate credential API to Kork
- Fiat update to bridge over the migrated classes
- Clouddriver update to migrate to the new classes (and removal of the old ones which are now in Kork). 
- A Gate update to make sure all the proper APIs are exported

jvz:
Yeah, that's the gist, though I've got some other API updates I'm working on (over the past couple months) that would be useful to include at the same time. Mostly some updates related to the schema and support for conditional updates (like via E-Tag comparisons). The Fiat updates would mostly be relevant for unifying the Authorization enum (probably easier with a monorepo) and indexing Kayenta metrics accounts (can be done separately). Clouddriver will mostly be deleting code (or replacing with extended classes from Kork). Gate will have updates related to supporting multiple potential credentials backend services (like Kayenta, Igor, or anything else that could benefit from it).

kork-credentials/src/main/java/com/netflix/spinnaker/credentials/validator/UserSecretsValidator.java
```diff
@@ -0,0 +1,149 @@
...
+    BeanWrapper bean = PropertyAccessorFactory.forBeanPropertyAccess(definition);
+    validate(bean, errors, authentication, true);
+  }
+
+  private void validate(
+      BeanWrapper bean, Errors errors, Authentication authentication, boolean saving) {
+    // when saving an account, only admins can use external secrets
+    boolean denyUserSecrets = saving && !AccessControlled.isAdmin(authentication);
+    Stream.of(bean.getPropertyDescriptors())
+        .filter(descriptor -> descriptor.getPropertyType() == String.class)
```
jvz, line 75: Good thing we waited on this. This needs to consider more types.

jvz:
Superseded by https://github.com/spinnaker/kork/pull/1127

PR #1015, opened 2023-02-07
chore(dependencies): use version 1.2.10 of logback to resolve CVE-2021-42550

dbyron-sf:
spring boot 2.4.13 brings in logback 1.2.7 which is vulnerable to CVE-2021-42550, fixed in 1.2.9.  spring boot 2.5.14 brings in 1.2.11 (the newest as of 7-feb-23), but use 1.2.10 to avoid https://jira.qos.ch/browse/LOGBACK-1623 which makes V2BakeryControllerTest in rosco hang.

### DIFF
--- a/spinnaker-dependencies/spinnaker-dependencies.gradle
+++ b/spinnaker-dependencies/spinnaker-dependencies.gradle
@@ -15,6 +15,10 @@ ext {
     jooq             : "3.13.6",
     jsch             : "0.1.54",
     jschAgentProxy   : "0.0.9",
+    // spring boot 2.5.14 specifies logback 1.2.11, but a rosco test hung with
+    // 1.2.11 from https://jira.qos.ch/browse/LOGBACK-1623 so stick with 1.2.10
+    // until 1.2.12 appears.
+    logback          : "1.2.10",
     protobuf         : "3.19.4",
     okhttp           : "2.7.5", // CVE-2016-2402
     okhttp3          : "3.14.9",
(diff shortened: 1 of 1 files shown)


spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -70,6 +74,9 @@ dependencies {
 
   constraints {
     api("cglib:cglib-nodep:3.3.0")
+    api("ch.qos.logback:logback-access:${versions.logback}")
+    api("ch.qos.logback:logback-classic:${versions.logback}")
+    api("ch.qos.logback:logback-core:${versions.logback}")
```
jasonmcintosh, line 79: Wondering if... logback-parent or something might be better?  IF there's an option like that or a BOM on that... for now... approve, just hate individual version constraints like this potentially missing stuff :(

jasonmcintosh (approved):

spinnaker-dependencies/spinnaker-dependencies.gradle
dbyron-sf: Agreed.  I don't see a bom...There is [logback-parent](https://repo.maven.apache.org/maven2/ch/qos/logback/logback-parent/1.2.10/logback-parent-1.2.10.pom) but it seems to me it specifies too much.

jasonmcintosh: agreed... let's stick with this and hopefully rip it out later.

PR #1016, opened 2023-02-08
fix(web): properly handle an empty MDC in MdcCopyingAsyncTaskExecutor

dbyron-sf:
This is a corner case that's unlikely to happen, but we may as well fix it.


dbyron-sf:
@mergifyio update

mattgogerly (approved):

PR #1022, opened 2023-02-22
fix(jedis): get reference to delegate's internal pool

Young-Zen:
The internalPool field is declared in JedisPool's super class, getDeclaredField("internalPool") could not get internalPool field from super class.

### DIFF
--- a/kork-jedis/src/test/java/com/netflix/spinnaker/kork/jedis/telemetry/InstrumentedJedisPoolTest.java
+++ b/kork-jedis/src/test/java/com/netflix/spinnaker/kork/jedis/telemetry/InstrumentedJedisPoolTest.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2023 THL A29 Limited, a Tencent company.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.kork.jedis.telemetry;
+
+import static org.junit.Assert.assertNotNull;
+import static org.mockito.Mockito.mock;
+
+import com.netflix.spectator.api.Registry;
+import org.junit.Test;
+import redis.clients.jedis.JedisPool;
+
+public class InstrumentedJedisPoolTest {
+
+  private Registry registry = mock(Registry.class);
+
+  private JedisPool jedisPool = new JedisPool();
+
+  private InstrumentedJedisPool instrumentedJedisPool =
+      new InstrumentedJedisPool(registry, jedisPool);
+
+  @Test
+  public void getInternalPoolReference() {
+    assertNotNull(instrumentedJedisPool.getInternalPoolReference());
+  }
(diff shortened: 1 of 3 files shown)


dbyron-sf:
This seems a like a pretty fundamental thing...like, I'm curious why no one has run into this before.  With the current code (i.e. before this PR), what does it take to trigger this bug, and what does failure look like?

Young-Zen:
> This seems a like a pretty fundamental thing...like, I'm curious why no one has run into this before. With the current code (i.e. before this PR), what does it take to trigger this bug, and what does failure look like?

Yes, I'm not the first to trigger this bug, like this [code comment](https://github.com/spinnaker/clouddriver/blob/3842ef77845120d09a575b421a0441b155bf0c4e/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/RedisConfig.groovy#L147): Thanks to groovy for avoiding this bug (um, I think Groovy doesn't respect the private modifier). Before this PR, calling the getInternalPoolReference() method would throw NoSuchFieldException.

kork-jedis/src/test/java/com/netflix/spinnaker/kork/jedis/telemetry/InstrumentedJedisPoolTest.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright 2023 Netflix, Inc.
```
dbyron-sf, line 2: Do you work for Netflix?

dbyron-sf:
Fair enough.  What I guess I'm still curious about is...how did you run into this bug?  What's the code path that ends up calling getInternalPoolReference ?

Young-Zen:
> Fair enough. What I guess I'm still curious about is...how did you run into this bug? What's the code path that ends up calling getInternalPoolReference ?

I found this bug when I wanted to print the current state parameters of internalPool (such as [NumActive](https://github.com/spinnaker/kork/blob/4a020a130ca4fd9113f23bd8c9fba150860a8d78/kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/telemetry/InstrumentedJedisPool.java#L99)) in the log. I think there is no other place to call getInternalPoolReference method other than the above mentioned Groovy code may call.

kork-jedis/src/test/java/com/netflix/spinnaker/kork/jedis/telemetry/InstrumentedJedisPoolTest.java
Young-Zen: Oh, no, I'm working at Tencent now. It seems to be automatically generated when new a Java class. I will delete it.

kork-jedis/src/test/java/com/netflix/spinnaker/kork/jedis/telemetry/InstrumentedJedisPoolTest.java
```diff
@@ -0,0 +1,23 @@
+package com.netflix.spinnaker.kork.jedis.telemetry;
```
dbyron-sf, line 1: Would you please put back the license header, but please add an appropriate copyright holder?

Young-Zen: Okay, I will follow this [copyright](https://github.com/spinnaker/clouddriver/blob/3842ef77845120d09a575b421a0441b155bf0c4e/clouddriver-tencentcloud/src/test/java/com/netflix/spinnaker/clouddriver/tencentcloud/cache/KeysTest.java#L2).

dbyron-sf: Thanks!

dbyron-sf (approved):

PR #1023, opened 2023-03-03
feat(secrets): Add CacheManager support to secret engines

jvz:
Using a `CacheManager` bean from Spring, this updates the `SecretEngine` implementations to use named `Cache`s from said cache manager. A default fallback bean is provided.


jvz:
With https://github.com/spinnaker/kork/pull/1049, this can be more easily accomplished via `@Cacheable`-style annotations instead.

PR #1025, opened 2023-03-17
feat(retrofit): Add getHeaders() method to SpinnakerHttpException

abe21412:
This PR adds a `getHeaders()` method to `SpinnakerHttpException` so that the response headers for an HTTP request can be examined during exception handling. This is needed as part of the effort to [consume SpinnakerRetrofitHandler in each service](https://github.com/spinnaker/spinnaker/issues/5473).

### DIFF
--- a/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
+++ b/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
@@ -28,6 +28,7 @@ import retrofit.client.Response;
 @NonnullByDefault
 public class SpinnakerHttpException extends SpinnakerServerException {
   private final Response response;
+  private HttpHeaders headers;
 
   public SpinnakerHttpException(RetrofitError e) {
     super(e);
@@ -56,8 +57,10 @@ public class SpinnakerHttpException extends SpinnakerServerException {
   }
 
   public HttpHeaders getHeaders() {
-    HttpHeaders headers = new HttpHeaders();
-    response.getHeaders().forEach(header -> headers.add(header.getName(), header.getValue()));
+    if (headers == null) {
+      headers = new HttpHeaders();
+      response.getHeaders().forEach(header -> headers.add(header.getName(), header.getValue()));
+    }
     return headers;
   }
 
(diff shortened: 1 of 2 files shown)


abe21412:
> Thanks for this @abe21412 ! Would you mind including a link to some code that could use this?

Sure, [this method](https://github.com/spinnaker/fiat/blob/092eff250bc4f10ae6681d6174bf5a56d889c5bf/fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GithubTeamsUserRolesProvider.java#L328) in `fiat` uses the response headers from `RetrofitError` to log rate limiting information from GitHub. This method needs to be refactored to use `SpinnakerHttpException` but would lose access to the response headers without the changes in this PR.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -54,6 +56,10 @@ public int getResponseCode() {
     return response.getStatus();
   }
 
+  public List<Header> getHeaders() {
```
dbyron-sf, line 59: I'm a little torn about this.  Part of me wants to leave it like this.  Part of me is concerned that as we add support for retrofit2 (with the goal of removing retrofit1 to resolve CVEs) that we'll wish this wasn't here.

What do you think about using [MultiValueMap<String, String>](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/MultiValueMap.html) or [HttpHeaders](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/HttpHeaders.html) ?

dbyron-sf: If it's unsavory to including spring-web in kork-retrofit, maybe guava's [Multimap](https://guava.dev/releases/30.0-jre/api/docs/com/google/common/collect/Multimap.html) is better?

abe21412: I had a similar thought because I saw another pull request to add support for retrofit2. I'm okay with returning `org.springframework.http.HttpHeaders` instead.

dbyron-sf: Yes please.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -56,8 +55,10 @@ public int getResponseCode() {
     return response.getStatus();
   }
 
-  public List<Header> getHeaders() {
-    return response.getHeaders();
+  public HttpHeaders getHeaders() {
+    HttpHeaders headers = new HttpHeaders();
```
dbyron-sf, line 59: Store this the first time to avoid re-calculating every time?

dbyron-sf (approved):

PR #1026, opened 2023-03-20
feat(SpinnakerRetrofitErrorHandler): Wraps an exception of kind Retrofit CONVERSION

Luthan95:
assuming that a conversion error is unlikely to succeed on retry, So wraps an exception and set retry on failure to false


kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerRetrofitErrorHandler.java
```diff
@@ -56,6 +56,11 @@ public Throwable handleError(RetrofitError e) {
           retval.setRetryable(false);
         }
         return retval;
+        // assuming that a conversion error is unlikely to succeed on retry
+      case CONVERSION:
+        SpinnakerConversionException ex = new SpinnakerConversionException(e);
+        ex.setRetryable(false);
```
dbyron-sf, line 62: Let's move this logic into SpinnakerConversionException since I figure those are never retryable, no matter where they're constructed.

dbyron-sf:
handled by https://github.com/spinnaker/kork/pull/1099

PR #981, opened 2022-10-12
feat(spel): Add feature flag for DoNotEval SpEL.

armory-abedonik:
**Usage**

```
# orca-local.yml

expression:
  do-not-eval-spel:
    enabled: true
```

**References**:

* https://github.com/spinnaker/kork/pull/978
* https://github.com/spinnaker/orca/pull/4309

PR #1028, opened 2023-03-21
feat(doNoEval): default doNotEval SPeL expression to On

ciurescuraul:
Enabling “doNotEval” SPeL expression by default. This feature is disabled only when the flag is set explicitly to false.

The "doNotEval" SpEL helper makes it possible to skip SpEL evaluation in other SpEL helpers e.g. toJson as it is described in 
[Next release preview](https://spinnaker.io/docs/releases/next-release-preview/#donoteval-spel-helper) documentation. 

One reason is that a SpEL expression failure appears while using Terraformer to serialize data from a Terraform Plan execution.  The execution creates a JSON object instead of a JSON string.
The cause is that the SpEL processor in Spinnaker throws a null value if the SpEL expression has ${ in it, and is visible in Spinnaker's code [here](https://github.com/spinnaker/kork/blob/d895e2b820baadc0986dd4451e7b18f030a9a865/kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/ExpressionsSupport.java#L198)

Also, if the users are using an older Terraform version (<= 0.12), then it is also using the template provider from Terraform.
The template provider exposes data sources to use templates to generate strings for other Terraform resources or outputs.

❗️ HashiCorp has deprecated the template provider since v0.12 the announcement is [here](https://registry.terraform.io/providers/hashicorp/template/latest/docs)

Users should use the [template file](https://www.terraform.io/language/functions/templatefile?_ga=2.68579986.223689003.1655389310-554336950.1655389310) function instead

Relates to: [BOB-30735](https://armory.atlassian.net/browse/BOB-30735)

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 kotlinVersion=1.4.32
 org.gradle.parallel=true
-spinnakerGradleVersion=8.25.0
+spinnakerGradleVersion=8.26.0
 targetJava11=true
 includeRuntimes=actuator,core,eureka,retrofit,secrets-aws,secrets-gcp,stackdriver,swagger,tomcat,web
--- a/kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/config/ExpressionProperties.java
+++ b/kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/config/ExpressionProperties.java
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.kork.expressions.config;
 
 import lombok.Data;
+import lombok.experimental.Accessors;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 
 @Data
@@ -26,12 +27,8 @@ public class ExpressionProperties {
   private final FeatureFlag doNotEvalSpel = new FeatureFlag().setEnabled(true);
 
   @Data
+  @Accessors(chain = true)
   public static class FeatureFlag {
     private boolean enabled;
-
-    public FeatureFlag setEnabled(boolean enabled) {
-      this.enabled = enabled;
-      return this;
-    }
   }
 }
--- a/kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/config/ExpressionProperties.java
+++ b/kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/config/ExpressionProperties.java
@@ -27,6 +27,6 @@ public class ExpressionProperties {
 
   @Data
   public static class FeatureFlag {
-    private boolean enabled;
+    private boolean enabled = true;
   }
 }
--- a/kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/config/ExpressionProperties.java
+++ b/kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/config/ExpressionProperties.java
@@ -25,8 +25,13 @@ public class ExpressionProperties {
 
   private final FeatureFlag doNotEvalSpel = new FeatureFlag();
 
+  public FeatureFlag getDoNotEvalSpel() {
+    doNotEvalSpel.setEnabled(true);
+    return doNotEvalSpel;
+  }
+
   @Data
   public static class FeatureFlag {
-    private boolean enabled = true;
+    private boolean enabled;
   }
 }
--- a/kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/config/ExpressionProperties.java
+++ b/kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/config/ExpressionProperties.java
@@ -23,15 +23,15 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 @ConfigurationProperties(prefix = "expression")
 public class ExpressionProperties {
 
-  private final FeatureFlag doNotEvalSpel = new FeatureFlag();
-
-  public FeatureFlag getDoNotEvalSpel() {
-    doNotEvalSpel.setEnabled(true);
-    return doNotEvalSpel;
-  }
+  private final FeatureFlag doNotEvalSpel = new FeatureFlag().setEnabled(true);
 
   @Data
   public static class FeatureFlag {
     private boolean enabled;
+
+    public FeatureFlag setEnabled(boolean enabled) {
+      this.enabled = enabled;
+      return this;
+    }
   }
 }


ovidiupopa07 (approved):
lgtm

ovidiupopa07:
> Please update the PR description and commit message to `feat(doNoEval)` since this is a behavior change. But then also, what's the reasoning behind this change? Please include that as well in the PR description + commit message.
> 
> And then finally, I don't think it's quite time to do this. [This release note](https://github.com/spinnaker/spinnaker.io/blob/1566d78587759ed118765e21490b3f1a2b14cd98/content/en/docs/releases/next-release-preview/index.md#donoteval-spel-helper) says the flag is new for 1.30, but 1.30 hasn't been released yet. Let's wait until it has before merging this.

hi @dbyron-sf . The initial [PR](https://github.com/spinnaker/kork/pull/981) was back-ported to 1.28 and 1.29. The intention here is to enable this feature by default in 1.30 instead of manually opting in for it

dbyron-sf:
> > Please update the PR description and commit message to `feat(doNoEval)` since this is a behavior change. But then also, what's the reasoning behind this change? Please include that as well in the PR description + commit message.
> > And then finally, I don't think it's quite time to do this. [This release note](https://github.com/spinnaker/spinnaker.io/blob/1566d78587759ed118765e21490b3f1a2b14cd98/content/en/docs/releases/next-release-preview/index.md#donoteval-spel-helper) says the flag is new for 1.30, but 1.30 hasn't been released yet. Let's wait until it has before merging this.
> 
> hi @dbyron-sf . The initial [PR](https://github.com/spinnaker/kork/pull/981) was back-ported to 1.28 and 1.29. The intention here is to enable this feature by default in 1.30 instead of manually opting in for it. As per the [RFC](https://github.com/spinnaker/governance/pull/303) we should enable a feature in the next LTS (since it was back-ported to previous LTS releases)

OK, it helps to know the feature has already been released.  I don't see where https://github.com/spinnaker/governance/pull/303 says "we should enable a feature in the next LTS", but even if it does, I'd still like to know the specific reason for doing it for this flag.  Like, what are the consequences for changing this flag on existing spinnaker users/installations/pipelines/etc. ?  Who is it going to break vs. who is it going to help.  Or perhaps put another way, I'd like to see the release note describing this change.

As a bit of an aside, https://github.com/spinnaker/governance/pull/303 does say

> We propose that this change take effect immediately in the backport
> policy and relevant contributor documentation be updated to describe
> the facilities for feature flagging in the project.

but I don't think https://spinnaker.io/docs/community/contributing/code/releasing/#release-branch-patch-criteria has been updated.

kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/config/ExpressionProperties.java
```diff
@@ -27,6 +27,6 @@ public class ExpressionProperties {
 
   @Data
   public static class FeatureFlag {
-    private boolean enabled;
+    private boolean enabled = true;
```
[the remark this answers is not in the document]
ciurescuraul: this change has been addressed. thx @dbyron-sf and @ovidiupopa07  for the comments!

kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/config/ExpressionProperties.java
```diff
@@ -25,6 +25,11 @@ public class ExpressionProperties {
 
   private final FeatureFlag doNotEvalSpel = new FeatureFlag();
 
+  public FeatureFlag getDoNotEvalSpel() {
```
ovidiupopa07, line 28: I would prefer having a fluent setter in the `FeatureFlag` class like this
```
class FeatureFlag {
  private enabled;

  public void setEnabled(boolean enabled) {
     this.enabled = enabled'
     return this;
  }
}
```

that can be used in `ExpressionProperties` like this `private final FeatureFlag doNotEvalSpel = new FeatureFlag().setEnabled(true);`

ovidiupopa07 (changes requested):

dbyron-sf:
> I guess the worst issue that could happen is the pipeline will not work as expected.

This is a bit of an unsatisfying answer, but with SpEL expressions it may be the best we can do.  It does look like https://spinnaker.io/docs/releases/next-release-preview/#donoteval-spel-helper could use a tweak when this PR goes in.  Would you mind putting up a PR with those changes?  At that point I'm OK with this going in, though I still think it may not make 1.30.0.

ovidiupopa07:
> > I guess the worst issue that could happen is the pipeline will not work as expected.
> 
> This is a bit of an unsatisfying answer, but with SpEL expressions it may be the best we can do. It does look like https://spinnaker.io/docs/releases/next-release-preview/#donoteval-spel-helper could use a tweak when this PR goes in. Would you mind putting up a PR with those changes? At that point I'm OK with this going in, though I still think it may not make 1.30.0.

https://github.com/spinnaker/spinnaker.io/pull/317 -> PR for the docs

dbyron-sf:
Now that 1.30.0 is out, https://github.com/spinnaker/spinnaker.io/pull/317 could use an update @ovidiupopa07 , and I think this is good to go.

ciurescuraul:
Created new PR with updated docs for Release 1.31 -> [spinnaker/spinnaker.io#324](https://github.com/spinnaker/spinnaker.io/pull/324)

dbyron-sf (approved):

dbyron-sf:
@mergifyio update

dbyron-sf:
Looks like this needs a rebase, but otherwise it looks good to go to me.  You agree @ovidiupopa07 ?  I pushed the button too fast and merged https://github.com/spinnaker/spinnaker.io/pull/324 already...

gradle.properties
```diff
@@ -1,5 +1,5 @@
 kotlinVersion=1.4.32
 org.gradle.parallel=true
-spinnakerGradleVersion=8.25.0
+spinnakerGradleVersion=8.26.0
```
dbyron-sf, line 3: This seems fishy as the [HEAD of master](https://github.com/spinnaker/kork/blob/7e6e0d496f2bae8e9828ff84c7a5ac1844f05053/gradle.properties#L3) is already using 8.26.0.

ciurescuraul: > This seems fishy as the [HEAD of master](https://github.com/spinnaker/kork/blob/7e6e0d496f2bae8e9828ff84c7a5ac1844f05053/gradle.properties#L3) is already using 8.26.0.

Created new PR to eliminate any doubt ->  https://github.com/spinnaker/kork/pull/1030

dbyron-sf: Bummer...then we lose all the context.  Sure there's no way to bring this back to life?

ciurescuraul: Didn't realize at the moment what the impact would be. 
It doesn't seem to let me reopen this one,  [the new PR #1030](https://github.com/spinnaker/kork/pull/1030) has a reference to the [old PR #1028](https://github.com/spinnaker/kork/pull/1028). I hope this helps

PR #1020, opened 2023-02-14
fix(cve): Remove CVE-2022-23457 caused by esapi library version 2.1.0.1 that is resolved transitively trough spring-security-saml-dsl-core

ovidiupopa07:
**Before applying the constraint**
```
 ./gradlew gate-saml:dependencyInsight --dependency esapi                

> Task :gate-saml:dependencyInsight
org.owasp.esapi:esapi:2.1.0.1
   variant "compile" [
      org.gradle.status              = release (not requested)
      org.gradle.usage               = java-api
      org.gradle.libraryelements     = jar (compatible with: classes+resources)
      org.gradle.category            = library

      Requested attributes not found in the selected variant:
         org.gradle.dependency.bundling = external
         org.gradle.jvm.version         = 11
   ]

org.owasp.esapi:esapi:2.1.0.1
\--- org.springframework.security.extensions:spring-security-saml2-core:1.0.9.RELEASE
     +--- compileClasspath (requested org.springframework.security.extensions:spring-security-saml2-core)
     +--- io.spinnaker.kork:kork-bom:7.153.0
     |    \--- compileClasspath
     \--- org.springframework.security.extensions:spring-security-saml-dsl-core:1.0.5.RELEASE (requested org.springframework.security.extensions:spring-security-saml2-core:1.0.4.RELEASE)
          +--- compileClasspath (requested org.springframework.security.extensions:spring-security-saml-dsl-core)
          \--- io.spinnaker.kork:kork-bom:7.153.0 (*)

(*) - dependencies omitted (listed previously)

A web-based, searchable dependency report is available by adding the --scan option.
```

**After applying the constraint**
```
./gradlew gate-saml:dependencyInsight --dependency esapi

> Task :gate-saml:dependencyInsight
org.owasp.esapi:esapi:2.3.0.0
   variant "compile" [
      org.gradle.status              = release (not requested)
      org.gradle.usage               = java-api
      org.gradle.libraryelements     = jar (compatible with: classes+resources)
      org.gradle.category            = library

      Requested attributes not found in the selected variant:
         org.gradle.dependency.bundling = external
         org.gradle.jvm.version         = 11
   ]
   Selection reasons:
      - By constraint
      - Forced

org.owasp.esapi:esapi:2.3.0.0
\--- io.spinnaker.kork:kork-bom:0.1.0-SNAPSHOT
     \--- compileClasspath (requested io.spinnaker.kork:kork-bom:7.153.0)

[...]

PR #1038, opened 2023-04-04
fix(saml): revert esapi library change

jasonmcintosh:
See: https://github.com/ESAPI/esapi-java-legacy/issues/567
and
https://github.com/spinnaker/kork/pull/1020#issuecomment-1495005465

ESAPI on newer releases removed a class that opensaml referenced in a hard coded manner.  We need to bump opensaml to a later release & do a LOT more validation.


mattgogerly:
@Mergifyio backport release-1.30.x

PR #1049, opened 2023-05-23
fix(secrets): Make SecretEngine loading lazy

jvz:
This addresses a couple issues:

- SecretEngine implementations should be loaded later in the ApplicationContext lifecycle. As it was before, these were being registered as an incredibly high priority `BeanPostProcessor`. This should instead be registered as an incredibly low priority `BeanPostProcessor` in order to take advantage of other Spring features in secret engines and be registered before singleton beans are subsequently created (such as plugin configurations or configuration properties classes).
- This fixes numerous info messages logged at startup regarding the ineligibility of many beans to be processed by `BeanPostProcessor` classes.
- Additionally, this fixes `SecretAwarePropertySource` to use generics properly along with delegating appropriate methods to its wrapped `EnumerablePropertySource<T>` instance.


jvz:
A nice side effect here is that you should be able to set up HTTP tracing, metrics, etc., all via Spring Boot features (or similar approaches) rather than having to set anything up manually. For example, you can use the `@Cacheable` annotation inside a `SecretEngine` since the support for that will be loaded before anything in the secrets API is even touched. The laziness introduced in `SecretEngineRegistry` helps most of the way there, though the updated `BeanPostProcessor` ensures that it's loaded at the right time, too.

nemesisOsorio (approved):

kork-secrets/src/test/java/com/netflix/spinnaker/kork/secrets/SecretAwarePropertySourceTest.java
```diff
@@ -12,42 +13,33 @@
...
-import org.junit.Rule;
 import org.junit.Test;
-import org.junit.rules.ExpectedException;
-import org.springframework.core.env.EnumerablePropertySource;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+import org.springframework.core.env.MapPropertySource;
 
+@RunWith(MockitoJUnitRunner.class)
```
dbyron-sf, line 22: I'm a little fuzzy here, but isn't `@ExtendWith(MockitoExtension.class)` the more junit5/jupiter-y way?

jvz: Yes, that would be the case, though this test was already in JUnit 4. I updated this as Mockito warns about using `initMocks` nowadays because you're supposed to close it after the tests, hence the recommendation on using the JUnit runner (v4) or extension (v5) to handle setup and cleanup automatically.

dbyron-sf: Fair enough.

dbyron-sf (approved):

jvz:
Damn, now there's a formatting violation.

PR #1044, opened 2023-04-12
Upgrade okhttp 4.x

droustchev-pd:
This is motivated by https://github.com/square/okhttp/issues/6738 and is based on the official upgrade instructions: https://square.github.io/okhttp/changelogs/upgrading_to_okhttp_4/

Any guidance on additional tests/changes desired would be appreciated, since this is my first time contributing to this project.

PR #1050, opened 2023-06-05
Cryptography module and OkHttp3 config updates

jvz:
I've got two inter-related changes to contribute:

1. Addition of a `kork-crypto` module for common cryptographic operations. Some initial goodies there include a PEM key/cert file parser (no more need to specify a key algorithm when parsing PEM files), a refreshable key store (technically a refreshable `KeyManager`), a builder for `SecureRandom` instances via DRBG, a function for password loading, support for using a separate password for a key store versus a private key, and an updated version of Bouncycastle.
2. Updates to `kork-web` to support refreshable mTLS credentials with an OkHttp3 client. This is an opt-in feature flag for now and relies on a new API for customizing OkHttp stuff similar to the Spring classes `WebClientCustomizer` and `WebClient.Builder`.

### DIFF
--- a/kork-credentials/src/main/java/com/netflix/spinnaker/credentials/CredentialsTypeBaseConfiguration.java
+++ b/kork-credentials/src/main/java/com/netflix/spinnaker/credentials/CredentialsTypeBaseConfiguration.java
@@ -26,6 +26,7 @@ import java.util.Optional;
 import javax.annotation.Nullable;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
+import org.springframework.beans.factory.BeanFactoryUtils;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.config.ConstructorArgumentValues;
 import org.springframework.beans.factory.support.DefaultListableBeanFactory;
@@ -171,7 +172,8 @@ public class CredentialsTypeBaseConfiguration<
   protected <T> Optional<T> getParameterizedBean(
       ApplicationContext applicationContext, Class<T> paramClass, Class<?>... generics) {
     ResolvableType resolvableType = ResolvableType.forClassWithGenerics(paramClass, generics);
-    String[] beanNames = applicationContext.getBeanNamesForType(resolvableType);
+    String[] beanNames =
+        BeanFactoryUtils.beanNamesForTypeIncludingAncestors(applicationContext, resolvableType);
     if (beanNames.length == 1) {
       return Optional.of((T) applicationContext.getBean(beanNames[0]));
     }
--- a/kork-credentials/src/test/java/com/netflix/spinnaker/credentials/CredentialsTypeBaseConfigurationTest.java
+++ b/kork-credentials/src/test/java/com/netflix/spinnaker/credentials/CredentialsTypeBaseConfigurationTest.java
@@ -25,6 +25,7 @@ import lombok.Data;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.mockito.Mockito;
+import org.springframework.beans.factory.BeanFactoryUtils;
 import org.springframework.context.support.StaticApplicationContext;
 
 public class CredentialsTypeBaseConfigurationTest {
@@ -117,6 +118,24 @@ public class CredentialsTypeBaseConfigurationTest {
     config.getCredentialsLoader().load();
 
     String[] beanNames = context.getBeanNamesForType(CredentialsRepository.class);
+    assertThat(beanNames).hasSize(1);
+    assertThat(beanNames[0]).isEqualTo("customRepository");
+  }
+
+  @Test
+  void testOverrideBeanInParentContext() {
+    var childContext = new StaticApplicationContext(context);
+    config = new CredentialsTypeBaseConfiguration<>(childContext, props);
+    TestCredentialsRepository repository =
+        new TestCredentialsRepository(CREDENTIALS_TYPE, new NoopCredentialsLifecycleHandler<>());
+    context.getBeanFactory().registerSingleton("customRepository", repository);
+    config.afterPropertiesSet();
+    // This test runner ignores PostConstruct annotations
+    config.getCredentialsLoader().load();
+
+    String[] beanNames =
+        BeanFactoryUtils.beanNamesForTypeIncludingAncestors(
+            childContext, CredentialsRepository.class);
     assertThat(beanNames).hasSize(1);
     assertThat(beanNames[0]).isEqualTo("customRepository");
   }
--- a/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/RetrofitException.java
+++ b/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/RetrofitException.java
@@ -18,6 +18,7 @@ package com.netflix.spinnaker.kork.retrofit.exceptions;
 
 import java.io.IOException;
 import java.lang.annotation.Annotation;
+import java.util.Objects;
 import okhttp3.ResponseBody;
 import retrofit2.Converter;
 import retrofit2.Response;
@@ -55,6 +56,10 @@ public class RetrofitException extends RuntimeException {
   RetrofitException(String message, Response response, Throwable exception, Retrofit retrofit) {
     super(message, exception);
     this.response = response;
+    if (response != null) {
+      // Fail fast instead of checking for null in e.g. getErrorBodyAs.
+      Objects.requireNonNull(response.errorBody());
+    }
     this.retrofit = retrofit;
   }
 
@@ -70,9 +75,10 @@ public class RetrofitException extends RuntimeException {
    *     the specified {@code type}.
    */
   public <T> T getErrorBodyAs(Class<T> type) {
-    if (response == null || response.errorBody() == null) {
+    if (response == null) {
       return null;
     }
+
     Converter<ResponseBody, T> converter = retrofit.responseBodyConverter(type, new Annotation[0]);
     try {
       return converter.convert(response.errorBody());
--- a/kork-retrofit/src/test/java/com/netflix/spinnaker/kork/retrofit/exceptions/RetrofitExceptionTest.java
+++ b/kork-retrofit/src/test/java/com/netflix/spinnaker/kork/retrofit/exceptions/RetrofitExceptionTest.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2023 OpsMx, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.kork.retrofit.exceptions;
+
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import okhttp3.MediaType;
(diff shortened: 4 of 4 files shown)


kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttpClientComponents.java
```diff
@@ -0,0 +1,277 @@
...
+    var retryOnConnectionFailure = clientProperties.isRetryOnConnectionFailure();
+    return builder ->
+        builder
+            .connectTimeout(connectTimeout)
+            .readTimeout(readTimeout)
+            .retryOnConnectionFailure(retryOnConnectionFailure);
+  }
+
+  @Bean
+  public OkHttpClientCustomizer httpLoggingCustomizer() {
```
jvz, line 257: Comments around this interceptor mention it should come last, so I put it here at the bottom.

kork-web/src/main/groovy/com/netflix/spinnaker/config/RetrofitConfiguration.groovy
```diff
@@ -19,13 +19,11 @@ package com.netflix.spinnaker.config
 import com.netflix.spinnaker.retrofit.Retrofit2ConfigurationProperties
 import com.netflix.spinnaker.retrofit.RetrofitConfigurationProperties
 import okhttp3.logging.HttpLoggingInterceptor
-import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
 import retrofit.RestAdapter
 
 @Configuration
-@EnableConfigurationProperties([RetrofitConfigurationProperties, Retrofit2ConfigurationProperties])
```
jvz, line 28: Moved to `OkHttpClientComponents` as IntelliJ is really bad at parsing Spring+Groovy things.

spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -198,8 +198,8 @@ dependencies {
     api("net.logstash.logback:logstash-logback-encoder:4.11")
     api("net.minidev:json-smart:2.4.1") // TODO: remove this with upgrade of spring-boot version to 2.5.0 or above
     api("org.apache.commons:commons-exec:1.3")
-    api("org.bouncycastle:bcpkix-jdk15on:${versions.bouncycastle}")
-    api("org.bouncycastle:bcprov-jdk15on:${versions.bouncycastle}")
+    api("org.bouncycastle:bcpkix-jdk18on:${versions.bouncycastle}")
```
jvz, line 201: Note that `gate-x509.gradle` requires an update corresponding with this to use the new module name.

dbyron-sf:
Thanks much for this @jvz !  Any chance you know how it plays with https://github.com/spinnaker/kork/pull/1044 ?

kork-crypto/kork-crypto.gradle
```diff
@@ -0,0 +1,33 @@
...
+  api project(":kork-annotations")
+
+  implementation "org.bouncycastle:bcpkix-jdk18on"
+  implementation "org.springframework:spring-aop"
+
+  testImplementation "org.springframework.boot:spring-boot-starter-test"
+}
+
+test {
+  useJUnitPlatform()
```
dbyron-sf, line 32: Don't think we need this given [this in build.gradle](https://github.com/spinnaker/kork/blob/4f71ac098c3b12c9aa8e3687a2d2c3b888fc1491/build.gradle#L34).

kork-crypto/kork-crypto.gradle
```diff
@@ -0,0 +1,33 @@
...
+
+apply plugin: "java-library"
+apply from: "$rootDir/gradle/lombok.gradle"
+
+dependencies {
+  api(platform(project(":spinnaker-dependencies")))
+
+  api project(":kork-annotations")
+
+  implementation "org.bouncycastle:bcpkix-jdk18on"
```
dbyron-sf, line 25: [spinnaker-dependencies](https://github.com/spinnaker/kork/blob/4f71ac098c3b12c9aa8e3687a2d2c3b888fc1491/spinnaker-dependencies/spinnaker-dependencies.gradle#L201-L202) mentions `bcpkix-jdk15on`.  Not sure what's bringing in `org.bouncycastle:bcpkix-jdk18on` (i.e. why it works without a version).  Better to add it to spinnaker-dependencies?

jvz: Starting from BC 1.73, they updated how they package Java-specific versions: https://bouncycastle.org/latest_releases.html

I'd rather not have duplicate copies of BC floating around!

jvz:
> Thanks much for this @jvz ! Any chance you know how it plays with #1044 ?

Not really, no, unless they've maintained a v3-compatible API there. I didn't bother porting this to v2 since that's deprecated, but the underlying TLS code can certainly be reused for OkHttp 4. Speaking of which, note that OkHttp v5 is in development (there in the 5.0.0 alpha releases stage). Even Spring Framework 6 only references OkHttp v3 in any place, but it's in an obscure API (https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/client/OkHttp3ClientHttpRequestFactory.html).

kork-crypto/kork-crypto.gradle
jvz: Updated (with a rebase into the first commit).

kork-crypto/kork-crypto.gradle
dbyron-sf: Fair enough.  Better to reference `bcpkix-jdk15on` here until we upgrade to 1.73 then / switch to bcpkix-jdk18on?  Looks like we're on [1.70 at the moment](https://github.com/spinnaker/kork/blob/4f71ac098c3b12c9aa8e3687a2d2c3b888fc1491/spinnaker-dependencies/spinnaker-dependencies.gradle#L11).

kork-crypto/src/main/java/com/netflix/spinnaker/kork/crypto/RefreshableX509Identity.java
```diff
@@ -0,0 +1,53 @@
...
+      throw new UncheckedIOException(e);
+    }
+  }
+
+  @Override
+  protected boolean requiresRefresh() {
+    Instant lastLoaded = identitySource.getLastLoaded();
+    Instant lastModified = identitySource.getLastModified();
+    Instant expiresAt = identitySource.getExpiresAt();
+    return lastLoaded.isBefore(lastModified) || Instant.now().isAfter(expiresAt);
```
dbyron-sf, line 51: this is a nit, but is the current time is expiresAt, we need a refresh, right?
```suggestion
    return lastLoaded.isBefore(lastModified) || !Instant.now().isBefore(expiresAt);
```

jvz: I suppose in theory, but if the file hasn't been modified at this point, then it's unlikely to be modified in the following second. I'd apply your suggestion, but I'm not sure on the best way to do it from the UI that won't prevent this from merging due to improper git hygiene :D

dbyron-sf: no worries.

kork-crypto/kork-crypto.gradle
jvz: Would you rather the BC upgrade be done independent of this PR? It's starting in 1.71 where this new module is available.

kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttp3ClientConfiguration.groovy
```diff
@@ -48,40 +48,56 @@ import java.util.concurrent.TimeUnit
...
     this.retrofit2LogLevel = retrofit2LogLevel
     this.spinnakerRequestHeaderInterceptor = spinnakerRequestHeaderInterceptor
+    this.httpClientBuilderFactory = httpClientBuilderFactory
   }
 
   public OkHttp3ClientConfiguration(OkHttpClientConfigurationProperties okHttpClientConfigurationProperties,
                                     OkHttp3MetricsInterceptor okHttp3MetricsInterceptor) {
     this.okHttpClientConfigurationProperties = okHttpClientConfigurationProperties
     this.okHttp3MetricsInterceptor = okHttp3MetricsInterceptor
+    retrofit2LogLevel = null
```
dbyron-sf, line 80: nit, and lingering from before this PR...cleaner to call
```
this(okHttpClientConfigurationProperties, okHttp3MetricsInterceptor, null, null, { new OkHttpClient.Builder() })
```
and then in the next one, call this one?

jvz: Ah good idea. In theory, I'd love to remove the other constructors, but other code is still using it, and being a Groovy class, it's trickier to find usages of it.

kork-web/src/main/java/com/netflix/spinnaker/config/okhttp3/RefreshableOkHttpClientBuilderProvider.java
```diff
@@ -0,0 +1,40 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.core.annotation.Order;
+import org.springframework.stereotype.Component;
+
+@Component
+@Order(0)
+@ConditionalOnProperty("ok-http-client.refreshable-keys.enabled")
+@NonnullByDefault
+@RequiredArgsConstructor
+public class RefreshableOkHttpClientBuilderProvider implements OkHttpClientBuilderProvider {
```
dbyron-sf, line 33: Possible to add a test for this?  Or maybe there is one and I missed it.

spinnaker-dependencies/spinnaker-dependencies.gradle
dbyron-sf: Mind making a draft PR in gate so it's ready to go when this goes in?

kork-crypto/kork-crypto.gradle
dbyron-sf: I would, in case it causes problems with existing code.  If it's too painful to separate it's OK to happen here.

kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttp3ClientConfiguration.groovy
dbyron-sf: Agreed.

kork-web/src/main/java/com/netflix/spinnaker/config/okhttp3/RefreshableOkHttpClientBuilderProvider.java
jvz: Tested in kork-crypto. I could write another test like that one which uses OkHttp instead of JDK HttpClient.

kork-crypto/kork-crypto.gradle
jvz: I can update that most likely. Will need to verify I'm not relying on anything new first.

dbyron-sf: Thanks!

kork-web/src/main/java/com/netflix/spinnaker/config/okhttp3/RefreshableOkHttpClientBuilderProvider.java
dbyron-sf: Sorry if I'm being dense...I don't see a mention of RefreshableOkHttpClientBuilderProvider in kork-crypto.

kork-crypto/kork-crypto.gradle
jvz: Seems like I was referring to one updated bit in `SHAKEDigest`, but that's not a blocker.

kork-web/src/main/java/com/netflix/spinnaker/config/okhttp3/RefreshableOkHttpClientBuilderProvider.java
jvz: I'm going to add a test for this class in particular. I have a general test for the SSLContext bit in the crypto module that uses different client and server APIs.

PR #1051, opened 2023-06-05
fix(credentials): Use bean factory hierarchy in config

jvz:
This fixes a bug in `CredentialsTypeBaseConfiguration` where only beans from the current `ApplicationContext` would be checked for various optional credentials-related bean definitions. When a plugin is installed containing credentials code, then the plugin's `ApplicationContext` is used here (a child context from the service's `ApplicationContext`), and looking up beans on a context only considers the locally defined beans unless `BeanFactoryUtils` is used to query the entire hierarchy. As credentials storage and account management for artifact accounts relies on these optional beans being picked up, this fixes the bean resolution to consider the entire hierarchy.


dbyron-sf (approved):

dbyron-sf:
@mergifyio update

mattgogerly (approved):

PR #1055, opened 2023-06-09
test(retrofit): demonstrate the behavior of Spinnaker*Exception when constructed from a RetrofitException

SheetalAtre:
test(retrofit): demonstrate the behavior of Spinnaker*Exception when constructed from a RetrofitException

* The test to check the response custom message and rawMessage, depend on storing the parsed response body as a map, and will get added in that PR.

### DIFF
--- a/kork-retrofit/src/test/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerServerExceptionTest.java
+++ b/kork-retrofit/src/test/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerServerExceptionTest.java
@@ -16,18 +16,26 @@
 
 package com.netflix.spinnaker.kork.retrofit.exceptions;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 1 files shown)


kork-retrofit/src/test/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerServerExceptionTest.java
```diff
@@ -16,18 +16,26 @@
...
+import okhttp3.ResponseBody;
+import org.apache.groovy.json.internal.CharBuf;
 import org.junit.jupiter.api.Test;
+import org.springframework.http.HttpStatus;
 import retrofit.RetrofitError;
 import retrofit.client.Response;
+import retrofit2.Retrofit;
+import retrofit2.converter.jackson.JacksonConverterFactory;
+
+import static org.junit.jupiter.api.Assertions.*;
```
dbyron-sf, line 34: Please avoid star imports.

PR #1065, opened 2023-06-19
refactor(retrofit): Move rawMessage and response body map to SpinnakerHttpException

SheetalAtre:
refactor(retrofit): Move rawMessage and response body map to SpinnakerHttpException as network / unexpected do not have a json response body

The ErrorHandlingExecutorCallAdapterFactory::execute() does not return/add a retrofit2 Response body if either of:
* network error like IOException(SpinnakerNetworkException) during server communication occurs or
* unexpected error like RuntimeException(SpinnakerServerException) while creating/decoding the request/response is thrown

### DIFF
--- a/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerServerException.java
+++ b/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerServerException.java
@@ -16,13 +16,8 @@
 
 package com.netflix.spinnaker.kork.retrofit.exceptions;
 
-import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
-import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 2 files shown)


kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerServerException.java
```diff
@@ -16,44 +16,25 @@
...
 
   /**
-   * A message derived from a RetrofitError's response body, or null if a custom message has been
-   * provided.
-   */
-  private final String rawMessage;
-
-  /**
-   * Parses the message from the {@link RetrofitErrorResponseBody} of a {@link RetrofitError}.
+   * Parses the message from the {@link RetrofitError}.
```
dbyron-sf, line 28: This isn't actually true anymore.

dbyron-sf: Let's deal with this in a followup PR.

PR #1069, opened 2023-06-29
feat(artifacts): Add ArtifactStore to Kork

xibz:
This commit adds a new feature called ArtifactStore. This allows for storage of embedded/base64 types to some storage, eg S3. So instead of having duplication of some artifact in a context, it'll now be referenced by a artifact URI.

Also, there's a nice README that explains various settings as well as how to setup a local S3 for testing artifact storage

[README link](kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/README.md )

The related PRs
https://github.com/spinnaker/clouddriver/pull/5976
https://github.com/spinnaker/orca/pull/4481
https://github.com/spinnaker/rosco/pull/998
https://github.com/spinnaker/gate/pull/1674

Further to enable this feature can be enabled via feature flag

```yaml
artifactStore:
  enabled: true
  applicationsRegex: "only-these-applications-*" # optional regex to enable certain regex application names to have this enabled
  s3:
    enabled: true
    bucket: s3-artifact-bucket
```

We've also allowed for certain applications to have it enabled with shortens the blast radius of testing.

Also for some metrics, we saw large pipelines be reduced by 70%. It really comes down to how often an artifact is referenced in each stage. Simple smaller pipelines were reduced by 30%.

Screen Shot 2023-06-29 at 12 56 15 AM

Without any formatting, the JSON file is 26,960,606 bytes in length, out of which over 20M are for the reference fields of embedded artifacts (almost exactly 75%).

---

# TODO

- [x] Create PR for Gate
- [ ] Create PR for Deck

### DIFF
--- a/kork-artifacts/kork-artifacts.gradle
+++ b/kork-artifacts/kork-artifacts.gradle
@@ -8,11 +8,9 @@ dependencies {
   implementation project(":kork-exceptions")
   implementation project(":kork-security")
   implementation "com.fasterxml.jackson.core:jackson-databind"
-  implementation platform('software.amazon.awssdk:bom')
   api 'software.amazon.awssdk:s3'
   api 'software.amazon.awssdk:sts'
   implementation "org.apache.httpcomponents:httpclient"
-  implementation 'software.amazon.awssdk:s3'
   implementation "org.springframework.boot:spring-boot-autoconfigure"
   implementation "org.springframework.security:spring-security-core"
   implementation 'org.apache.logging.log4j:log4j-api'
--- a/kork-artifacts/kork-artifacts.gradle
+++ b/kork-artifacts/kork-artifacts.gradle
@@ -5,11 +5,23 @@ dependencies {
   api(platform(project(":spinnaker-dependencies")))
 
   implementation project(":kork-annotations")
+  implementation project(":kork-exceptions")
+  implementation project(":kork-security")
   implementation "com.fasterxml.jackson.core:jackson-databind"
+  implementation platform('software.amazon.awssdk:bom')
+  api 'software.amazon.awssdk:s3'
+  api 'software.amazon.awssdk:sts'
+  implementation "org.apache.httpcomponents:httpclient"
+  implementation 'software.amazon.awssdk:s3'
+  implementation "org.springframework.boot:spring-boot-autoconfigure"
+  implementation "org.springframework.security:spring-security-core"
+  implementation 'org.apache.logging.log4j:log4j-api'
+  implementation 'org.apache.logging.log4j:log4j-core'
   api "com.hubspot.jinjava:jinjava"
 
   testImplementation "org.assertj:assertj-core"
   testImplementation "org.junit.jupiter:junit-jupiter-api"
   testImplementation "org.junit.jupiter:junit-jupiter-params"
+  testImplementation "org.mockito:mockito-core"
   testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine"
 }
--- a/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/README.md
+++ b/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/README.md
@@ -1,14 +1,14 @@
 # Artifact Storage
 
-Artifact Storage is a feature that allows for embedded aritfacts to be
+Artifact Storage is a feature that allows for embedded artifacts to be
 persisted to some storage, eg the S3ArtifactStore. Spinnaker keeps a history,
 which is called the pipeline context, that contains everything that an
 execution had done. Pipeline contexts can be very large, especially with
-complicated pipeines, and size can be further increased when users have large
+complicated pipelines, and size can be further increased when users have large
 artifacts. Spinnaker will duplicate
 these artifacts whenever any stage uses any of those artifacts. Using an
 artifact store reduces this overhead by providing a reference link of,
-`ref://<spinnaker-application/<content-hash>`. This reduces the context size
+`ref://<spinnaker-application>/<content-hash>`. This reduces the context size
 tremendously, but will vary depending on the size of the pipeline, as well as
 how that artifact is used, but we've seen improvements of 80% for some
 pipelines.
@@ -81,7 +81,7 @@ check if some `String` is a remote base64 URI, and if it is, retrieve it.
 To enable artifact storage, simple add this to your `spinnaker-local.yml` file
 
 ```yaml
-artifactStore:
+artifact-store:
   enabled: true
   s3:
     enabled: true
@@ -93,11 +93,11 @@ artifactStore:
 ### S3
 
 [S3](https://aws.amazon.com/s3/) is an object store provided by AWS. The
-currently S3ArtifactStore implementation provides various ways to authenticate
+current S3ArtifactStore implementation provides various ways to authenticate
 against AWS.
 
 ```yaml
-artifactStore:
+artifact-store:
   enabled: true
   s3:
     enabled: true
@@ -108,7 +108,7 @@ artifactStore:
 
 While the implementation is S3 specific, this does not limit usages of other S3
 compatible storage engines. For example, something like
-[SeaweedFS](https://github.com/seaweedfs/seaweedfs) can be used to test locally
+[SeaweedFS](https://github.com/seaweedfs/seaweedfs) can be used to test locally.
 with
 
 ## Local Testing
@@ -119,7 +119,7 @@ To test the artifact store locally, we will use SeaweedFS. To start the storage 
 Next enable the configuration
 
 ```yaml
-artifactStore:
+artifact-store:
   enabled: true
   s3:
     enabled: true
--- a/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/README.md
+++ b/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/README.md
@@ -36,7 +36,7 @@ pipelines.
 
 
 Artifact storage is divided into two operations of get and store, and there are
-primarily two services that utilizes each of these operations. Further the
+primarily two services that utilize each of these operations. Further the
 artifact storage system relies on Spring's (de)serializing to call these
 operations to limit the amount of code changes needed within these services.
 
@@ -44,10 +44,9 @@ When bootstrapping Spring we add in custom bean serializers and deserializers to
 handle storage or retrieval of an artifact.
 
 Rosco is primarily used for baking artifacts which will generate something
-deployable. When Spring sends a request back to Orca for a bake request, the
-custom serializer injected at Orca startup, will use the artifact storage and
-store the artifact, and instead return a `remote/base64` artifact instead of the
-usual `embedded/base64`.
+deployable.  When Rosco responds to a bake request, the custom serializer
+injected in Rosco at startup stores the artifact and returns a `remote/base64`
+artifact instead of the usual `embedded/base64`.
 
 Clouddriver, for this document, handles mostly deployment and has some endpoints
 regarding artifacts. It does do a little more than this, but we only care about
--- a/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/README.md
+++ b/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/README.md
@@ -1,0 +1,130 @@
+# Artifact Storage
+
+Artifact Storage is a feature that allows for embedded aritfacts to be
+persisted to some storage, eg the S3ArtifactStore. Spinnaker keeps a history,
+which is called the pipeline context, that contains everything that an
+execution had done. Pipeline contexts can be very large, especially with
+complicated pipeines, and size can be further increased when users have large
+artifacts. Spinnaker will duplicate
+these artifacts whenever any stage uses any of those artifacts. Using an
+artifact store reduces this overhead by providing a reference link of,
+`ref://<spinnaker-application/<content-hash>`. This reduces the context size
+tremendously, but will vary depending on the size of the pipeline, as well as
+how that artifact is used, but we've seen improvements of 80% for some
+pipelines.
+
+## Architecture
+
+                                 +-----------+
+                                 |           |
+                                 |   Orca    |
+                                 |           |
+                                 +-----------+
+                                   |        ^
+                                   |        | (outgoing artifact compressed)
+                                   |        +----------------------+
+                                   | (bake request)                |
+                                   +---------------------+         |
+                                                         v         |
+        +---------------+            (fetch)           +-------------+
+        |               |<-----------------------------|             |
+        |  Clouddriver  |                              |  Rosco      |
+        |               |                              |             |
+        |  s3 get       |   (full artifact returned)   |  s3 stores  |
+        |  artifacts    |----------------------------->|  artifacts  |
+        +---------------+                              +-------------+
+
+
+Artifact storage is divided into two operations of get and store, and there are
+primarily two services that utilizes each of these operations. Further the
+artifact storage system relies on Spring's (de)serializing to call these
+operations to limit the amount of code changes needed within these services.
+
+When bootstrapping Spring we add in custom bean serializers and deserializers to
+handle storage or retrieval of an artifact.
+
+Rosco is primarily used for baking artifacts which will generate something
+deployable. When Spring sends a request back to Orca for a bake request, the
+custom serializer injected at Orca startup, will use the artifact storage and
+store the artifact, and instead return a `remote/base64` artifact instead of the
+usual `embedded/base64`.
+
+Clouddriver, for this document, handles mostly deployment and has some endpoints
+regarding artifacts. It does do a little more than this, but we only care about
+these two particular operations. When any request comes in, Spring will use its
+custom deserializers to expand any artifact in its payload since any request
+with artifacts, probably wants to do some operation on those artifacts. Further
+Clouddriver also allows for fetching of artifacts. Orca and Rosco both make
+calls to the `/artifact/fetch` endpoint. Where Rosco uses it to fetch an
+artifact to be baked, and Orca uses it primarily when dealing with deploy
+manifests. When a request is sent to the fetch endpoint, Clouddriver will always
+return the full `embedded/base64` artifact back to the service. It is up to the
+service receiving the artifact to compress it. Luckily, for Orca, we don't have
+to worry about compression, since this no longer becomes an artifact, but a
+manifest instead.
+
+Orca is a special case as it mostly does orchestration, but does cause some
+artifacts to be duplicated when handling the expected artifact logic. We inject
+some logic to handle the duplication along with ensuring that matching against
+expected artifacts still works properly. So if a `embedded/base64` type needs to
+match against a `remote/base64` type, Orca will use the artifact store to
+retrieve that artifact, and do the comparison. In addition, Orca will store any
+expected artifacts, to limit the context size.
+
+Orca also handles SpEL evaluation which means our new `remote/base64` type
+should be backwards compatible with existing pipelines. To ensure this, we
+utilized the Spring converters, and injected our own custom converter that will
+check if some `String` is a remote base64 URI, and if it is, retrieve it.
+
+## Configuration
+
+To enable artifact storage, simple add this to your `spinnaker-local.yml` file
+
+```yaml
+artifactStore:
+  enabled: true
+  s3:
+    enabled: true
+    bucket: some-artifact-store-bucket
+```
+
+## Storage Options
+
+### S3
+
+[S3](https://aws.amazon.com/s3/) is an object store provided by AWS. The
+currently S3ArtifactStore implementation provides various ways to authenticate
+against AWS.
+
+```yaml
+artifactStore:
+  enabled: true
+  s3:
+    enabled: true
+    profile: dev # if you want to authenticate using a certain profile
+    region: us-west-2 # allows for specified regions
+    bucket: some-artifact-store-bucket
+```
+
+While the implementation is S3 specific, this does not limit usages of other S3
+compatible storage engines. For example, something like
+[SeaweedFS](https://github.com/seaweedfs/seaweedfs) can be used to test locally
+with
+
+## Local Testing
+
+To test the artifact store locally, we will use SeaweedFS. To start the storage simply run
+`docker run -p 8333:8333 chrislusf/seaweedfs server -s3`
+
+Next enable the configuration
+
+```yaml
+artifactStore:
+  enabled: true
+  s3:
+    enabled: true
+    url: http://localhost:8333 # this URL will be used to make S3 API requests to
+    bucket: some-artifact-store-bucket
+```
+
+Start Spinnaker, and you should see reference links in your pipeline contexts.
--- a/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfiguration.java
+++ b/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfiguration.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2023 Apple Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.artifacts.artifactstore;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.spinnaker.kork.artifacts.artifactstore.s3.S3ArtifactStore;
+import java.net.URI;
+import java.util.Optional;
+import lombok.extern.log4j.Log4j2;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.access.PermissionEvaluator;
+import software.amazon.awssdk.auth.credentials.*;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.S3ClientBuilder;
+
+@Configuration
+@ConditionalOnExpression("${artifact-store.enabled:false}")
+@EnableConfigurationProperties(ArtifactStoreConfigurationProperties.class)
+@Log4j2
+public class ArtifactStoreConfiguration {
+  /**
+   * this is strictly used due to Spring and Jackson not behaving nicely together.
+   * Unfortunately, @JsonDeserializer will construct its own deserializer utilizing beans and thus
+   * not using the object mapper we want to use
+   */
+  @Bean
+  public ObjectMapper artifactObjectMapper() {
+    return new ObjectMapper();
+  }
+
+  @Bean
+  public ArtifactStoreURIBuilder artifactStoreURIBuilder() {
+    return new ArtifactStoreURISHA256Builder();
+  }
+
+  @Bean
+  @ConditionalOnExpression("${artifact-store.s3.enabled:false}")
+  public ArtifactStore s3ArtifactStore(
+      Optional<PermissionEvaluator> permissionEvaluator,
+      ArtifactStoreConfigurationProperties properties,
+      @Qualifier("artifactS3Client") S3Client s3Client,
+      ArtifactStoreURIBuilder artifactStoreURIBuilder) {
+
+    if (permissionEvaluator.isEmpty()) {
+      log.warn(
+          "PermissionEvaluator is not present. This means anyone will be able to access any artifact in the store.");
+    }
+
+    ArtifactStore storage =
+        new S3ArtifactStore(
+            s3Client,
+            permissionEvaluator.orElse(null),
+            properties.getS3().getBucket(),
+            artifactStoreURIBuilder,
+            properties.getApplicationsRegex());
+
+    ArtifactStore.setInstance(storage);
+    return storage;
+  }
+
+  @Bean
+  @ConditionalOnExpression("${artifact-store.s3.enabled:false}")
+  public S3Client artifactS3Client(ArtifactStoreConfigurationProperties properties) {
+    S3ClientBuilder builder = S3Client.builder();
+    ArtifactStoreConfigurationProperties.S3ClientConfig config = properties.getS3();
+
+    // Overwriting the URL is primarily used for S3 compatible object stores
+    // like seaweedfs
+    if (config.getUrl() != null) {
+      builder =
+          builder
+              .credentialsProvider(getCredentialsProvider(config))
+              .forcePathStyle(config.isForcePathStyle())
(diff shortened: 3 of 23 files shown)


kork-artifacts/src/test/java/com/netflix/spinnaker/kork/artifacts/artifactstore/EmbeddedArtifactSerializerTest.java
```diff
@@ -0,0 +1,100 @@
...
+    for (TestCase testCase : cases) {
+      ArtifactStore storage = Mockito.mock(S3ArtifactStore.class);
+      when(storage.store(Mockito.any())).thenReturn(testCase.mockArtifact);
+
+      EmbeddedArtifactSerializer serializer =
+          new EmbeddedArtifactSerializer(new ObjectMapper(), storage);
+      ObjectMapper objectMapper = new ObjectMapper();
+      SimpleModule module = new SimpleModule();
+      module.addSerializer(Artifact.class, serializer);
+      objectMapper.registerModule(module);
```
nemesisOsorio, line 94: I think we can move these lines outside of the `for`

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+Artifact Storage is a feature that allows for embedded aritfacts to be
+persisted to some storage, eg the S3ArtifactStore. Spinnaker keeps a history,
+which is called the pipeline context, that contains everything that an
+execution had done. Pipeline contexts can be very large, especially with
+complicated pipeines, and size can be further increased when users have large
+artifacts. Spinnaker will duplicate
+these artifacts whenever any stage uses any of those artifacts. Using an
+artifact store reduces this overhead by providing a reference link of,
+`ref://<spinnaker-application/<content-hash>`. This reduces the context size
```
dbyron-sf, line 11:
```suggestion
`ref://<spinnaker-application>/<content-hash>`. This reduces the context size
```

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/README.md
```diff
@@ -0,0 +1,67 @@
...
+    enabled: true
+    bucket: spinnaker-qa-artifact-store-bucket
+```
+
+## Storage Options
+
+### S3
+
+[S3](https://aws.amazon.com/s3/) is an object store provided by AWS. The
+currently S3ArtifactStore implementation provides various ways to authenticate
```
dbyron-sf, line 33:
```suggestion
current S3ArtifactStore implementation provides various ways to authenticate
```

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/README.md
```diff
@@ -0,0 +1,67 @@
...
+  s3:
+    enabled: true
+    profile: dev # if you want to authenticate using a certain profile
+    region: us-west-2 # allows for specified regions
+    bucket: spinnaker-qa-artifact-store-bucket
+```
+
+While the implementation is S3 specific, this does not limit usages of other S3
+compatible storage engines. For example, something like
+[SeaweedFS](https://github.com/seaweedfs/seaweedfs) can be used to test locally
```
dbyron-sf, line 48:
```suggestion
[SeaweedFS](https://github.com/seaweedfs/seaweedfs) can be used to test locally.
```
and strike the next line.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    return builder.build();
+  }
+
+  /**
+   * hasAuthorization will ensure that the user has proper permissions for retrieving the stored
+   * artifact
+   */
+  private void hasAuthorization(String id, String userId) {
```
nemesisOsorio, line 161: should we move this method to the abstract class and enforce to check authorization? for future `ArtifactStore` implementations?

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/EmbeddedArtifactSerializer.java
```diff
@@ -0,0 +1,63 @@
...
+  /**
+   * shouldStore will return whether we want to store the reference in the ArtifactStore or not.
+   * This checks to ensure the reference isn't null or an empty string. Further we only care about
+   * 'embedded/base64' artifact types, since that is directly embedding the artifacts into the
+   * context
+   */
+  private static boolean shouldStoreArtifact(Artifact artifact) {
+    String ref = artifact.getReference();
+    return artifact.getType().equals(ArtifactTypes.EMBEDDED_BASE64.getMimeType())
+        && !(ref == null || ref.isEmpty());
```
sergio-quintero, line 61:
```suggestion
    return ArtifactTypes.EMBEDDED_BASE64.getMimeType().equals(artifact.getType())
        && !(ref == null || ref.isEmpty());
```

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/README.md
```diff
@@ -0,0 +1,67 @@
...
+tremendously, but will vary depending on the size of the pipeline, as well as
+how that artifact is used, but we've seen improvements of 80% for some
+pipelines.
+
+## Configuration
+
+To enable artifact storage, simple add this to your `spinnaker-local.yml` file
+
+```yaml
+artifactStore:
```
sergio-quintero, line 21: Seems like there is some discrepancy between `artifactStore` and `artifact-store`.
```suggestion
artifact-store:
```

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/README.md
```diff
@@ -0,0 +1,67 @@
...
+## Storage Options
+
+### S3
+
+[S3](https://aws.amazon.com/s3/) is an object store provided by AWS. The
+currently S3ArtifactStore implementation provides various ways to authenticate
+against AWS.
+
+```yaml
+artifactStore:
```
sergio-quintero, line 37:
```suggestion
artifact-store:
```

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Local Testing
+
+To test the artifact store locally, we will use SeaweedFS. To start the storage simply run
+`docker run -p 8333:8333 chrislusf/seaweedfs server -s3`
+
+Next enable the configuration
+
+```yaml
+artifactStore:
```
sergio-quintero, line 59:
```suggestion
artifact-store:
```

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
```diff
@@ -0,0 +1,203 @@
...
+import software.amazon.awssdk.services.s3.model.NoSuchKeyException;
+import software.amazon.awssdk.services.s3.model.PutObjectRequest;
+import software.amazon.awssdk.services.s3.model.Tag;
+import software.amazon.awssdk.services.s3.model.Tagging;
+
+/**
+ * S3ArtifactStore will store artifacts in a s3 compatible service
+ *
+ * <p>Note: It is very important that the S3 bucket has object lock on it to prevent multiple writes
+ * {@see https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock-overview.html}
```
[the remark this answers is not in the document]
xibz: Yes, also works for deletion. You can think of the object lock as an `SQL` row lock

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
xibz: Hmm, potentially? I'll need to think about this some, cause it would require some movement of some stuff.

kork-artifacts/src/test/java/com/netflix/spinnaker/kork/artifacts/artifactstore/EmbeddedArtifactSerializerTest.java
xibz: Hmm, I don't think so. The for loop is iterating for each test case. From that test case we create a storage which then creates a serializer.

That serializer then is used to register the module which is then used  to register itself to the object mapper.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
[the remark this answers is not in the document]
nemesisOsorio: I mean, what is the plan for deleting old artifacts in the bucket?

xibz:
Also added gate PR, https://github.com/spinnaker/gate/pull/1674

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfiguration.java
```diff
@@ -0,0 +1,113 @@
...
+import java.net.URI;
+import java.util.Optional;
+import lombok.extern.log4j.Log4j2;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.access.PermissionEvaluator;
+import software.amazon.awssdk.auth.credentials.*;
```
dbyron-sf, line 29: please avoid star imports.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/README.md
```diff
@@ -0,0 +1,67 @@
...
+these artifacts whenever any stage uses any of those artifacts. Using an
+artifact store reduces this overhead by providing a reference link of,
+`ref://<spinnaker-application>/<content-hash>`. This reduces the context size
+tremendously, but will vary depending on the size of the pipeline, as well as
+how that artifact is used, but we've seen improvements of 80% for some
+pipelines.
+
+## Configuration
+
+To enable artifact storage, simple add this to your `spinnaker-local.yml` file
```
dbyron-sf, line 18: Please add some details about which microservices read from / write to s3.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfiguration.java
```diff
@@ -0,0 +1,113 @@
...
+
+  @Bean
+  @ConditionalOnExpression("${artifact-store.s3.enabled:false}")
+  public ArtifactStore s3ArtifactStore(
+      Optional<PermissionEvaluator> permissionEvaluator,
+      ArtifactStoreConfigurationProperties properties,
+      @Qualifier("artifactS3Client") S3Client s3Client,
+      ArtifactStoreURIBuilder artifactStoreURIBuilder) {
+
+    if (permissionEvaluator.isEmpty()) {
```
dbyron-sf, line 62: How does this happen in practice?

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfigurationProperties.java
```diff
@@ -0,0 +1,44 @@
...
+import lombok.Data;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@Data
+@ConfigurationProperties("artifact-store")
+public class ArtifactStoreConfigurationProperties {
+  private String applicationsRegex = null;
+  /** Configuration for an s3 client which will utilize credentials in the AWS credentials file. */
+  @Data
+  public static class S3ClientConfig {
```
dbyron-sf, line 27: Pondering how helpful it would be to match the knobs clouddriver provides in [S3ArtifactProviderProperties](https://github.com/spinnaker/clouddriver/blob/3e774977fd4287e2aae9e32db3ca4c78ef05ea9e/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/s3/S3ArtifactProviderProperties.java) and [S3ArtifactAccount](https://github.com/spinnaker/clouddriver/blob/3e774977fd4287e2aae9e32db3ca4c78ef05ea9e/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/s3/S3ArtifactAccount.java).

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreURISHA256Builder.java
```diff
@@ -0,0 +1,34 @@
...
+package com.netflix.spinnaker.kork.artifacts.artifactstore;
+
+import com.google.common.hash.Hashing;
+import com.netflix.spinnaker.kork.artifacts.model.Artifact;
+
+/**
+ * Helper class to abstract away the need for other classes to know the {@link * #uriPrefix} format.
+ */
+public class ArtifactStoreURISHA256Builder extends ArtifactStoreURIBuilder {
+  public String buildArtifactURI(String context, Artifact artifact) {
```
dbyron-sf, line 25: `@Override`, here and for `buildRawURI`?

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreURIBuilder.java
```diff
@@ -0,0 +1,35 @@
...
+
+public abstract class ArtifactStoreURIBuilder {
+  /**
+   * uriScheme is used as an HTTP scheme to let us further distinguish a String that is a URI to an
+   * artifact. This is helpful in determining what is an artifact since sometimes we are only given
+   * a string rather than a full artifact.
+   */
+  public static final String uriScheme = "ref";
+
+  public abstract String buildArtifactURI(String context, Artifact artifact);
```
dbyron-sf, line 28: javadoc please, for this and the parameters for buildRawURI.  Curious to see an example for context + raw.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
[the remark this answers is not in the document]
dbyron-sf: Perhaps [OldPipelineCleanupPollingNotificationAgent](https://github.com/spinnaker/orca/blob/002510bb0e5a694c5762a6719505a73e7f5c1556/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/cleanup/OldPipelineCleanupPollingNotificationAgent.kt) could have some additional logic to clean up artifacts when cleaning up old pipeline executions since the logic is more complex than "older than"?  I think it's OK for that to happen in a followup PR.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
```diff
@@ -0,0 +1,203 @@
...
+   *
+   * <p>This method also persists "permissions" by storing the execution id that made the original
+   * store call. In the event a service wants to retrieve said artifact, they will also need to
+   * provide the proper execution id
+   */
+  @Override
+  public Artifact store(Artifact artifact) {
+    String application = AuthenticatedRequest.getSpinnakerApplication().orElse(null);
+    if (application == null) {
+      return artifact;
```
dbyron-sf, line 83: Log a warning/error here?  This isn't supposed to happen is it?

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    if (applicationsRegex != null && !Pattern.matches(applicationsRegex, application)) {
+      return artifact;
+    }
+
+    String ref = uriBuilder.buildArtifactURI(application, artifact);
+    if (objectExists(ref)) {
+      return artifact.toBuilder()
```
dbyron-sf, line 92: nit: build this once in a variable since there's duplicate code below.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
```diff
@@ -0,0 +1,203 @@
...
+    return artifact.toBuilder()
+        .type(ArtifactTypes.REMOTE_BASE64.getMimeType())
+        .reference(ref)
+        .build();
+  }
+
+  public byte[] getReferenceAsBytes(Artifact artifact) {
+    String reference = artifact.getReference();
+    if (reference == null) {
+      return null;
```
dbyron-sf, line 119: log something / throw an exception?  Looks like the caller above assumes a non-null response.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
```diff
@@ -0,0 +1,203 @@
...
+    for (ArtifactDecorator decorator : decorators) {
+      builder = decorator.decorate(builder);
+    }
+
+    return builder.build();
+  }
+
+  /**
+   * hasAuthorization will ensure that the user has proper permissions for retrieving the stored
+   * artifact
```
dbyron-sf, line 159: Worth mentioning that it throws AuthenticationServiceException when unauthorized?

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
```diff
@@ -0,0 +1,203 @@
...
+    return reference.getBytes();
+  }
+
+  /**
+   * get will return the Artifact with the provided id, and will lastly run the {@link
+   * ArtifactDecorator} to further populate the artifact for returning
+   */
+  @Override
+  public Artifact get(String id, ArtifactDecorator... decorators) {
+    hasAuthorization(id, AuthenticatedRequest.getSpinnakerUser().orElse(null));
```
dbyron-sf, line 136: Is it worth calling hasAuthorization if AuthenticatedRequest.getSpinnakerUser() is null?

Maybe hasAuthorization could use a null check either way?

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/ExpectedArtifact.java
```diff
@@ -89,25 +96,53 @@ public boolean matches(Artifact other) {
...
-    return true;
+    return matches(matchArtifact.getReference(), other.getReference());
   }
 
   private boolean matches(@Nullable String us, @Nullable String other) {
     return StringUtils.isEmpty(us) || (other != null && patternMatches(us, other));
   }
 
+  /**
+   * matchTypes will check to see if embedded types are used and if they are, see if they have been
```
dbyron-sf, line 107: This confused me a bit since it made me think the return value depended on whether the artifact had actually been stored in s3 or not.  It's simpler than that though, right?

kork-artifacts/src/test/java/com/netflix/spinnaker/kork/artifacts/artifactstore/EmbeddedArtifactSerializerTest.java
```diff
@@ -0,0 +1,100 @@
...
+    public TestCase(String name, String expectedJSON, Artifact artifact, Artifact mockArtifact) {
+      this.name = name;
+      this.expectedJSON = expectedJSON;
+      this.artifact = artifact;
+      this.mockArtifact = mockArtifact;
+    }
+  }
+
+  @Test
+  public void serializeEmbeddedBase64Artifact_test() throws IOException {
```
dbyron-sf, line 47: nit: @ParameterizedTest / @MethodSource is tempting.

kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/ArtifactUriToReferenceConverter.java
```diff
@@ -0,0 +1,62 @@
...
+      return false;
+    }
+
+    return sourceType.getObjectType() == String.class && targetType.getObjectType() == String.class;
+  }
+
+  @Override
+  public Object convertValue(
+      Object value, TypeDescriptor sourceType, @NotNull TypeDescriptor targetType) {
+    // For some obscene reason(s), SpEL does not use this in the
```
dbyron-sf, line 50: I could some more context.  How is SpEL relevant here?

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfigurationProperties.java
xibz: Not a bad idea. I think originally I was trying to keep v2 of the AWS SDK as separate as possible, but went a little overboard with the config :p

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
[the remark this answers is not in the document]
xibz: Yea, it's just out of scope for S3 artifact stores since they have their own way of handling clean up. However, I do think if someone adds some storage that doesn't implement that, then some logic will be needed to handle that. I haven't thought too much on the design of that since that was outside of the scope, but am totally okay with some future PR to address this.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
xibz: I don't think `getSpinnakerUser` can ever be `null`. I am pretty sure if it is, it gets populated with `anonymous`

dbyron-sf: Is `Objects.requireNonNull` too paranoid?

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/ExpectedArtifact.java
xibz: Yea, it's just returning if the types are "matchable", but due to `embedded/base64` and `remote/base64` being compatible to one another, we check the 4 possibilities.

1. type_a == type_b
2. type_a is embedded/base64 and type_b is remote/base64
3. type_b is embedded/base64 and type_a is remote/base64
4. false otherwise since no matchable types were found

dbyron-sf: Yup, makes sense.  Maybe tweak the description?

kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/ArtifactUriToReferenceConverter.java
xibz: This class is only used in SpEL.

So what happens is when a SpEL function is called and a parameter is a `String`, it will check to see if it is an remote artifact URI, and if it is, return the `base64` reference to the SpEL function.

The reason for this comment is you'd think SpEL would call this `canConvert` automatically when calling these converts, but it doesn't, so we explicitly call them

dbyron-sf: Is there something worth mentioning in the PR description / release notes about some new functionality of the `fromBase64` function?

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
xibz: Yea, but I wanted to account for any possible cases in Spinnaker state. Im not 100% confident in all that Spinnaker does, so my thinking was worst case, we just return the `embedded/base64` artifact back, so that doesn't break anyone.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
xibz: I went with just throwing an exception as you eluded. I think since it is probably never `null` if my memory serves me correctly, and if it is, something went really wrong.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfigurationProperties.java
xibz: We can refactor and instead look up the account based on some account name that is part of the artifact store properties. However, new fields would need to be added to the current `S3ArtifactAccount` class and a new `account-name` field to the artifact store configuration. We'd also need to see what is all compatible in configuration between V1 and V2 of the client configuration. I imagine this shouldn't be an issue.

This does require a little bit more work. Mostly restructuring things to be moved out of clouddriver into kork, and then adding the appropriate fields.

I feel like this can be a separate PR however. I would like to ensure that we implement it internally first so we don't deviate too far from upstream, if that works for you guys. Let me know your thoughts

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfiguration.java
jasonmcintosh: I think you don't get a permissions evaluator UNLESS fiat is enabled, which is in many cases not on by default.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfigurationProperties.java
dbyron-sf: That sounds good, and yeah, totally agree with not holding up this PR for it.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfiguration.java
dbyron-sf: If that's true, this has the potential to be pretty spammy, yes?

dbyron-sf: I dunno, maybe that's ok.

xibz: it only logs it once on startup, so I dont think it'll be too spammy

dbyron-sf: Aaah yes, thanks.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/README.md
```diff
@@ -0,0 +1,130 @@
...
+        |               |<-----------------------------|             |
+        |  Clouddriver  |                              |  Rosco      |
+        |               |                              |             |
+        |  s3 get       |   (full artifact returned)   |  s3 stores  |
+        |  artifacts    |----------------------------->|  artifacts  |
+        +---------------+                              +-------------+
+
+
+Artifact storage is divided into two operations of get and store, and there are
+primarily two services that utilizes each of these operations. Further the
```
dbyron-sf, line 39:
```suggestion
primarily two services that utilize each of these operations. Further the
```

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/README.md
```diff
@@ -0,0 +1,130 @@
...
+operations to limit the amount of code changes needed within these services.
+
+When bootstrapping Spring we add in custom bean serializers and deserializers to
+handle storage or retrieval of an artifact.
+
+Rosco is primarily used for baking artifacts which will generate something
+deployable. When Spring sends a request back to Orca for a bake request, the
+custom serializer injected at Orca startup, will use the artifact storage and
+store the artifact, and instead return a `remote/base64` artifact instead of the
+usual `embedded/base64`.
```
dbyron-sf, line 50:
```suggestion
Rosco is primarily used for baking artifacts which will generate something
deployable.  When Rosco responds to a bake request, the custom serializer
injected in Rosco at startup stores the artifact and returns a `remote/base64`
artifact instead of the usual `embedded/base64`.
```

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreURIBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+   * Returns the remote artifact URI that will be associated with some artifact.
+   *
+   * @param context is the context in which this artifact was run in, e.g. the application.
+   * @param artifact that will be associated with the generated URI.
+   * @return the remote URI
+   */
+  public abstract String buildArtifactURI(String context, Artifact artifact);
+
+  /**
+   * buildRawURI is used when you have the raw path and context. This method just simply return the
```
dbyron-sf, line 38:
```suggestion
   * buildRawURI is used when you have the raw path and context. This method just simply returns the
```

dbyron-sf (approved):

kork-artifacts/kork-artifacts.gradle
```diff
@@ -5,11 +5,23 @@ dependencies {
   api(platform(project(":spinnaker-dependencies")))
 
   implementation project(":kork-annotations")
+  implementation project(":kork-exceptions")
+  implementation project(":kork-security")
   implementation "com.fasterxml.jackson.core:jackson-databind"
+  implementation platform('software.amazon.awssdk:bom')
```
sergio-quintero, line 11: I wasn't able to build clouddriver locally using a local kork snapshot, but it built fine after removing this line. I think this is unnecessary?

xibz: Hmm, I built it just fine. I had to clear gradle cache however.

nemesisOsorio: seems like `software.amazon.awssdk:s3` is duplicated

sergio-quintero (approved):

nemesisOsorio (approved):
I'm excited to test this out

PR #1074, opened 2023-07-06
refactor(retrofit): store the RetrofitException response error body as hashmap 

SheetalAtre:
refactor(retrofit):  store the RetrofitException response error body as hashmap to get more details from it. 
Move the SpinnakerHttpException test to new http related test class.
Return error instead of a RuntimeException when response body is in invalid json format or cannot be converted/parsed.


kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -166,4 +173,16 @@ private static final class RetrofitErrorResponseBody {
       this.message = message;
     }
   }
+
+  public Map<String, Object> getResponseBody() {
+    return this.responseBody;
+  }
+
+  public Map<String, Object> getErrorResponseBody(RuntimeException e) {
```
dbyron-sf, line 181: Let's ditch this function and move the logic back into the appropriate constructor.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -166,4 +173,16 @@ private static final class RetrofitErrorResponseBody {
       this.message = message;
     }
   }
+
+  public Map<String, Object> getResponseBody() {
```
dbyron-sf, line 177: Please add at least some happy-path test for this.

kork-retrofit/src/test/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerServerExceptionTest.java
```diff
@@ -16,8 +16,7 @@
 
 package com.netflix.spinnaker.kork.retrofit.exceptions;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.*;
```
dbyron-sf, line 19: Please avoid star imports.

kork-retrofit/src/test/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerServerExceptionTest.java
```diff
@@ -96,6 +95,7 @@ public void testSpinnakerHttpExceptionFromRetrofitException() {
             .build();
     RetrofitException retrofitException = RetrofitException.httpError(response, retrofit2Service);
     SpinnakerHttpException notFoundException = new SpinnakerHttpException(retrofitException);
+    assertNotNull(notFoundException.getResponseBody());
```
dbyron-sf, line 98: Let's do a bit more than this...Let's also assert that there's a `name` key in the map whose value is `test`.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -46,24 +47,29 @@ public class SpinnakerHttpException extends SpinnakerServerException {
...
+
   public SpinnakerHttpException(RetrofitError e) {
     super(e);
     this.response = e.getResponse();
     this.retrofit2Response = null;
-    RetrofitErrorResponseBody body =
-        (RetrofitErrorResponseBody) e.getBodyAs(RetrofitErrorResponseBody.class);
+    responseBody = (Map<String, Object>) e.getBodyAs(HashMap.class);
+
     this.rawMessage =
```
dbyron-sf, line 58: I have a feeling we'll end up initializing rawMessage something like this
```
this.rawMessage = e.getMessage();
```
when we start handling e.g. non-json response bodies.  It's possible that would simplify the code here even before that.  I'm OK to leave this as it is for now....especially if there's a way to make rawMessage totally disappear.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -166,4 +173,8 @@ private static final class RetrofitErrorResponseBody {
       this.message = message;
     }
   }
+
```
dbyron-sf, line 176: Nothing is using RetrofitErrorResponseBody anymore.  Let's remove it, and the import for Getter once that annotation is gone.

dbyron-sf (approved):

PR #1079, opened 2023-07-18
refactor(retrofit/test): simplify the code by replacing RetrofitException.httpError with SpinnakerHttpException

SheetalAtre:
refactor(retrofit/test): simplify the code by replacing RetrofitException.httpError with SpinnakerHttpException


kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -44,6 +49,10 @@ public class SpinnakerHttpException extends SpinnakerServerException {
   private final String rawMessage;
 
   private final Map<String, Object> responseBody;
+  private retrofit2.Retrofit retrofit;
```
dbyron-sf, line 52:
```suggestion
  private final retrofit2.Retrofit retrofit;
```

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -100,6 +98,28 @@ public SpinnakerHttpException(String message, SpinnakerHttpException cause) {
     this.responseBody = cause.responseBody;
   }
 
+  /**
+   * The constructor handles the HTTP retrofit2 exception, similar to retrofit logic. It is used
+   * with {@link com.netflix.spinnaker.kork.retrofit.ErrorHandlingExecutorCallAdapterFactory}.
+   */
+  public <T> SpinnakerHttpException(retrofit2.Response<T> syncResp, retrofit2.Retrofit retrofit) {
```
dbyron-sf, line 105:
```suggestion
  public <T> SpinnakerHttpException(retrofit2.Response<T> retrofit2Response, retrofit2.Retrofit retrofit) {
```

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -100,6 +98,28 @@ public SpinnakerHttpException(String message, SpinnakerHttpException cause) {
...
+    super(
+        syncResp.code() + " " + syncResp.message(),
+        new Throwable(syncResp.code() + " " + syncResp.message()));
```
dbyron-sf, line 108: These two lines don't feel right.  We're not chaining exceptions here, right?  And, getMessage() has logic that I figure would handle this.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -100,6 +98,28 @@ public SpinnakerHttpException(String message, SpinnakerHttpException cause) {
...
+    this.retrofit2Response = syncResp;
+    this.response = null;
+    this.retrofit = retrofit;
+    responseBody = this.getErrorBodyAs();
```
dbyron-sf, line 112: This may belong in a followup PR, but...if we can't parse the response body as json to build a map, we'd eventually like the exception message to include some part of the response.  I think that means we need a function like processResponse that sets responseBody and either rawMessage or the "plain" message field in this exceptional case.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -157,4 +177,18 @@ public SpinnakerHttpException newInstance(String message) {
   public Map<String, Object> getResponseBody() {
     return this.responseBody;
   }
+
+  private Map<String, Object> getErrorBodyAs() {
```
dbyron-sf, line 181: Let's delete RetrofitException as part of this PR.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -157,4 +177,18 @@ public SpinnakerHttpException newInstance(String message) {
...
+    if (retrofit2Response == null) {
+      return null;
+    }
+
+    Converter<ResponseBody, Map> converter =
+        retrofit.responseBodyConverter(Map.class, new Annotation[0]);
+    try {
+      return converter.convert(retrofit2Response.errorBody());
+    } catch (IOException e) {
+      return jsonErrorResponseBody;
```
dbyron-sf, line 191: This change isn't really related to the rest of the refactor....and it's not really a refactor...it's either a fix or a new feature to improve the handling of non-json responses.  The [original code](https://github.com/spinnaker/kork/blob/7f5fe05a5928d6a9266371d09414dd4fb4f3c7f5/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/RetrofitException.java#L79) throws a RuntimeException.  This PR is already big enough, so instead of splitting it into a separate commit, let's leave it for a followup PR and continue to throw a RuntimeException here.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -98,6 +93,28 @@ public SpinnakerHttpException(String message, SpinnakerHttpException cause) {
...
+    this.retrofit = null;
+  }
+
+  /**
+   * The constructor handles the HTTP retrofit2 exception, similar to retrofit logic. It is used
+   * with {@link com.netflix.spinnaker.kork.retrofit.ErrorHandlingExecutorCallAdapterFactory}.
+   */
+  public <T> SpinnakerHttpException(
+      retrofit2.Response<T> retrofit2Response, retrofit2.Retrofit retrofit) {
+    super(new Throwable(retrofit2Response.code() + " " + retrofit2Response.message()));
```
dbyron-sf, line 105: Why make up a cause here?  Doesn't the logic in getMessage() handle building the message?

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -141,7 +158,7 @@ public String getMessage() {
           "Status: %s, URL: %s, Message: %s",
           retrofit2Response.code(),
           retrofit2Response.raw().request().url().toString(),
-          getRawMessage());
+          retrofit2Response.code() + " " + retrofit2Response.message());
```
dbyron-sf, line 161: This doesn't look correct.  Won't the resulting message have the code twice, and ignore the message property in the response body if there is one?

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -98,6 +93,28 @@ public SpinnakerHttpException(String message, SpinnakerHttpException cause) {
...
+    this.retrofit = null;
+  }
+
+  /**
+   * The constructor handles the HTTP retrofit2 exception, similar to retrofit logic. It is used
+   * with {@link com.netflix.spinnaker.kork.retrofit.ErrorHandlingExecutorCallAdapterFactory}.
+   */
+  public SpinnakerHttpException(
+      retrofit2.Response<?> retrofit2Response, retrofit2.Retrofit retrofit) {
+    super(new Throwable());
```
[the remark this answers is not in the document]
SheetalAtre: Added a no-arg constructor in the super SpinnakerServerException class to fix this

dbyron-sf (approved):

PR #1085, opened 2023-08-08
refactor(retrofit/test): moved http tests to new SpinnakerHttpExceptionTest class

SheetalAtre:
refactor(retrofit/test): moved http tests to new SpinnakerHttpExceptionTest class


kork-retrofit/src/test/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerServerExceptionTest.java
```diff
@@ -81,28 +74,4 @@ public void testSpinnakerServerException_NewInstance() {
       assertEquals(e, newException.getCause());
     }
   }
-
```
dbyron-sf, line 84: Let's move testSpinnakerHttpException_NewInstance as well.

PR #1087, opened 2023-08-10
chore(dependencies): remove unnecessary dependency on log4j-core

dbyron-sf:
and make dependencies on log4j-api explicit.

https://github.com/spinnaker/kork/pull/1069 introduced the dependency on org.apache.logging.log4j.log4j-core in kork-artifacts, but it's not necessary.  At least, after removing it, and verifying the dependency is gone from
```
$ ./gradle kork-artifacts:dependencies
```
the code still builds fine.  As well
```
$ git grep "import org.apache.logging.log4j"
kork-secrets/src/main/java/com/netflix/spinnaker/kork/secrets/user/UserSecretTypeProvider.java:import org.apache.logging.log4j.LogManager; kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequestDecorator.java:import org.apache.logging.log4j.ThreadContext;
```
turns up classes present in log4j-api, but nothing in log4j-core.

As well uses of the `@Log4j2` annotation depend on log4j-api (see https://projectlombok.org/api/lombok/extern/log4j/Log4j2)
```
$ git grep @Log4j2
kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfiguration.java:@Log4j2 kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java:@Log4j2 kork-core/src/main/java/com/netflix/spinnaker/kork/jackson/JsonTypeNameParser.java:@Log4j2 kork-credentials/src/main/java/com/netflix/spinnaker/credentials/jackson/SensitiveSerializer.java:@Log4j2 kork-secrets/src/main/java/com/netflix/spinnaker/kork/secrets/EncryptedSecret.java:@Log4j2
```
This all makes the comment in [spinnaker-dependencies.gradle](https://github.com/spinnaker/kork/blob/ed082ade1ec460a0310b78d71e84906862ff269c/spinnaker-dependencies/spinnaker-dependencies.gradle#L58) true again:

> Per analysis, log4j-core is not included in dependencies

While we're at it, get log4j-bom up to date.

### DIFF
--- a/kork-artifacts/kork-artifacts.gradle
+++ b/kork-artifacts/kork-artifacts.gradle
@@ -14,7 +14,6 @@ dependencies {
   implementation "org.springframework.boot:spring-boot-autoconfigure"
   implementation "org.springframework.security:spring-security-core"
   implementation 'org.apache.logging.log4j:log4j-api'
-  implementation 'org.apache.logging.log4j:log4j-core'
   api "com.hubspot.jinjava:jinjava"
 
   testImplementation "org.assertj:assertj-core"
--- a/kork-core/kork-core.gradle
+++ b/kork-core/kork-core.gradle
@@ -18,6 +18,7 @@ dependencies {
   api "io.github.resilience4j:resilience4j-retry"
   api "io.github.resilience4j:resilience4j-spring-boot2"
 
+  implementation "org.apache.logging.log4j:log4j-api"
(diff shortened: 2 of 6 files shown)


xibz (approved):
LGTM!

PR #1090, opened 2023-08-14
Support additional auth types

jvz:
This has a couple changes related to supporting principals with the type `AuthenticatedPrincipal` in addition to the existing support for `UserDetails`. This allows for future migrations to the built-in SAML and OAuth2 Spring Security modules that we'll need to switch to at some point as those APIs rely on this API for federated identities. I've also ensured that anonymous users are using a consistent username.

Another nicety is a dedicated `GrantedAuthority` implementation for the allowed accounts concept. This can help with migrating that to regular security APIs over time.

There are numerous places in Gate (and other services less so) that can be updated with these changes. In particular, all the uses of the deprecated `User` class from `kork-security` should be migrated over time, especially any of the APIs exposing data or requiring a `User` parameter.


jvz:
Discovered a small issue with this idea, so putting back in draft.

jvz:
Going to have to redo the relevant changes from here as I've included some of these changes in a different PR.

jvz:
Included in https://github.com/spinnaker/spinnaker/pull/7245 and no longer relevant.

PR #1095, opened 2023-08-25
feat(retrofit/test): handle non json http response

SheetalAtre:
feat(retrofit/test): handle non json http response


dbyron-sf:
implemented in https://github.com/spinnaker/kork/pull/1097

PR #1100, opened 2023-09-25
feat (retrofit): Add logic to retrofit2 to handle json processing exceptions equivalent to retrofit1.

Pranav-b-7:
Written logic to handle all kind of Json Processing exceptions thrown by Jackson converter,

- A new catch block added under the execute() method of the class ErrorHandlingExecutorCallAdapterFactory.java.
- This block catches all kind of exceptions under JsonProcessingException . These are : 
- DatabindException
- IgnoredPropertyException
- InputCoercionException
- InvalidDefinitionException
- InvalidFormatException
- InvalidNullException
- InvalidTypeIdException
- JacksonYAMLParseException
- JsonEOFException
- JsonGenerationException
- JsonMappingException
- JsonParseException
- MismatchedInputException
- MissingKotlinParameterException
- PropertyBindingException
- StreamReadException
- StreamWriteException
- UnrecognizedPropertyException
- UnresolvedForwardReference
- ValueInstantiationException

When execute() method throws this exception , this gets caught in the catch block and a customised exception is thrown namely SpinnakerConversionException.


kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerJsonProcessingException.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.kork.retrofit.exceptions;
+
+import com.netflix.spinnaker.kork.exceptions.SpinnakerException;
+
+public class SpinnakerJsonProcessingException extends SpinnakerException {
```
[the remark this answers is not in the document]
Pranav-b-7: By the naming convention I felt SpinnakerJsonProcessingException makes more sense,
and also SpinnakerConversion is used for retrofit1 and thought of creating a new class for handling the retrofit2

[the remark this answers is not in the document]
Pranav-b-7: Deleted SpinnakerJsonProcessingException class as suggested , and replaced this part of the catch block with the existing class SpinnakerConversionException.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -149,6 +151,8 @@ public Response<T> execute() {
         if (syncResp.isSuccessful()) {
           return syncResp;
         }
+      } catch (JsonProcessingException jpe) {
```
dbyron-sf, line 154: Is this sufficient to handle the various kinds of converters that are possible to use with retrofit2?  I see [this method](https://github.com/square/retrofit/blob/parent-2.8.1/retrofit/src/main/java/retrofit2/Converter.java#L42) throws an IOException.  It is possible to get access to this lower-down call so we can know more specifically that it's what's failing?

Pranav-b-7: @dbyron-sf - This block catches all kind of exceptions under JsonProcessingException . These are :
DatabindException
IgnoredPropertyException
InputCoercionException
InvalidDefinitionException
InvalidFormatException
InvalidNullException
InvalidTypeIdException
JacksonYAMLParseException
JsonEOFException
JsonGenerationException
JsonMappingException
JsonParseException
MismatchedInputException
MissingKotlinParameterException
PropertyBindingException
StreamReadException
StreamWriteException
UnrecognizedPropertyException
UnresolvedForwardReference
ValueInstantiationException

I have verified it from the Jackson Converter library, So I think catching the parent exception of all types JSON parse failure would ideally be sufficient

dbyron-sf: Right...but what if someone uses [GsonResponseBodyConverter](https://github.com/square/retrofit/blob/parent-2.8.1/retrofit-converters/gson/src/main/java/retrofit2/converter/gson/GsonResponseBodyConverter.java) or some other implementation of [Converter](https://github.com/square/retrofit/blob/parent-2.8.1/retrofit/src/main/java/retrofit2/Converter.java)?

Pranav-b-7:
@dbyron-sf - as suggested I have reverted the exception handling logic thrown by gson-converter  library. This PR now only contains the logic to handle the exceptions thrown by jackson-converter.
Pls refer to this link : https://github.com/square/retrofit/issues/3972 to get more insights about less/ well designed converter libraries supported by retrofit2

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
dbyron-sf: See https://github.com/square/retrofit/issues/3972 for why it's hard to handle arbitrary converters.

dbyron-sf (approved):

PR #1101, opened 2023-09-26
feat(artifacts): Add ArtifactReferenceURI to artifact store

xibz:
This commit adds a new URI type to the artifact store signature. This allows for storage implementation to more easily use certain pieces of the URI for storage. For example, with this commit, artifacts stored as S3 objects will no longer have the `ref://` prefix, allowing for a better AWS console experience

### DIFF
--- a/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
+++ b/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
@@ -17,6 +17,7 @@ package com.netflix.spinnaker.kork.artifacts.artifactstore.s3;
 
 import com.netflix.spinnaker.kork.artifacts.ArtifactTypes;
 import com.netflix.spinnaker.kork.artifacts.artifactstore.ArtifactDecorator;
+import com.netflix.spinnaker.kork.artifacts.artifactstore.ArtifactReferenceURI;
 import com.netflix.spinnaker.kork.artifacts.artifactstore.ArtifactStore;
 import com.netflix.spinnaker.kork.artifacts.artifactstore.ArtifactStoreURIBuilder;
 import com.netflix.spinnaker.kork.artifacts.model.Artifact;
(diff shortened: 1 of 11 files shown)


xibz:
PR to update clouddriver, https://github.com/spinnaker/clouddriver/pull/6058

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
```diff
@@ -193,15 +197,16 @@ private void hasAuthorization(String id, String userId) {
    * writes of the same object is important, another filestore/db needs to be used, possibly
    * dynamodb.
    */
-  private boolean objectExists(String id) {
-    HeadObjectRequest request = HeadObjectRequest.builder().bucket(bucket).key(id).build();
+  private boolean objectExists(ArtifactReferenceURI uri) {
+    HeadObjectRequest request = HeadObjectRequest.builder().bucket(bucket).key(uri.paths()).build();
     try {
       s3Client.headObject(request);
+      log.info("Artifact exists. No need to store. reference={}", uri.uri());
```
dbyron-sf, line 204: Is this going to get a bit noisy?  Maybe better as debug?

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
```diff
@@ -193,15 +197,16 @@ private void hasAuthorization(String id, String userId) {
...
       return true;
     } catch (NoSuchKeyException e) {
       // pretty gross that we need to use exceptions as control flow, but the
       // java SDK doesn't have any other way of check if an object exists in s3
-      log.warn("Artifact does not exist reference={}", id, e);
+      log.warn("Artifact does not exist reference={}", uri.uri());
```
dbyron-sf, line 209: Is this really a warning?  Like, nothing is really wrong when this happens, is it?  Feels like info is better.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
xibz: done!

Yea, we talked about this internally, and this was my exact concern. The reasoning I received for it was pretty sound. The idea is in the event where you want to see if `objectExists` is reached, you wouldn't know other than the missing log which may or may not have reached this point.

However, with that said, I think we can go with a hybrid approach. Make this debug, and add more logging elsewhere. That will be a separate PR however. I like my PRs properly scoped :p

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
xibz: done :)

dbyron-sf (approved):

PR #1103, opened 2023-10-02
fix(artifact): Handle exceptions better in artifact store

xibz:
Prior to this commit, when the head object in s3 artifact store throws an exception in (de)serializer, we get a pretty bland exception back

```json
{"timestamp":"2023-09-25T17:54:02.550+00:00","status":500,"error":"Internal
Server Error"}
```

With this commit we will now see:
```json
{"timestamp":"2023-10-02T23:21:42.183+00:00","status":500,"error":"Internal Server Error","exception":"com.netflix.spinnaker.kork.artifacts.artifactstore.exceptions.ArtifactStoreIOException","message":"com.netflix.spinnaker.kork.exceptions.SpinnakerException: Failed to query artifact due to IAM permissions either on the bucket or object"}
```

The fix here was to intercept any `Exception` from the artifact store calls and ensure we rethrow a `ArtifactStoreIOException`. The reason why we rethrow an `IOException` is due to what jackson does with exceptions. If we do not throw an `IOException`, then jackson will wrap it in a JSON exception which is confusing to end user.

### DIFF
--- a/kork-artifacts/kork-artifacts.gradle
+++ b/kork-artifacts/kork-artifacts.gradle
@@ -20,5 +20,6 @@ dependencies {
   testImplementation "org.junit.jupiter:junit-jupiter-api"
   testImplementation "org.junit.jupiter:junit-jupiter-params"
   testImplementation "org.mockito:mockito-core"
+  testImplementation project(":kork-core")
   testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine"
 }
--- a/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/exceptions/ArtifactStoreIOException.java
+++ b/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/exceptions/ArtifactStoreIOException.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2023 Apple Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.artifacts.artifactstore.exceptions;
+
+import java.io.IOException;
+import java.util.function.Supplier;
+
+/**
+ * An exception used by the artifact store (de)serializers since the (de)serialize methods only
+ * throw IOExceptions, and if any other exception is thrown jackson assumes it's some JSON error.
+ */
+public class ArtifactStoreIOException extends IOException {
+  public ArtifactStoreIOException(Exception e) {
+    super(e);
+  }
+
+  /**
+   * Helper methods to catch any exception thrown by a method and instead throw an
+   * ArtifactStoreIOException
+   */
+  public static <T> T throwIOException(Supplier<T> fn) throws IOException {
+    try {
+      return fn.get();
+    } catch (Exception e) {
+      throw new ArtifactStoreIOException(e);
+    }
+  }
+}
--- a/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
+++ b/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
@@ -21,11 +21,13 @@ import com.netflix.spinnaker.kork.artifacts.artifactstore.ArtifactReferenceURI;
 import com.netflix.spinnaker.kork.artifacts.artifactstore.ArtifactStore;
 import com.netflix.spinnaker.kork.artifacts.artifactstore.ArtifactStoreURIBuilder;
 import com.netflix.spinnaker.kork.artifacts.model.Artifact;
+import com.netflix.spinnaker.kork.exceptions.SpinnakerException;
 import com.netflix.spinnaker.security.AuthenticatedRequest;
 import java.util.Base64;
 import java.util.NoSuchElementException;
 import java.util.regex.Pattern;
 import lombok.extern.log4j.Log4j2;
+import org.apache.http.HttpStatus;
 import org.springframework.security.access.PermissionEvaluator;
 import org.springframework.security.authentication.AuthenticationServiceException;
 import org.springframework.security.core.Authentication;
@@ -40,6 +42,7 @@ import software.amazon.awssdk.services.s3.model.GetObjectTaggingResponse;
 import software.amazon.awssdk.services.s3.model.HeadObjectRequest;
 import software.amazon.awssdk.services.s3.model.NoSuchKeyException;
 import software.amazon.awssdk.services.s3.model.PutObjectRequest;
+import software.amazon.awssdk.services.s3.model.S3Exception;
 import software.amazon.awssdk.services.s3.model.Tag;
 import software.amazon.awssdk.services.s3.model.Tagging;
 
@@ -208,6 +211,37 @@ public class S3ArtifactStore extends ArtifactStore {
       // java SDK doesn't have any other way of check if an object exists in s3
       log.info("Artifact does not exist reference={}", uri.uri());
       return false;
+    } catch (S3Exception e) {
+      int statusCode = e.statusCode();
+      log.error(
+          "Artifact store failed head object request statusCode={} reference={}",
+          statusCode,
+          uri.uri());
+
+      if (statusCode != 0) {
+        // due to this being a HEAD request, there is no message giving a clear
+        // indication of what failed. Rather than seeing a useful message back
+        // to gate, we instead see just null. To alleviate this, we wrap the
+        // exception with a more meaningful message
+        throw new SpinnakerException(buildHeadObjectExceptionMessage(e), e);
+      }
+
+      throw new SpinnakerException("S3 head object failed", e);
+    }
+  }
+
+  /**
+   * S3's head object can only return 400, 403, and 404, and based on the HTTP status code, we will
+   * return the appropriate message back
+   */
+  private static String buildHeadObjectExceptionMessage(S3Exception e) {
+    switch (e.statusCode()) {
+      case HttpStatus.SC_FORBIDDEN:
+        return "Failed to query artifact due to IAM permissions either on the bucket or object";
+      case HttpStatus.SC_BAD_REQUEST:
+        return "Failed to query artifact due to invalid request";
+      default:
+        return String.format("Failed to query artifact: %d", e.statusCode());
     }
   }
 }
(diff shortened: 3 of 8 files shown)


kork-artifacts/kork-artifacts.gradle
```diff
@@ -20,5 +20,6 @@ dependencies {
   testImplementation "org.junit.jupiter:junit-jupiter-api"
   testImplementation "org.junit.jupiter:junit-jupiter-params"
   testImplementation "org.mockito:mockito-core"
+  testImplementation "junit:junit"
```
dbyron-sf, line 23: If this is junit 4.x (and I think it is)...we're trying to move away from it, and towards jupiter...

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
```diff
@@ -208,6 +211,34 @@ private boolean objectExists(ArtifactReferenceURI uri) {
...
       log.info("Artifact does not exist reference={}", uri.uri());
       return false;
+    } catch (S3Exception e) {
+      int statusCode = e.statusCode();
+      if (statusCode != 0) {
+        // due to this being a HEAD request, there is no message giving a clear
+        // indication of what failed. Rather than seeing a useful message back
+        // to gate, we instead see just null. To alleviate this, we wrap the
+        // exception with a more meaningful message
+        log.error("Artifact store failed head object request statusCode={}", statusCode);
```
dbyron-sf, line 221: Can't decide if this is a nit or a more legit concern.  If the statusCode is 0, don't we still want to log it?  As in, does it make sense to move this log message outside the if?

Plus also, is it helpful to include `uri.uri()` no matter what the status code is?  So then I think I'm asking to combine this message with the the one below.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
```diff
@@ -208,6 +211,34 @@ private boolean objectExists(ArtifactReferenceURI uri) {
...
+    }
+  }
+
+  /**
+   * S3's head object can only return 400, 403, and 404, and based on the HTTP status code, we will
+   * return the appropriate message back
+   */
+  private static String buildHeadObjectExceptionMessage(S3Exception e) {
+    switch (e.statusCode()) {
+      case HttpStatus.SC_FORBIDDEN:
```
dbyron-sf, line 236: This is a bigger request, but using the s3 response code as spinnaker's response code would be super nice....so caller's wouldn't bother retrying on a 400/403.

kork-artifacts/src/test/java/com/netflix/spinnaker/kork/artifacts/artifactstore/exceptions/ArtifactStoreIOExceptionTest.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.artifacts.artifactstore.exceptions;
+
+import static org.junit.jupiter.api.Assertions.*;
```
dbyron-sf, line 18: please avoid star imports, even in tests.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
xibz: I think the addition of the `uri.uri()` is useful, but I dont think we need to include the `statusCode == 0` case, as that'll throw and log a different message

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
xibz: Yep, this is going to be a separate PR on retrofit! Going to fix the NPE bug due to HEAD requests, and handle the new exception :)

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
```diff
@@ -208,6 +211,37 @@ private boolean objectExists(ArtifactReferenceURI uri) {
...
+        // to gate, we instead see just null. To alleviate this, we wrap the
+        // exception with a more meaningful message
+        log.error(
+            "Artifact store failed head object request statusCode={} reference={}",
+            statusCode,
+            uri.uri());
+        throw new SpinnakerException(buildHeadObjectExceptionMessage(e), e);
+      }
+
+      log.warn("head object failed unexpectedly reference={}", uri.uri());
```
dbyron-sf, line 228: I think I'd still make this an error since it still seems pretty fatal.  There's always a bit of tension between what the exception handler logs and logging here at the source of the exception.  At the moment, I believe this is the handler, and so it'll log two messages:
* [a warning about using the generic handler](https://github.com/spinnaker/kork/blob/792ada8061cce7b89392eacbc304ba317c3596aa/kork-web/src/main/java/com/netflix/spinnaker/kork/web/exceptions/GenericExceptionHandlers.java#L131)
* [an "Internal Server Error" error](https://github.com/spinnaker/kork/blob/792ada8061cce7b89392eacbc304ba317c3596aa/kork-web/src/main/java/com/netflix/spinnaker/kork/web/exceptions/GenericExceptionHandlers.java#L162)

Both log the entire exception, so perhaps it's not worth logging it here (i.e. passing `e` as an argument to log)?  Let's see what happens in the next PR where this handler/these two log messages are likely no longer relevant.

xibz: Hmm, I did think about that, but I think only useful thing you'd get is the stack trace. The exception itself is very bare due to it being a `HEAD` request with no body. But yea, I'll go ahead and change it to an `error` level logging, and include the exceptions. It doesn't hurt to include them in this particular case

kork-artifacts/kork-artifacts.gradle
```diff
@@ -20,5 +20,6 @@ dependencies {
   testImplementation "org.junit.jupiter:junit-jupiter-api"
   testImplementation "org.junit.jupiter:junit-jupiter-params"
   testImplementation "org.mockito:mockito-core"
+  testImplementation project(":kork-core")
```
xibz, line 23: needed for `Header.APPLICATION` in new test

kork-artifacts/src/test/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStoreTest.java
```diff
@@ -0,0 +1,58 @@
...
+        new S3ArtifactStore(client, null, "my-bucket", new ArtifactStoreURISHA256Builder(), null);
+    String expectedExceptionMessage = "Failed to query artifact due to invalid request";
+    SpinnakerException e =
+        Assertions.assertThrows(
+            SpinnakerException.class,
+            () -> {
+              artifactStore.store(
+                  Artifact.builder()
+                      .type(ArtifactTypes.EMBEDDED_BASE64.getMimeType())
+                      .reference("aGVsbG8gd29ybGQK")
```
dbyron-sf, line 53: Possible to get this from `ArtifactStoreURISHA256Builder`, or somehow not hard code it?

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/exceptions/ArtifactStoreIOException.java
```diff
@@ -0,0 +1,50 @@
...
+  /**
+   * Helper methods to catch any exception thrown by a method and instead throw an
+   * ArtifactStoreIOException
+   */
+  public static <T> T throwIOException(Supplier<T> fn, String exceptionMessage) throws IOException {
+    try {
+      return fn.get();
+    } catch (Exception e) {
+      throw new ArtifactStoreIOException(
+          (exceptionMessage == null) ? e.getMessage() : exceptionMessage, e);
```
dbyron-sf, line 43: Is duplicating the message here going to help?  Seems cleaner to only call the constructor that takes a message if there really is one.

kork-artifacts/src/test/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStoreTest.java
xibz: This is just an arbitrary data. This could be `foo` for example

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/exceptions/ArtifactStoreIOException.java
xibz: Yea, the issue is due to retrofit not handling the `ArtifactStoreIOException` the message will get dropped and we would be left with no message again. So this may be temporary for now

kork-artifacts/src/test/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStoreTest.java
dbyron-sf: Perhaps
```suggestion
                      .reference("aGVsbG8gd29ybGQK") // arbitrary
```
then?

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/exceptions/ArtifactStoreIOException.java
xibz: One apporach may be to write `getMessage` for `ArtifactStoreIOException` instead

kork-artifacts/src/test/java/com/netflix/spinnaker/kork/artifacts/artifactstore/EmbeddedArtifactSerializerTest.java
```diff
@@ -51,6 +60,39 @@ public void serializeEmbeddedBase64Artifact_test(
...
+
+    ArtifactStoreIOException e =
+        assertThrows(
+            ArtifactStoreIOException.class,
+            () -> {
+              objectMapper.writeValue(
+                  new ByteArrayOutputStream(),
+                  Artifact.builder()
+                      .type(ArtifactTypes.EMBEDDED_BASE64.getMimeType())
+                      .reference("aGVsbG8gd29ybGQK")
```
xibz, line 87:
```suggestion
                      .reference("aGVsbG8gd29ybGQK") // arbitrary
```

dbyron-sf (approved):

PR #1104, opened 2023-10-09
feat(retrofit2): Handled http response bodies for http errors that are not json.

Pranav-b-7:
Handled http response bodies for http errors that are not json. This exception is handled with the default catch block/ i.e with the parent exception class : "Exception", as the way of handling would remain same for all types of exceptions.

Scenario when this logic will be executed is : When the retrofit receives any http failure responses from the client , like with the status codes of 5xx series or 4xx series , then this error will be handled in kork with a customised exception: "SpinnakerHttpException",
This class will try to parse the error details and if the error details parsing fails then,  this will be handled gracefully with this new code changes.

### DIFF
--- a/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
+++ b/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
@@ -218,10 +218,11 @@ public class SpinnakerHttpException extends SpinnakerServerException {
   }
 
   /**
-   * HTTP response body converted to specified {@code type}. {@code null} if there is no response.
+   * HTTP error response body converted to specified {@code type}. {@code null} if there is no
+   * response.
    *
-   * @throws RuntimeException wrapping the underlying IOException if unable to convert the body to
-   *     the specified {@code type}.
+   * @return null wrapping the underlying Exception if unable to convert the body to the specified
+   *     {@code type}.
    */
   private Map<String, Object> getErrorBodyAs(Retrofit retrofit) {
     if (retrofit2Response == null) {
@@ -233,9 +234,8 @@ public class SpinnakerHttpException extends SpinnakerServerException {
     try {
       return converter.convert(retrofit2Response.errorBody());
     } catch (Exception e) {
-      Map<String, Object> errorBody = new HashMap<>();
-      errorBody.put("message", e.toString());
-      return errorBody;
+      log.debug("Exception occurred while parsing the HTTP error response body : {}", e);
+      return null;
     }
   }
 }
--- a/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
+++ b/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
@@ -234,10 +234,11 @@ public class SpinnakerHttpException extends SpinnakerServerException {
       return converter.convert(retrofit2Response.errorBody());
     } catch (Exception e) {
       log.debug(
-          "Exception occurred while parsing the HTTP error response body. Status code : {}, URL : {}, stack trace : ",
-          retrofit2Response.code(),
-          retrofit2Response.raw().request().url(),
-          e);
+        "unable to convert response to map ({} {}, {})",
+        retrofit2Response.raw().request().method(),
+        retrofit2Response.code(),
+        retrofit2Response.raw().request().url(),
+        e);
       return null;
     }
   }
--- a/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
+++ b/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
@@ -18,7 +18,6 @@ package com.netflix.spinnaker.kork.retrofit.exceptions;
 
 import com.google.common.base.Preconditions;
 import com.netflix.spinnaker.kork.annotations.NullableByDefault;
-import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.util.HashMap;
 import java.util.Map;
@@ -233,8 +232,10 @@ public class SpinnakerHttpException extends SpinnakerServerException {
         retrofit.responseBodyConverter(Map.class, new Annotation[0]);
     try {
       return converter.convert(retrofit2Response.errorBody());
-    } catch (IOException e) {
-      throw new RuntimeException(e);
+    } catch (Exception e) {
+      Map<String, Object> errorBody = new HashMap<>();
+      errorBody.put("message", e.toString());
+      return errorBody;
     }
   }
 }
--- a/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
+++ b/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
@@ -218,11 +218,10 @@ public class SpinnakerHttpException extends SpinnakerServerException {
   }
 
   /**
-   * HTTP error response body converted to specified {@code type}. {@code null} if there is no
-   * response.
+   * HTTP error response body converted to the specified {@code type}.
    *
-   * @return null wrapping the underlying Exception if unable to convert the body to the specified
-   *     {@code type}.
+   * @return null if there's no response or unable to convert the body to the specified {@code
+   *     type}.
    */
   private Map<String, Object> getErrorBodyAs(Retrofit retrofit) {
     if (retrofit2Response == null) {
@@ -234,7 +233,11 @@ public class SpinnakerHttpException extends SpinnakerServerException {
     try {
       return converter.convert(retrofit2Response.errorBody());
     } catch (Exception e) {
-      log.debug("Exception occurred while parsing the HTTP error response body : {}", e);
+      log.debug(
+          "Exception occurred while parsing the HTTP error response body. Status code : {}, URL : {}, stack trace : ",
+          retrofit2Response.code(),
+          retrofit2Response.raw().request().url(),
+          e);
       return null;
     }
   }
--- a/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
+++ b/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
@@ -234,11 +234,11 @@ public class SpinnakerHttpException extends SpinnakerServerException {
       return converter.convert(retrofit2Response.errorBody());
     } catch (Exception e) {
       log.debug(
-        "unable to convert response to map ({} {}, {})",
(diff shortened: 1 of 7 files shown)


kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -233,8 +232,10 @@ private Map<String, Object> getErrorBodyAs(Retrofit retrofit) {
         retrofit.responseBodyConverter(Map.class, new Annotation[0]);
     try {
       return converter.convert(retrofit2Response.errorBody());
-    } catch (IOException e) {
-      throw new RuntimeException(e);
+    } catch (Exception e) {
```
dbyron-sf, line 235: Isn't IOException enough since that's what [convert](https://github.com/square/retrofit/blob/parent-2.8.1/retrofit/src/main/java/retrofit2/Converter.java#L42) throws?  In other words, won't we get some kind of IOException when the response body isn't json?

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -233,8 +232,10 @@ private Map<String, Object> getErrorBodyAs(Retrofit retrofit) {
...
+      Map<String, Object> errorBody = new HashMap<>();
```
dbyron-sf, line 236: I think I'd rather return null here to match how the equivalent code for retrofit1 works.  Please also update the javadoc.

kork-retrofit/src/test/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerRetrofit2ErrorHandleTest.java
```diff
@@ -193,6 +193,24 @@ void testSpinnakerConversionException() {
     assertEquals("Failed to process response body", spinnakerConversionException.getMessage());
   }
 
+  @Test
+  public void testNonJsonRetrofitErrorResponse() {
```
dbyron-sf, line 197:
```suggestion
  public void testNonJsonHttpErrorResponse() {
```

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
Pranav-b-7: As discussed over the call. Handling the exceptions with the default catch block would be more optimised solution. As it handles not only JsonParseException but also other exceptions as well

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
Pranav-b-7: I have modified as per the comments. Pls check

kork-retrofit/src/test/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerRetrofit2ErrorHandleTest.java
Pranav-b-7: I have modified as per the comments. Pls check

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
dbyron-sf: I'm not sure optimization has much to do with it.  The [retrofit1 code catches Exception](https://github.com/spinnaker/kork/blob/c1d5676f0fcb326308fc759a2f4fc1f633ecac4b/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java#L85), so it makes sense to match that behavior.  Because the convert method here throws a more specific exception, we have an opportunity to potentially do better here, but we're making a call that the user would rather get a SpinnakerHttpException that includes a URL and http response code and no info regarding the response body than a RuntimeException (or something else) that happens while converting the body.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -219,10 +218,11 @@ public String getReason() {
   }
 
   /**
-   * HTTP response body converted to specified {@code type}. {@code null} if there is no response.
+   * HTTP error response body converted to specified {@code type}. {@code null} if there is no
+   * response.
    *
-   * @throws RuntimeException wrapping the underlying IOException if unable to convert the body to
-   *     the specified {@code type}.
+   * @return null wrapping the underlying Exception if unable to convert the body to the specified
```
dbyron-sf, line 224:
```suggestion
   * @return null if there's no response or unable to convert the body to the specified
```

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -219,10 +218,11 @@ public String getReason() {
   }
 
   /**
-   * HTTP response body converted to specified {@code type}. {@code null} if there is no response.
+   * HTTP error response body converted to specified {@code type}. {@code null} if there is no
```
dbyron-sf, line 221: Let's keep the info about the return value in one place.
```suggestion
   * HTTP error response body converted to the specified {@code type}.
```

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -233,8 +233,9 @@ private Map<String, Object> getErrorBodyAs(Retrofit retrofit) {
         retrofit.responseBodyConverter(Map.class, new Annotation[0]);
     try {
       return converter.convert(retrofit2Response.errorBody());
-    } catch (IOException e) {
-      throw new RuntimeException(e);
+    } catch (Exception e) {
+      log.debug("Exception occurred while parsing the HTTP error response body : {}", e);
```
dbyron-sf, line 237:
```suggestion
      log.debug("Exception occurred while parsing the HTTP error response body", e);
```
but also, please include the url and response code from the response to make the log message more useful.  This means adding a new argument to the method.

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
Pranav-b-7: Modified the javadoc statement to : @return null if there's no response or unable to convert the body to the specified

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
Pranav-b-7: Modified the javadoc statement to : HTTP error response body converted to the specified {@code type}

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
Pranav-b-7: Modified the log message to print the status code , url and stack trace

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
```diff
@@ -233,8 +232,13 @@ private Map<String, Object> getErrorBodyAs(Retrofit retrofit) {
         retrofit.responseBodyConverter(Map.class, new Annotation[0]);
     try {
       return converter.convert(retrofit2Response.errorBody());
-    } catch (IOException e) {
-      throw new RuntimeException(e);
+    } catch (Exception e) {
+      log.debug(
+          "Exception occurred while parsing the HTTP error response body. Status code : {}, URL : {}, stack trace : ",
```
dbyron-sf, line 237: Let's make this match the [retrofit1 error message](https://github.com/spinnaker/kork/blob/cf1bcaa69088c84ec3734f1ea952d199d11e91d3/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java#L95-L99)
```suggestion
          "unable to convert response to map ({}, {})",
```
although it's tempting to extend it to include the request method as well, like:
```
      log.debug(
          "unable to convert response to map ({} {}, {})",
          retrofit2Response.raw().request().method(),
          retrofit2Response.code(),
          retrofit2Response.raw().request().url(),
          e);
```

dbyron-sf:
@Pranav-b-7 This looks great, thank you.  For completeness, would you mind adding a test to SpinnakerHttpExceptionTest that uses the retrofit2 constructor with a non-json response?

kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpException.java
dbyron-sf: There are more places where it's interesting to use the request method so perhaps let's leave that to a followup PR.

Pranav-b-7:
> @Pranav-b-7 This looks great, thank you. For completeness, would you mind adding a test to SpinnakerHttpExceptionTest that uses the retrofit2 constructor with a non-json response?

Added the test case to invoke the SpinnakerHttpException constructor with retrofit2 parameters to verify the parse failure of a non json error response body

kork-retrofit/src/test/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerHttpExceptionTest.java
```diff
@@ -147,4 +155,53 @@ void testNonJsonErrorResponse() {
     assertThat(spinnakerHttpException.getUrl()).isEqualTo(url);
     assertThat(spinnakerHttpException.getReason()).isEqualTo(reason);
   }
+
+  @Test
+  void testSpinnakerHttpExceptionRetrofit2NonJsonHttpErrorResponse() {
+
+    String invalidJsonTypeResponseBody = "{'errorResponse': 'Failure'";
```
dbyron-sf, line 162: OK, if we have to work this hard to call the retrofit2 SpinnakerHttpException constructor directly, I don't think it's worth it.  People looking for tests of this functionality will have to dig enough to find SpinnakerRetrofit2ErrorHandlerTest.  Sorry for the thrashing, but please remove this test.

Pranav-b-7: Reverted the test case commit

PR #1108, opened 2023-10-26
fix(retrofit): Add exception handler to service factory

xibz:
This commit adds the SpinnakerRetroExceptionHandler when creating service clients which allows us to return a proper exception back instead of relying on RetrofitError.

This commit also moves the UpstreamBadRequest to the more appropriate package of kork-retrofit. This allows for other classes within kork-retrofit to use it without creating a circular dependency. In addition this also makes the UpstreamBadRequest class serializable by not passing the RetrofitError into the super call.

### DIFF
--- a/kork-retrofit/kork-retrofit.gradle
+++ b/kork-retrofit/kork-retrofit.gradle
@@ -22,6 +22,8 @@ dependencies {
   testImplementation "org.springframework.boot:spring-boot-starter-test"
   testImplementation "com.netflix.spectator:spectator-reg-micrometer"
   testImplementation "com.squareup.okhttp3:mockwebserver"
+  testImplementation "com.github.tomakehurst:wiremock-jre8-standalone"
+  testImplementation "ru.lanwen.wiremock:wiremock-junit5:1.2.0"
   testRuntimeOnly "cglib:cglib-nodep"
   testRuntimeOnly "org.objenesis:objenesis"
 }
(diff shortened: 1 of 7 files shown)


kork-retrofit/kork-retrofit.gradle
```diff
@@ -22,6 +22,8 @@ dependencies {
   testImplementation "org.springframework.boot:spring-boot-starter-test"
   testImplementation "com.netflix.spectator:spectator-reg-micrometer"
   testImplementation "com.squareup.okhttp3:mockwebserver"
+  testImplementation "com.github.tomakehurst:wiremock-jre8-standalone"
+  testImplementation "ru.lanwen.wiremock:wiremock-junit5:1.2.0"
```
dbyron-sf, line 26: Perhaps not in this PR, but this seems worth pinning in spinnaker-dependencies since there are various versions sprinkled around (e.g. [here](https://github.com/spinnaker/orca/blob/b1071bd1725c6c73eb06465e1529eb421aa594cf/orca-clouddriver/orca-clouddriver.gradle#L61), [here](https://github.com/spinnaker/clouddriver/blob/9beb889c961d8231a8290b3b18d663f787bb33a4/clouddriver-artifacts/clouddriver-artifacts.gradle#L68) and [here](https://github.com/spinnaker/clouddriver/blob/9beb889c961d8231a8290b3b18d663f787bb33a4/clouddriver-cloudfoundry/clouddriver-cloudfoundry.gradle#L77)).

xibz:
Closing in favor of: https://github.com/spinnaker/kork/pull/1185

PR #1109, opened 2023-10-28
update(retrofit): Adds url to SpinnakerServerException

xibz:
This adds the url field to the SpinnakerServerException class allowing
for igor and any other service to easily retrieve the url.


dbyron-sf:
Looks like spotless is complaining.  What do you think of adding a test that uses a "real" retrofit interface + SpinnakerRetrofitErrorHandler to end up with SpinnakerServerException and demonstrate that the url is populated properly?

PR #1011, opened 2023-01-25
chore(dependencies): Upgrade Spring Boot to 2.5.14 

j-sandy:
Spring boot 2.5.x brings groovy 3.x as its transitive dependency.
https://docs.spring.io/spring-boot/docs/2.5.14/reference/html/dependency-versions.html#appendix.dependency-versions

Currently spinnaker services use gradle 6.x, that does not support groovy 3.x.
https://docs.gradle.org/6.8.1/userguide/compatibility.html

Restricting groovy to 2.5.x using the code snippet suggested below, till upgrade of gradle to 7.x.
https://docs.gradle.org/current/userguide/resolution_rules.html#sec:denying_version
##############################################################################

fix(dependency): Issue with kork-jedis while upgrading spring-boot to 2.5.14

While upgrading the spring-boot, the compilation of kork-jedis module failed with following error:
```
> Task :kork-jedis:compileJava FAILED
/kork/kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/telemetry/InstrumentedJedis.java:2652: error: slowlogGetBinary(long) in InstrumentedJedis cannot implement slowlogGetBinary(long) in AdvancedBinaryJedisCommands
  public List<byte[]> slowlogGetBinary(long entries) {
                      ^
  return type List<byte[]> is not compatible with List<Object>
/kork/kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/telemetry/InstrumentedJedis.java:2646: error: slowlogGetBinary() in InstrumentedJedis cannot implement slowlogGetBinary() in AdvancedBinaryJedisCommands
  public List<byte[]> slowlogGetBinary() {
                      ^
  return type List<byte[]> is not compatible with List<Object>
/kork/kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/telemetry/InstrumentedJedis.java:2645: error: method does not override or implement a method from a supertype
  @OverRide
  ^
/kork/kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/telemetry/InstrumentedJedis.java:2648: error: incompatible types: inference variable T has incompatible bounds
    return instrumented(command, () -> delegated.slowlogGetBinary());
                       ^
    lower bounds: List<byte[]>,Object
    lower bounds: List<Object>
  where T is a type-variable:
    T extends Object declared in method <T>instrumented(String,Callable<T>)
[...]

PR #1111, opened 2023-10-30
minor bug fix - changed duplicate logback-classic dependency to logback-access

kirangodishala:
By mistake one of the PRs replaced logback-access to logback-classic. This PR is to revert that change.


dbyron-sf (approved):
Yup, it was https://github.com/spinnaker/kork/pull/1011/files#diff-4f290478509c316a6dadfbd91cbf74598bd0969cf02a9bd7eb796624fe043ec6R98 that duplicated classic.

PR #1112, opened 2023-11-01
config(crypto): Update the default TLS cipher suites

jvz:
This adds a common `CipherSuites` class for listing various profiles of cipher suites for use with TLS. The default cipher suites and TLS versions are updated to a slightly more secure version of the intermediate compatibility baseline from the Mozilla recommendations. These settings are used for both Tomcat and OkHttpClient. Of particular note, this adds in the required ciphers in TLS 1.3 to improve compatibility with modern clients and servers.

### DIFF
--- a/kork-tomcat/kork-tomcat.gradle
+++ b/kork-tomcat/kork-tomcat.gradle
@@ -8,6 +8,7 @@ dependencies {
   implementation("com.netflix.spectator:spectator-api")
   implementation("com.google.guava:guava")
   implementation(project(":kork-core"))
+  implementation(project(":kork-crypto"))
 
   testImplementation "org.springframework.boot:spring-boot-starter-test"
 }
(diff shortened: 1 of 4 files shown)


jvz:
Of extra particular note: without these settings, technically, TLS 1.3 is only partially supported as the mandatory ciphers from 1.3 were not all present in the default list (which is understandable since some of said ciphers weren't added to standard Java until Java 11). For example, if you connect through a proxy that supports one cipher that Spinnaker doesn't, it's possible to end up confusing middleware when Spinnaker is the one violating the spec :)

kork-tomcat/kork-tomcat.gradle
```diff
@@ -8,6 +8,7 @@ dependencies {
   implementation("com.netflix.spectator:spectator-api")
   implementation("com.google.guava:guava")
   implementation(project(":kork-core"))
+  implementation(project(":kork-crypto"))
```
jasonmcintosh, line 11: Wondering... api vs. implementation?  I'd lean towards API to expose crypto to downstream stuff.

jasonmcintosh:
Overall I'm inclined to approve and shrink down & remove some old ciphers.  That said would like some release notes to call this out in case people are doing strange things like using TLS1.1 certs... which... dear god I hope not!

jasonmcintosh (approved):

PR #1114, opened 2023-11-06
fix(headers): avoid retrofit and okhttp adding duplicate headers

dmart:
With the addition of the okHttp3 request interceptor `SpinnakerRequestHeaderInterceptor` in https://github.com/spinnaker/kork/pull/1004 having retrofit1 add `X-SPINNAKER-*` headers is now redundant, this results in the headers being added twice to all retrofit1 client requests.


dmart:
For some requests the duplicate headers break Fiat auth calls which show up like the following, however I am unable to determine why some requests break and not others:
`FiatPermissionEvaluator   : [david.martin@****, david.martin@****] Cannot get whole user permission for user david.martin@****, david.martin@****, reason: 404`

dbyron-sf:
I suspect you're right about this.  Would you mind adding a test that demonstrates it?

dmart:
@dbyron-sf Added a test

kork-retrofit/src/test/kotlin/com/netflix/spinnaker/kork/retrofit/RetrofitServiceProviderTest.kt
```diff
@@ -29,22 +29,33 @@ import com.netflix.spinnaker.config.DefaultServiceClientProvider
...
 import org.springframework.context.annotation.Configuration
+import org.springframework.context.annotation.Primary
 import retrofit.Callback
 import retrofit.http.GET
 import retrofit.http.Path
 import strikt.api.expect
 import strikt.assertions.isA
 import strikt.assertions.isEqualTo
+import strikt.assertions.isNotNull
+import java.util.*
```
dbyron-sf, line 57: please avoid star imports

kork-retrofit/src/test/kotlin/com/netflix/spinnaker/kork/retrofit/RetrofitServiceProviderTest.kt
```diff
@@ -91,8 +128,20 @@ private open class TestConfiguration {
     HttpTracing.newBuilder(Tracing.newBuilder().build()).build()
 
   @Bean
-  open fun okHttpClient(httpTracing: HttpTracing): OkHttpClient {
-    return RawOkHttpClientFactory().create(OkHttpClientConfigurationProperties(), emptyList(), httpTracing)
+  @Primary
```
dbyron-sf, line 131: `@Primary` always makes me nervous, so I tried taking it out and got an error that there was more than one SpinnakerRequestHeaderInterceptor bean in the context.  Looks like it's coming from [OkHttpClientComponents](https://github.com/spinnaker/kork/blob/7bd9a472d09d5f733af11afe223eeaa68d9daa8c/kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttpClientComponents.java#L88-L91).  So then I removed the whole SpinnakerRequestHeaderInterceptor bean and the tests still passed.

I can go ahead and push the commit to do this, but perhaps there's some reason we need another bean here?  Or you'd like to make the changes yourself?

dbyron-sf: Looks like I don't have my story straight since this PR also removes the bean from OkHttpClientComponents....Lemme keep digging.

dbyron-sf: Aaah, this PR removes `SpinnakerRequestInterceptor` from OkHttpClientComponents, but `SpinnakerRequestHeaderInterceptor` remains, so I think my original comment is correct.

kork-web/src/main/groovy/com/netflix/spinnaker/okhttp/SpinnakerRequestInterceptor.groovy
```diff
@@ -1,41 +0,0 @@
...
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
-package com.netflix.spinnaker.okhttp
-
-import com.netflix.spinnaker.security.AuthenticatedRequest
-import retrofit.RequestInterceptor
-
-class SpinnakerRequestInterceptor implements RequestInterceptor {
```
dbyron-sf, line 22: Comments like [this one in MdcCopyingAsyncTaskExecutor](https://github.com/spinnaker/kork/blob/7bd9a472d09d5f733af11afe223eeaa68d9daa8c/kork-web/src/main/java/com/netflix/spinnaker/kork/web/context/MdcCopyingAsyncTaskExecutor.java#L33) could use a tweak now that SpinnakerRequestHeaderInterceptor is doing the copying.

I also see some other places that use SpinnakerRequestInterceptor that are going to need adjusting:
* echo: [IgorConfig](https://github.com/spinnaker/echo/blob/7b1db17df4acc7b57e935ef96c87cea9aea2e4eb/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/config/IgorConfig.java#L50), [Front50Config](https://github.com/spinnaker/echo/blob/7b1db17df4acc7b57e935ef96c87cea9aea2e4eb/echo-web/src/main/java/com/netflix/spinnaker/echo/config/Front50Config.java#L54)
* [fiat](https://github.com/spinnaker/fiat/blob/master/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAuthenticationConfig.java#L69)
* [orca](https://github.com/spinnaker/orca/blob/7e1e74f90b5d9ab786d2a5efd18100a4a9a794c7/orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/config/CloudDriverConfigurationTest.java#L81)

Can you make draft PRs to resolve those, or...maybe we need some more discussion about how to resolve them?

dmart: Maybe removing the bean entirely is not the way to go, perhaps marking it deprecated and simply having `intercept` noop so we don't break other projects? I don't love either solution.

dbyron-sf: Yeah, I don't love that either.  I'd love to at least explore what it would be like to remove it.  Or at least what the path is to stop using it.

dbyron-sf:
Related slack thread: https://spinnakerteam.slack.com/archives/C091CCWRJ/p1734041821986099

PR #1117, opened 2023-11-13
chore(dependencies): upgrade liquibase to 4.24.0

kirangodishala:
This liquibase upgrade PR 
- fixes vulnerabilities(CVE-2022-0839 and many from dependencies) and 
- addresses the limitations of using any liquibase version higher than 3.10.3.  

Some background is given below:
- The liquibase version starting from 4.0.0 till 4.12.0 has an issue w.r.t parsing the changelog file,  if found at multiple places within the classpath - https://github.com/liquibase/liquibase/issues/2818
- This duplicate changelog issue is fixed in 4.13.0, but identified another issue that gets introduced in 4.13.0. This issue(https://github.com/liquibase/liquibase/issues/3091) hindered the migration of sql scripts available in orca (https://github.com/spinnaker/orca/tree/master/orca-sql/src/main/resources/db/changelog), containing `afterColumn` construct, with a validation error for postgresql. 
- `addAfterColumn is not allowed on postgresql` issue is fixed in 4.24.0 (https://github.com/liquibase/liquibase/pull/3978)

After merging this PR, Orca's changelog files will need some changes as per the liquibase fix introduced in 4.24.0.


kork-sql/src/main/kotlin/com/netflix/spinnaker/kork/sql/migration/SpringLiquibaseProxy.kt
```diff
@@ -41,6 +43,7 @@ class SpringLiquibaseProxy(
   init {
     changeLog = "classpath:db/changelog-master.yml"
     dataSource = createDataSource()
+    labelFilter = "test" // this field is added to align with Liquibase versions 4.21.0 and above
```
dbyron-sf, line 46: Why "test" as opposed to some other value?

kirangodishala: This labelFilter is supposed to be used for running all those changeLogs having that label on them. But liquibase also runs all those which don't have any label. And by not assigning any value to that field either in SpringLiquibaseProxy object or SpringLiquibase object, liquibase is ignoring all those changeLogs defined through SpringLiquibase objects under SpringLiquibaseProxy's  afterPropertiesSet() method. 

So, if there is better text to assign to labelFilter, I can update it.

dbyron-sf: "test" does feel a little strange to have in src/main.  Perhaps "spinnaker"?  Including this explanation in the comment in the code would also help.  It's way more informative than "align with Liquibase...".

kirangodishala: updated accordingly.

kork-sql/src/main/kotlin/com/netflix/spinnaker/kork/sql/migration/SpringLiquibaseProxy.kt
```diff
@@ -41,6 +43,13 @@ class SpringLiquibaseProxy(
   init {
     changeLog = "classpath:db/changelog-master.yml"
     dataSource = createDataSource()
+    // Liquibase 4.21.0 onwards, there is a check in liquibase based on labelFilter value.
+    // This labelFilter is supposed to be used for running all those changeLogs having that label on them.
+    // But liquibase also runs all those which don't have any label.
```
dbyron-sf, line 48: Sorry to keep hammering on this, but...if we don't give it a label, and liquibase runs changelogs that don't have a label...then, don't things "just work"?

kirangodishala: ok, I see the confusion and so now I improvised the comment. Liquibase executes only one ChangeLog file(except if it's root changelog having includAll tag) but SpringLiquibaseProxy is designed to make liquibase run multiple changeLog files. Liquibase runs all the `changesets` having given label and also those `changeSets` having no label(all belong to same ChangeLog). But here the challenge is that liquibase is ignoring the additional `changeLog` files if we don't provide labelFilter.

kirangodishala: After testing different options found that the issue occurs only if one of the changeLog files has no changeSets. Which means, this issue is only with the test case. The changelogs of the clouddriver and orca are running fine without this change. An alternate approach to fix the test is to add a changeSet in the changelog-master.yml so we don't need to add labelFilter in SpringLiquibaseProxy.

dbyron-sf: If there's a way to make things work by only changing changelog-master.yml in kork, with no change here, let's do it.

kirangodishala: @dbyron-sf - I made changes in the changelog-master.yml and removed labelFilter from SpringLiquibaseProxy. No errors found with clouddriver and orca with that.

dbyron-sf (approved):

dbyron-sf (approved):

PR #1120, opened 2023-11-16
fix(artifacts): always retrieve s3 artifacts even when artifact-store.s3.enabled is false

dbyron-sf:
so changes to the config value while pipelines are executing, or actions on completed
pipelines (e.g. "Baked Manifest YAML") work without disruption.

Consider this sequence of events:

1. artifact-store.s3.enabled is true
2. a pipeline starts executing,
3. the pipeline execution context contains an artifact store reference
4. artifact-store.s3.enabled changes to false
5. the pipeline continues to execute and attempts to retrieve the artifact store reference

Before this change, step 5 would fail.  After this change, it succeeds.

This PR also removes the `artifact-store.enabled` flag, adds `artifact-store.type`, and changes the meaning of `artifact-store.s3.enabled`.

To enable both stores to and gets from s3:
```
artifact-store:
  type: s3
  s3:
    enabled: true
    region: <region>
    bucket: <bucket>
```
To enable gets from s3, but not stores:
```
artifact-store:
  type: s3
  s3:
    enabled: false
    region: <region>
    bucket: <bucket>
```

and more generally, there is always an ArtifactStore bean in the context, even if no artifact-store config flags are defined.

orca needs some bean naming tweaks to keep up with this.


kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfiguration.java
```diff
@@ -55,51 +56,73 @@ public ArtifactStoreURIBuilder artifactStoreURIBuilder() {
...
       ArtifactStoreURIBuilder artifactStoreURIBuilder) {
+    return new S3ArtifactStoreStorer(
+        s3Client,
+        properties.getS3().getBucket(),
+        artifactStoreURIBuilder,
+        properties.getApplicationsRegex());
+  }
+
+  @Bean
+  public S3ArtifactStoreGetter s3ArtifactStoreGetter(
```
xibz, line 88: Rather than having the bucket ternary, can we do something like
`@ConditionalOnProperty(name = "artifact-store.s3")`?

Because if bucket is not specify, it will fail gets no matter what, so may make sense just to see if s3 is configured. If not, just don't create the getter.

I wonder if that also hold's true for the `artifactS3Client` bean as well, so it might clean that up too.

We could probably even move all this s3 bootstrapping to an `S3ArtifactStoreConfiguration` and just have the `@ConditionalOnProperty` on there

xibz (changes requested):
Thanks for the PR, @dbyron-sf! Most of my comments are about moving things around and may trying a different approach. Let me know if you have any questions about the suggestions

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfiguration.java
```diff
@@ -55,51 +56,73 @@ public ArtifactStoreURIBuilder artifactStoreURIBuilder() {
...
+      ArtifactStoreGetter artifactStoreGetter, ArtifactStoreStorer artifactStoreStorer) {
+    return new ArtifactStore(artifactStoreGetter, artifactStoreStorer);
+  }
+
+  @Bean
+  @ConditionalOnProperty(
+      value = "artifact-store.s3.enabled",
+      havingValue = "false",
+      matchIfMissing = true)
+  public NoopArtifactStoreStorer noopArtifactStoreStorer() {
```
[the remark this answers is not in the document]
xibz: If we allow for multiple storers this will need to change a little bit, but we can cross that bridge when we get there.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/NoopArtifactStoreStorer.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.artifacts.artifactstore;
+
+import com.netflix.spinnaker.kork.artifacts.model.Artifact;
+
+/** A no-op ArtifactStoreStorer. In other words, don't actually store the artifact. */
+public class NoopArtifactStoreStorer implements ArtifactStoreStorer {
```
xibz, line 21: I would probably also just change this to `NoopArtifactStore implements ArtifactStoreGetter, ArtifactStoreGetter` and do both

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfiguration.java
xibz: Further you can then utilize the `NoopArtifactStore` on both methods: store and get

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreStorer.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package com.netflix.spinnaker.kork.artifacts.artifactstore;
+
+import com.netflix.spinnaker.kork.artifacts.model.Artifact;
+
+/**
+ * ArtifactStoreStorer is an interface that allows for different types of artifact storage to be
+ * used during runtime.
+ */
+public interface ArtifactStoreStorer {
```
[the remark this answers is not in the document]
dbyron-sf: Good point.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/NoopArtifactStoreStorer.java
dbyron-sf: In case we have a properly configured (but disabled) s3 artifact store, we use NoopArtifactStoreStorer along with S3ArtifactStoreGetter.

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
```diff
@@ -15,233 +15,8 @@
...
-import software.amazon.awssdk.services.s3.model.GetObjectResponse;
-import software.amazon.awssdk.services.s3.model.GetObjectTaggingRequest;
-import software.amazon.awssdk.services.s3.model.GetObjectTaggingResponse;
-import software.amazon.awssdk.services.s3.model.HeadObjectRequest;
-import software.amazon.awssdk.services.s3.model.NoSuchKeyException;
-import software.amazon.awssdk.services.s3.model.PutObjectRequest;
-import software.amazon.awssdk.services.s3.model.S3Exception;
-import software.amazon.awssdk.services.s3.model.Tag;
-import software.amazon.awssdk.services.s3.model.Tagging;
+public final class S3ArtifactStore {
```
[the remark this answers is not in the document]
dbyron-sf: OK, but where?

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/NoopArtifactStoreStorer.java
xibz: I just meant in relation to the comment below where if you dont have a proper s3 config, you cant get either. So should probably noop that as well and potentially warn on no s3 config

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfiguration.java
```diff
@@ -55,51 +56,73 @@ public ArtifactStoreURIBuilder artifactStoreURIBuilder() {
...
+    return new NoopArtifactStoreStorer();
+  }
+
   @Bean
   @ConditionalOnExpression("${artifact-store.s3.enabled:false}")
-  public ArtifactStore s3ArtifactStore(
-      Optional<PermissionEvaluator> permissionEvaluator,
+  public S3ArtifactStoreStorer s3ArtifactStoreStorer(
```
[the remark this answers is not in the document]
dbyron-sf: https://github.com/spinnaker/kork/pull/1120/commits/af4c8c891a56f0ee618b936d5fcacc13561f6232

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfiguration.java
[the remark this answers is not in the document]
dbyron-sf: https://github.com/spinnaker/kork/pull/1120/commits/af4c8c891a56f0ee618b936d5fcacc13561f6232

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfiguration.java
dbyron-sf: https://github.com/spinnaker/kork/pull/1120/commits/af4c8c891a56f0ee618b936d5fcacc13561f6232

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/NoopArtifactStoreStorer.java
dbyron-sf: https://github.com/spinnaker/kork/pull/1120/commits/af4c8c891a56f0ee618b936d5fcacc13561f6232

kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStore.java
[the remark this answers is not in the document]
xibz: I think we can move this to an enum, but it does not have to be this PR. I have a couple upcoming PRs where I can do that

xibz (approved):
Thanks for all the changes @dbyron-sf! Especially dealing with all the flags! Good work :). This looks good.

We were thinking of making this a bug fix and backporting it as well? I wonder if we should also mention it in the changelogs as this is technically a breaking change with the new `type` needing to be set.

PR #1126, opened 2023-12-05
chore(dependencies): add dependencies on spinnaker-dependencies

dbyron-sf:
When projects have dependencies on artifacts, but depend on spinnaker-dependencies to provide the version, this makes the version available and removes failures from ./gradlew dependencies output.

Some examples of failures that this removes:
```
testRuntimeOnlyDependenciesMetadata
+--- org.junit.jupiter:junit-jupiter-engine FAILED \--- org.junit.platform:junit-platform-launcher FAILED

compileOnly - Compile only dependencies for compilation 'main' (target  (jvm)). \--- org.projectlombok:lombok FAILED

compileOnlyDependenciesMetadata
\--- org.projectlombok:lombok FAILED

testCompileOnly - Compile only dependencies for compilation 'test' (target  (jvm)). \--- org.projectlombok:lombok FAILED

testCompileOnlyDependenciesMetadata
\--- org.projectlombok:lombok FAILED
```

### DIFF
--- a/kork-actuator/kork-actuator.gradle
+++ b/kork-actuator/kork-actuator.gradle
@@ -19,6 +19,7 @@ apply plugin: "java-library"
 apply from: "$rootDir/gradle/kotlin-test.gradle"
 
 dependencies {
+  compileOnly(platform(project(":spinnaker-dependencies")))
   implementation(platform(project(":spinnaker-dependencies")))
   implementation "org.springframework.security:spring-security-core"
   implementation "org.springframework.boot:spring-boot-starter-security"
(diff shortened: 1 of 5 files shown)


jasonmcintosh (approved):

kork-actuator/kork-actuator.gradle
```diff
@@ -19,6 +19,7 @@ apply plugin: "java-library"
 apply from: "$rootDir/gradle/kotlin-test.gradle"
 
 dependencies {
+  compileOnly(platform(project(":spinnaker-dependencies")))
```
jasonmcintosh, line 22: ONLY thought was that the below implementation dep would be precedence on this vs. compileOnly.  Kinda surprised this one was needed....

dbyron-sf: Yeah, I was also surprised.  But without it, there are failures.

PR #1132, opened 2023-12-14
feat(web): Improve generic exception handling

jvz:
As the default Spring exception handlers are great at hiding relevant error details, this overrides the default exception handler from `ResponseEntityExceptionHandler`.


mattgogerly (approved):

jvz:
May not be relevant for now.

PR #1138, opened 2023-12-20
feat(security): Add more Fiat abstractions

jvz:
Related to https://github.com/spinnaker/spinnaker/issues/6911

This adds a copy of the Authorization enum from fiat-core, an AbstractPermissionEvaluator base class which fiat-api will be updated to use, some extensions to AccessControlled, and some other security utility functions.


kork-security/src/main/java/com/netflix/spinnaker/security/AbstractPermissionEvaluator.java
```diff
@@ -0,0 +1,79 @@
...
+ * Base implementation for permission evaluators that support {@link AccessControlled} domain
+ * objects.
+ */
+public abstract class AbstractPermissionEvaluator implements PermissionEvaluator {
+
+  @Override
+  public boolean hasPermission(
+      Authentication authentication, Object targetDomainObject, Object permission) {
+    if (isDisabled()) {
+      return true;
```
jvz, line 34: I'll note that this is going to be a small change in behavior to Fiat based on a feature introduced quite a while back. Since this method used to be a stub that always returns `false`, but the feature flag used to control this is something I've been thinking should be enabled permanently after using this concept for over a year. Being able to make access control decisions based on the domain object directly speeds things up regarding the eventual consistency otherwise caused by Fiat using periodic synchronization to keep permissions up to date.

kork-security/src/main/java/com/netflix/spinnaker/security/Authorization.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Defines types of authorizations supported by {@link AccessControlled#isAuthorized(Authentication,
+ * Object)}.
+ */
+public enum Authorization {
+  READ,
+  WRITE,
+  EXECUTE,
+  CREATE,
```
jvz, line 29: Previously, I didn't have the `CREATE` permission defined in this Kork version of the class, but I figured it'd be simpler to mirror the permissions exactly from Fiat for ease of implementation.

kork-security/src/main/java/com/netflix/spinnaker/security/PermissionMapControlled.java
```diff
@@ -0,0 +1,52 @@
...
+import javax.annotation.Nullable;
+import org.springframework.security.core.Authentication;
+
+/**
+ * Common interface for access-controlled classes which use a permission map.
+ *
+ * @param <Authorization> Authorization enum type
+ */
+@Alpha
+public interface PermissionMapControlled<Authorization extends Enum<Authorization>>
```
jvz, line 32: Strange use of generics here `Authorization` instead of something simple like `A` to emphasize that this should be either of the `Authorization` enum classes. Since enums are final classes, we can't make Fiat extend Kork, hence the generic-based workaround here.

kork-security/src/main/java/com/netflix/spinnaker/security/SpinnakerUsers.java
```diff
@@ -0,0 +1,45 @@
...
+import com.netflix.spinnaker.kork.annotations.NonnullByDefault;
+import javax.annotation.Nullable;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContextHolder;
+
+/** Constants and utilities related to Spinnaker users (AKA principals). */
+@NonnullByDefault
+public class SpinnakerUsers {
+  /** String constant for the anonymous userid. */
+  public static final String ANONYMOUS = "anonymous";
```
jvz, line 29: Will note for future reference: this constant is made available in order for the various Spinnaker services to _consistently_ configure anonymous authentication. Turns out that Gate in particular uses a different username for anonymous by accident (default Spring Security config), so that's yet another thing I'll be fixing.

kork-security/src/test/groovy/com/netflix/spinnaker/security/AccessControlledSpec.groovy
```diff
@@ -0,0 +1,67 @@
...
+
+import static groovy.lang.Tuple.tuple
+
+class AccessControlledSpec extends Specification {
+  static class TestCredentials implements AuthorizationMapControlled {
+    Map<Authorization, Set<String>> permissions
+  }
+
+  static class TestPermissionEvaluator extends AbstractPermissionEvaluator {
+    Map<Tuple2<Serializable, String>, Map<String, Collection<Object>>> targetIdTypesToUsernamePermissions = [:]
```
jvz, line 31: If the type here is hard to grok, I'm indexing test permissions by a resource id/type pair, then indexing permissions by username. Hence the field name.

kork-security/src/main/java/com/netflix/spinnaker/security/Authorization.java
jasonmcintosh: Good plan :) I like CREATE as a separate permission as a rule!

kork-security/src/main/java/com/netflix/spinnaker/security/AuthorizationMapControlled.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Common interface for access-controlled classes which use a permission map of {@link
+ * Authorization} enums.
+ */
+public interface AuthorizationMapControlled extends PermissionMapControlled<Authorization> {
+  @Nullable
+  @Override
+  default Authorization valueOf(@Nullable Object authorization) {
+    return authorization != null
+        ? Enum.valueOf(Authorization.class, authorization.toString().toUpperCase(Locale.ROOT))
```
jasonmcintosh, line 32: MINOR caution.  This can throw an exception here on the valueOf - probably should catch & return null vs. hidden throw?  HOPEFULLY won't happen, given sitaution but if something added a new Authorization type... better safe than sorry.

kork-security/src/main/java/com/netflix/spinnaker/security/SpinnakerUsers.java
jasonmcintosh: THANK YOU :)

jvz: Filed as https://github.com/spinnaker/spinnaker/issues/6918

kork-security/src/main/java/com/netflix/spinnaker/security/SpinnakerUsers.java
```diff
@@ -0,0 +1,45 @@
...
+
+  /** Gets the userid of the provided authentication token. */
+  public static String getUserId(@Nullable Authentication authentication) {
+    return authentication != null ? authentication.getName() : ANONYMOUS;
```
jasonmcintosh, line 33: I can't recall if logins did this or it was done elsewhere - but all user identities SHOULD be lower cased.

kork-security/src/main/java/com/netflix/spinnaker/security/AuthorizationMapControlled.java
jvz: Good point; I can update this.

kork-security/src/main/java/com/netflix/spinnaker/security/SpinnakerAuthorities.java
```diff
@@ -0,0 +1,90 @@
...
+    if (authentication == null) {
+      return List.of();
+    }
+    return streamRoles(authentication).distinct().collect(Collectors.toList());
+  }
+
+  @Nonnull
+  private static Stream<String> streamRoles(@Nonnull Authentication authentication) {
+    return authentication.getAuthorities().stream()
+        .filter(SpinnakerAuthorities::isRole)
```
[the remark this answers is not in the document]
jvz: In the granted authority scheme here, all roles start with `ROLE_` (case-sensitive). These are typically transformed to lowercase in either the Gate or Fiat module in use for roles. For OAuth2, I'm not quite sure as I've got an updated version of that module using the built-in Spring Security library for that rather than the old extension library, and in that usage, I'm delegating to Fiat for role information, and in Fiat, the `Role` class always lowercases role names.

kork-security/src/main/java/com/netflix/spinnaker/security/AuthorizationMapControlled.java
jvz: Fixed.

kork-security/src/main/java/com/netflix/spinnaker/security/SpinnakerUsers.java
jvz: I don't think that is done; technically, email addresses are case-sensitive, and usernames likely are, too.

jvz: In fact, email addresses are really weird this way: the left hand side of the address is case-sensitive, but the domain name is case-insensitive!

jasonmcintosh: TECHNICALLY yes on username - domain names never supported case :) So not end of the world, just KNOW this has bit some users and a few places.  Wondering if should be standard to BE case insensitive....

jvz: That seems to be a Fiat-related responsibility.

jasonmcintosh: For reference: https://github.com/spinnaker/fiat/pull/818

Going to leave this alone for now and approve - case sensitive has just bit things badly in past.

jasonmcintosh (approved):

PR #1139, opened 2023-12-20
feature(credentials-api): CredentialsLoader extensions

jvz:
This adds a CredentialsLoader interface extracted from AbstractCredentialsLoader along with additional SpinnakerExtensionPoint classes updated to allow for plugins to define credentials beans.

Related to https://github.com/spinnaker/spinnaker/issues/6914


jvz:
Not sure if this should be a "refactor" change instead. Main idea here is exposing things as extension points and the extraction of the aforementioned interface.

dbyron-sf (approved):

PR #1145, opened 2024-02-03
fix(artifacts): Addresses a couple things regarding artifact store

xibz:
This PR has three commits which addresses 3 particular issues:

1. README updates to reflect on current artifact store usage
2. Moving of scheme from the builder to the `ArtifactReferenceURI` class instead as that makes more sense
3. Updates the artifact store to rely more easily on default values instead of requiring `null` checks for the individual storage types

Signed-off-by: benjamin-j-powell <bjp@apple.com>

### DIFF
--- a/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfigurationProperties.java
+++ b/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfigurationProperties.java
@@ -44,5 +44,12 @@ public class ArtifactStoreConfigurationProperties {
     private boolean forcePathStyle = true;
   }
 
-  private S3ClientConfig s3 = null;
+  @Data
+  public static class HelmConfig {
+    /** Enables Rosco to expand any artifact URIs passed as parameters for Helm. */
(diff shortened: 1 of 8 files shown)


kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfigurationProperties.java
```diff
@@ -44,5 +44,12 @@ public static class S3ClientConfig {
...
 
-  private S3ClientConfig s3 = null;
+  @Data
+  public static class HelmConfig {
+    /** Enables Rosco to expand any artifact URIs passed as parameters for Helm. */
+    private boolean expandOverrides = false;
+  }
+
+  private S3ClientConfig s3 = new S3ClientConfig();
+  private HelmConfig helm = new HelmConfig();
```
dbyron-sf, line 54: It's a bit of a bummer this is here since it's specific to rosco....but it's probably the least bad thing.  Adding the config to rosco means there are multiple spots making artifact-store config props and that's either bound to be confusing or lead to conflicts or some other badness.

dbyron-sf (approved):

PR #1149, opened 2024-02-19
feat(retrofit): Make HTTP method available in SpinnakerServerException

abe21412:
This PR adds the HTTP request method as an instance variable to `SpinnakerServerException`. The value is retrieved from the constructor params except in the case that the exception is created from a `RetrofitError` where the HTTP method is not available and thus set to `null` in the exception. There are also some test updates included to make sure the getter works as expected for server, HTTP, and network exceptions.


dbyron-sf:
This looks great!  Mind adding some assertions to make it clear that method is null when dealing with RetrofitError?

dbyron-sf (approved):

PR #1155, opened 2024-02-20
chore(dependencies): use version 9.0.83 of tomcat to fix CVE-2023-46589

dbyron-sf:


dbyron-sf:
> This looks good. Did you test this locally? Just want to make sure this dependency upgrade doesn't have surprises :p

Yup, been running it for awhile here.

PR #1156, opened 2024-02-21
feat(sql): make DefaultSqlConfiguration's transaction isolation configurable

dbyron-sf:
via two new properties.  The default behavior is the same as before this change.

- sql.setTransactionIsolation (defaults to true) -- only set transaction isolation when true
- sql.transactionIsolation (defaults to Connection.TRANSACTION_READ_COMMITTED) -- the transaction isolation level to set


j-sandy:
@Mergifyio update

dbyron-sf:
@mergifyio update

j-sandy (approved):

PR #1134, opened 2023-12-15
chore(dependency): upgrade spring boot from 2.5.15 to 2.6.15 and spring cloud from 2020.0.x to 2021.0.x

j-sandy:
Spring cloud release 2021.0.x is compatible with spring boot 2.6.x. 
https://github.com/spring-cloud/spring-cloud-release/wiki/Supported-Versions#supported-releases

While upgrading spring boot 2.6.15 and spring cloud 2021.0.8, encounter below errors in kork-plugins and kork-tomcat modules:

```
> Task :kork-plugins:compileTestKotlin FAILED
e: /kork/kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/SpringPluginStatusProviderTest.kt: (50, 11): Type mismatch: inferred type is Array<???> but SpringApplication! was expected
e: /kork/kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/SpringPluginStatusProviderTest.kt: (51, 11): Type mismatch: inferred type is ConfigurableEnvironment but Array<(out) String!>! was expected
e: /kork/kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/SpringPluginStatusProviderTest.kt: (52, 9): No value passed for parameter 'environment'
```
The root cause is the deprecation and removal of `org.springframework.boot.context.event.ApplicationEnvironmentPreparedEvent` constructor without `org.springframework.boot.SpringApplication` in spring boot 2.6.x. 
https://docs.spring.io/spring-boot/docs/2.5.15/api/org/springframework/boot/context/event/ApplicationEnvironmentPreparedEvent.html 
To fix this error add mock for `SpringApplication` in `ApplicationEnvironmentPreparedEvent` constructor.

```
DEBUG --- [o-auto-2-exec-1] o.a.coyote.http11.Http11InputBuffer      : Before fill(): parsingHeader: [true], parsingRequestLine: [true], parsingRequestLinePhase: [0], parsingRequestLineStart: [0], byteBuffer.position(): [0], byteBuffer.limit(): [0], end: [0]
DEBUG --- [o-auto-2-exec-1] o.a.coyote.http11.Http11InputBuffer      : Received [GET /test-controller HTTP/1.1
Accept: text/plain, application/json, application/*+json, */*
X-Dum@my: foo
User-Agent: Java/11.0.2
Host: localhost:35189
Connection: keep-alive

]
[...]

PR #1163, opened 2024-03-05
fix(cve): Update snakeyml to 1.31 to fix following CVEs CVE-2022-25857 CVE-2022-38749

ovidiupopa07:
We've tested this in our Harness internal environments for some time and have not seen any misbehavior.


dbyron-sf (changes requested):
I'd rather not do this yet.  Although it may be working, until we have https://github.com/spring-projects/spring-boot/issues/32228 which comes with spring boot 2.6.12, I feel like we're exposing ourselves to a potential bug when some user has some unexpected thing in a spinnaker config file.

We're quite close to merging https://github.com/spinnaker/kork/pull/1134 and going to spring boot 2.6.15.  Can we wait for that?

dbyron-sf:
handled by https://github.com/spinnaker/kork/pull/1173.

PR #1164, opened 2024-03-05
fix(artifacts): Do not store non-base64 artifacts

xibz:
This commit addresses a bug where storing SpEL can cause major issues down the chain, e.g. expected artifacts.

The easiest solution is to simply not store the artifact, but improvements could be made to evaluate the SpEL, if it makes sense to do


jasonmcintosh (approved):

ovidiupopa07:
@Mergifyio backport release-1.33.x release-1.32.x release-1.31.x

PR #1174, opened 2024-03-29
Upgrade testcontainers version

kirangodishala:
Upgrade testcontainers version to the latest 1.19.6.


kirangodishala:
Except clouddriver no other service need any change with this PR. 
Clouddriver changes are captured in this PR - https://github.com/spinnaker/clouddriver/pull/6180
Tested all the services locally with this kork PR; build and tests were successful. 
Echo however has a couple of test failures(at least in local environment) already. So the same tests continue to fail even with kork version having this PR changes.

dbyron-sf (approved):

PR #1180, opened 2024-04-27
fix(artifacts): authenticate against AuthenticatedRequest.getSpinnakerUser in S3ArtifactStoreGetter

dbyron-sf:
instead of SecurityContextHolder.getContext() which might be null.  Previously hasAuthorization would only user userId for logging.  Now it's used for authentication too.  This fixes the bug that https://github.com/spinnaker/kork/pull/1178 demonstrates.

depends on https://github.com/spinnaker/fiat/pull/1155


xibz (approved):
Looks good!

PR #1183, opened 2024-05-15
feat: Ability to easily compare versions using #toArtifactVersion

KristianOellegaard:
If you want to verify that your automatic pipeline does not deploy an old version because it is accidentally pushed to the docker repo

eg

```
#toArtifactVersion(execution.trigger.tag.replace("v", "")) >= #toArtifactVersion(#stage("Find Artifacts From Resource (Manifest)").manifest.spec.template.spec.containers[0].image.split(":")[1].replace("v", ""))
```

### DIFF
--- a/kork-expressions/src/test/java/com/netflix/spinnaker/kork/expressions/ExpressionsSupportTest.java
+++ b/kork-expressions/src/test/java/com/netflix/spinnaker/kork/expressions/ExpressionsSupportTest.java
@@ -163,6 +163,26 @@ public class ExpressionsSupportTest {
     assertThat(evaluated).isEqualTo("00000000-0000-0000-0000-000000000000");
   }
 
+  @Test
+  public void compareArtifactVersion() {
+    // If a thing is not an artifact URI, it should delegate to StandardTypeConverter
+    ExpressionProperties expressionProperties = new ExpressionProperties();
+
+    // StandardTypeConverter does things like convert ints to longs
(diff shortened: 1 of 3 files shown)


kork-expressions/src/test/java/com/netflix/spinnaker/kork/expressions/ExpressionsSupportTest.java
```diff
@@ -163,6 +163,26 @@ public void delegatesTypeConversion() {
...
+    // If a thing is not an artifact URI, it should delegate to StandardTypeConverter
+    ExpressionProperties expressionProperties = new ExpressionProperties();
+
+    // StandardTypeConverter does things like convert ints to longs
+    String testInput = ("${#toArtifactVersion('2.0.0') > #toArtifactVersion('1.0.0')}");
+    Map<String, Object> testContext = Map.of();
+
+    String evaluated =
+        new ExpressionTransform(parserContext, parser, Function.identity())
+            .transformString(
```
jasonmcintosh, line 177: SO this tries to translate the expression to a string.  What's interesting is I DON'T think it's ever actually USED.  Transform MAP is used.  Ignoring that, you can use this but this I believe returns a boolean which it tries to cast as a string and then fails.  Can switch to a transformMap (which is what's REALLY used) OR maybe add a .toString() on the results.  I'd be inclined to do the map based test via a "transformMap(Map.of("test", testInput)...).get("test").  " kinda test.

PR #1198, opened 2024-07-18
feat(swagger): Upgrade to swagger and migrate to springdocs

jasonmcintosh:
This replaces spring fox with spring docs to the latest releases.  
Primary migration guide: https://springdoc.org/migrating-from-springfox.html
MOST Of the changes are done downstream in various services that need to replace springfox resources and older outdated openapi spec APIs.  See: https://github.com/spinnaker/gate/pull/1813/files for an example of this change.  

This can mostly be automated through the use of openrewrite (which that PR documents).  See https://docs.openrewrite.org/recipes/java/springdoc/swaggertospringdoc for the details of what this rewrite recipe applies.  It DID NOT handle most groovy files which require a bit more hand holding on the migration.  This does clean up the dependency tree fairly nicely finally!

TO NOTE:  OpenAPI was completely broken due to various library upgrades (primarily spring fox to 3.0 to fix a CVE).  Per various conversations (e.g. https://github.com/springfox/springfox/issues/3193#issuecomment-558748502) there were attempts to try to make spring fox 3 which has some CVE fixes work with swagger & newer openapis, but it seems to be a struggle to get these to work, and the springfox no longer seems supported.  See https://github.com/springfox/springfox/issues/3834 for another example of "this library is dead and doesn't even work" on latest spring boot releases.


edgarulg (approved):

jasonmcintosh:
@Mergify backport release-1.36.x

PR #1203, opened 2024-09-24
feat(gce): support resourceManagerTags in instance template

edgarulg:
Support `resourceManagerTags` in GCE. See: https://cloud.google.com/compute/docs/reference/rest/v1/instanceTemplates/insert

These changes upgrade the version of `google-api-services-compute` and `google-api-client`.

Issue related: https://github.com/spinnaker/spinnaker/issues/6931

### DIFF
--- a/spinnaker-dependencies/spinnaker-dependencies.gradle
+++ b/spinnaker-dependencies/spinnaker-dependencies.gradle
@@ -95,13 +95,13 @@ dependencies {
       force = true
     }
     api("com.amazonaws:aws-java-sdk:${versions.aws}")
-    api("com.google.api-client:google-api-client:1.31.1") // TODO: Track update for CVE-2020-7692, reanalysis pending.
-    api("com.google.apis:google-api-services-admin-directory:directory_v1-rev105-1.25.0")
-    api("com.google.apis:google-api-services-cloudbuild:v1-rev836-1.25.0")
+    api("com.google.api-client:google-api-client:1.31.5") // TODO: Track update for CVE-2020-7692, reanalysis pending.
+    api("com.google.apis:google-api-services-admin-directory:directory_v1-rev20210531-1.31.5")
+    api("com.google.apis:google-api-services-cloudbuild:v1-rev20210625-1.31.5")
     api("com.google.apis:google-api-services-compute:beta-rev20210525-1.31.5")
-    api("com.google.apis:google-api-services-iam:v1-rev267-1.25.0")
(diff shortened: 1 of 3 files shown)


spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -95,10 +95,10 @@ dependencies {
       force = true
     }
     api("com.amazonaws:aws-java-sdk:${versions.aws}")
-    api("com.google.api-client:google-api-client:1.30.10") // TODO: Track update for CVE-2020-7692, reanalysis pending.
+    api("com.google.api-client:google-api-client:1.31.1") // TODO: Track update for CVE-2020-7692, reanalysis pending.
     api("com.google.apis:google-api-services-admin-directory:directory_v1-rev105-1.25.0")
     api("com.google.apis:google-api-services-cloudbuild:v1-rev836-1.25.0")
-    api("com.google.apis:google-api-services-compute:beta-rev20201102-1.30.10")
+    api("com.google.apis:google-api-services-compute:beta-rev20210525-1.31.5")
```
jasonmcintosh, line 101: My ONLY concern is that google-api-client - does that have any other changes required on the other libraries to get bumped as well?  We seem behind on a lot of places here...

jasonmcintosh: It MAY still be ok would just like confirmation.... :)

edgarulg: I see the dependency. Let me validate that all the libraries are upgraded at the same level to avoid issues.

edgarulg: I upgraded all the google-api-services to use the same api-client:1.31.5

edgarulg:
I published a snapshot version of these changes to upgrade clouddriver and verify everything works as expected

jasonmcintosh (approved):
I'd love more tests or SOME Tests around GcsSecretEngine... but..  we don't today :( and it's a library upgrade with the API change to the SecretEngine being internal, there IS a risk of downstream services having to do the same thing, but that should be caught via compile time updates.  My ask would be post this we make sure anything ELSE using any of the google libraries ALSO gets updated, aka the clouddriver PR and Halyard (and maybe Orca).

edgarulg:
I tested with clouddriver and I only see the `JacksonFactory`  does not exist. I am using GsonFactory as fix but I will check other services as well

PR #1205, opened 2024-09-26
feat(gce): upgrade google-api-client to 1.33.0

edgarulg:
Upgrade google api client to 1.33.0 to allow new features in GCE.

All the google api services were updated to the compatible version with 1.33.0

### DIFF
--- a/spinnaker-dependencies/spinnaker-dependencies.gradle
+++ b/spinnaker-dependencies/spinnaker-dependencies.gradle
@@ -98,13 +98,13 @@ dependencies {
       force = true
     }
     api("com.amazonaws:aws-java-sdk:${versions.aws}")
-    api("com.google.api-client:google-api-client:1.31.5") // TODO: Track update for CVE-2020-7692, reanalysis pending.
-    api("com.google.apis:google-api-services-admin-directory:directory_v1-rev20210531-1.31.5")
-    api("com.google.apis:google-api-services-cloudbuild:v1-rev20210625-1.31.5")
-    api("com.google.apis:google-api-services-compute:beta-rev20210525-1.31.5")
-    api("com.google.apis:google-api-services-iam:v1-rev20210519-1.31.5")
-    api("com.google.apis:google-api-services-monitoring:v3-rev20210618-1.31.5") //v3-rev20210618-1.31.5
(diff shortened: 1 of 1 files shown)


jasonmcintosh:
Time to do some bumps... clouddriver & halyard post this please!

jasonmcintosh:
ALSO:  we JUST had that other bump, did we miss something?  https://github.com/spinnaker/kork/pull/1203/files this one

edgarulg:
> ALSO: we JUST had that other bump, did we miss something? https://github.com/spinnaker/kork/pull/1203/files this one

yes, I upgraded the client to 1.31.5 but the field I want to include from GCE is still not present in that version. So I had to upgrade until 1.33.x in this new PR :( 

The versions of google-api-service-* are weird. They have the wrong client version in their name. Here is an example: https://mvnrepository.com/artifact/com.google.apis/google-api-services-compute/beta-rev20211228-1.32.1 
The versions says that the api-client is 1.32.1 but if you check the dependencies in maven it actually depends on 1.33.0

All the upgrades I did says 1.32.1 but that is not true. Everything is pointing to api-client 1.33.x

edgarulg:
The only breaking changed that I saw on the upgrade is the deprecation of `JacksonFactory`. We use `JacksonFactory` in kork / clouddriver (they might be more services) but we use it to generate a default instance that extends of `JsonFactory`.

In the newer versions we only had `GsonFactory` but as it also extends from `JsonFactory`. We can replace JacksonFactory` to `GsonFactory` and in theory we shouldn't have issues.

jasonmcintosh (approved):
Tied to other PR work.

edgarulg:
Clouddriver PR: https://github.com/spinnaker/clouddriver/pull/6282

edgarulg:
I will bump this change into Harlyard and will share the PR as well

edgarulg:
I tagged kork and all the autobumps were created. The following services had build issue regarding the google upgrade and I pushed the fix for all of them.
https://github.com/spinnaker/clouddriver/pull/6286
https://github.com/spinnaker/fiat/pull/1180
https://github.com/spinnaker/igor/pull/1278
https://github.com/spinnaker/halyard/pull/2186

PR #1209, opened 2024-11-07
feat(flags): Add OpenFeature integration

jvz:
This adds an OpenFeature-based feature flags module along with a built-in provider based on flagd.

As this implements `DynamicConfigService`, all existing dynamic config usage is supported out of the box here. This is primarily to get agreement with the rest of the project that we'll be using that API (along with the OpenFeature API) for feature flags. This should help a lot with experimental features.


kork-core/src/test/groovy/com/netflix/spinnaker/kork/dynamicconfig/OpenFeatureDynamicConfigServiceSpec.groovy
```diff
@@ -0,0 +1,63 @@
...
+package com.netflix.spinnaker.kork.dynamicconfig
+
+import dev.openfeature.sdk.OpenFeatureAPI
+import dev.openfeature.sdk.providers.memory.Flag
+import dev.openfeature.sdk.providers.memory.InMemoryProvider
+import org.springframework.mock.env.MockEnvironment
+import spock.lang.Specification
+import spock.lang.Subject
+
+class OpenFeatureDynamicConfigServiceSpec extends Specification {
```
dbyron-sf, line 27: What are the chances of doing this in java?  We're trying desperately to keep from adding more groovy code.

dbyron-sf:
Could you add some high-level motivation for this?  What are the benefits of OpenFeature?

jvz:
What I find to be a killer benefit is being able to use [targeting](https://openfeature.dev/docs/reference/technologies/server/java/#targeting) for feature flag enablement. While `DynamicConfigService` offers basic support for a few orthogonal targeting criteria, OpenFeature defines a more general feature flag API.

Then, I combined that with [flagd](https://flagd.dev) by using the Java provider as a reference implementation. What I like about flagd (besides being OSS and already well-integrated with OpenFeature) are the existing ways of [defining feature flags](https://flagd.dev/reference/flag-definitions/) which uses some [JsonLogic](https://jsonlogic.com) thing for defining more complex feature flag targeting rules.

Benefits-wise, there are a few different targets. The impetus behind integrating with this is to help enable more continuous delivery of Spinnaker itself by flagging experimental and new stuff behind flags that can be selectively rolled out in different ways. For example, I've been researching how to go about extending Fiat to support external policy engines, but due to the existing mess in Fiat, I've found that very hard to do without affecting a _lot_ of existing code, and I want to be able to selectively roll out changes in authorization based on context like the current user, application, resource type, and so on, and not just in a boolean fashion.

Another advantage is to develop things in a sort of default-on way rather than using conditional beans and other easy to break configuration methods for feature flags. While features are still in alpha, they're not active, but once they're stable, the feature flags can be removed.

dbyron-sf:
Thanks for that @jvz .  I need to digest it a bit.  One thing we have noticed is that, at least with the spring implementation of DynamicConfigService is that it becomes super slow/expensive when used in combination with lots of (clouddriver) accounts if they're defined in yml files.  More generally it slows down with lots of config properties, but lots of accounts is a way to get there.  Basically, any access of one property causes a copy (and perhaps parsing...can't remember) of all properties.  We've since moved accounts to the account API, but for folks who haven't, this could be a drag.  The tl;dr was basically, don't use the DynamicConfigService.  Does this implementation still use enough spring under the covers to suffer this way?

kork-core/src/main/java/com/netflix/spinnaker/kork/BootstrapComponents.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.List;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Import;
+
+@Import({OpenFeatureConfigConfiguration.class, TransientConfigConfiguration.class})
+public class BootstrapComponents {
+  @Bean
+  @ConditionalOnMissingBean(VersionResolver.class)
```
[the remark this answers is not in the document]
jvz: This is a class imported by an auto-config class in one case and one _named_ as an auto-config class in the other case. However, this has made me realize that there's a bit of a bean mess in the various Spinnaker services because `PluginsAutoConfiguration` imports `PlatformComponents` (which is an auto-config but is being directly imported as a regular config in this context), but `PluginsAutoConfiguration` is not itself an auto-config, and it is directly imported by different configuration classes in different services in such a way that the beans from there aren't directly visible by IntelliJ consistently through different modules. I want to look more closely at this.

And to be specific, this `BootstrapComponents` class is a de-duplication of the two beans defined exactly this way in both `PluginsAutoConfiguration` and `PlatformComponents` I found while doing something else at one point that got slightly mixed in here due to affected code (declarations of a default `DynamicConfigService` bean).

This whole area may need some deeper digging. My main purpose is to provide the beans for accessing `OpenFeatureAPI` directly, but I added the `DynamicConfigService` integration for a simpler way to get started using `flagd` (or whatever OpenFeature provider you want to use in theory) which can control all the existing features of Spinnaker that use `DynamicConfigService`. However, now this raises further questions about where to place the beans.

jvz:
> ... Does this implementation still use enough spring under the covers to suffer this way?

It allows it as a backup, though perhaps I could make that configurable as well. Right now, it configures a `DynamicConfigService` to fallback to a Spring-based one for undefined feature flags from OpenFeature.

dbyron-sf:
I think I see where you're coming from...or at least the ability to enable features per-user (or some other attribute) makes sense.  I'm not quite sure I see the part about conditional beans.  Can you give an example of how this would play out as a feature moves from alpha to stable?

kork-core/src/main/java/com/netflix/spinnaker/kork/BootstrapComponents.java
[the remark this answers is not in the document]
dbyron-sf: Fair enough.  This is no worse than it was before in PluginsAutoConfiguration and PlatformComponents.

dbyron-sf:
OK, agreed.  Seems like this is a bit of a separate thing from OpenFeature integration (i.e. always enable all the beans and check flags inside them).  As in, I don't think this PR actually moves in this direction.  And yes, as we get more flags, this will be more important.

dbyron-sf:
There is something nice about beans not loading at all as far as reasoning about changes in behavior.  It's worth the tradeoff though.

jvz:
I think the bean loading bit should be limited to optional features rather than unstable features. Otherwise, the combinatorial explosion of possible configurations to consider is too large for us humans to handle (once you have, say, 10 optional features, you now have 1024 combinations of flags possible, and it's likely that ~1000 of them are invalid combinations).

dbyron-sf:
You're saying conditional bean loading is for optional features only, or for unstable features only?  I think you're saying unstable.

jvz:
Conditional bean loading should be for optional features only, right. Unstable features that are meant to be part of the default system shouldn't bother creating conditional beans (but they should annotate their stability guarantees at least).

kork-core/src/main/java/com/netflix/spinnaker/kork/dynamicconfig/OpenFeatureConfigConfiguration.java
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.core.env.Environment;
+import org.springframework.stereotype.Component;
+
+/**
+ * Configuration to enable OpenFeature and set up a {@link DynamicConfigService} bean based on it.
+ */
+@Component
+@EnableOpenFeature
+public class OpenFeatureConfigConfiguration {
```
jasonmcintosh, line 31: DO we need to conditionally enable this configuration if "openfeature.eanbled" true kinda thing?

kork-flags-git/src/main/kotlin/com/netflix/spinnaker/kork/flags/flagd/git/GitRepositoryPoller.kt
```diff
@@ -0,0 +1,81 @@
...
+  private val properties: GitFlagsProperties,
+  private val applicationEventPublisher: ApplicationEventPublisher,
+) : AutoCloseable {
+  private val directory = Files.createTempDirectory("GitRepositoryPoller")
+  private val git = Git.cloneRepository()
+    .setURI(properties.repositoryUri)
+    .setBranch(properties.branch)
+    .setDepth(1)
+    .setDirectory(directory.toFile())
+    .setCredentialsProvider(UsernamePasswordCredentialsProvider("token", properties.apiToken))
```
jasonmcintosh, line 46: Tied to the above jgit concern.. API token vs. Oauth2 creds?

spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -178,6 +183,7 @@ dependencies {
     // some of the modules would still use <1.70 as they can't be upgraded without upgrading spring boot
     api("org.bouncycastle:bcpkix-jdk18on:${versions.bouncycastle}")
     api("org.bouncycastle:bcprov-jdk18on:${versions.bouncycastle}")
+    api("org.eclipse.jgit:org.eclipse.jgit:${versions.jgit}")
```
jasonmcintosh, line 186: https://github.com/spinnaker/clouddriver/blob/02a601956d8cabd83b8db1516ab8a97aaa531a1a/clouddriver-cloudrun/clouddriver-cloudrun.gradle#L27 IF we choose to really support jgit (vs. say... referencing an internal artifact load via clouddriver APis to provide artifacts which has chicken/egg headaches)... probably should consider  cleaning these other areas up.

jvz: Would probably make sense to use artifact support here, though how would you use feature flags in the artifact system?

kork-core/src/main/java/com/netflix/spinnaker/kork/dynamicconfig/OpenFeatureConfigConfiguration.java
jvz: Possibly. I was originally thinking of making this conditional on a `FeatureProvider` bean being available, but auto-configuration doesn't work all that well in this scenario. I'm thinking there may be some properties related to enabling use of OpenFeature for `DynamicConfigService` versus Spring properties versus both versus nothing.

kork-core/src/test/groovy/com/netflix/spinnaker/kork/dynamicconfig/OpenFeatureDynamicConfigServiceSpec.groovy
jvz: Ported over to Java.

kork-core/src/main/java/com/netflix/spinnaker/kork/dynamicconfig/OpenFeatureConfigConfiguration.java
jvz: Now uses `flags.providers.openfeature.enabled` which can disable it (enabled by default).

kork-core/src/main/java/com/netflix/spinnaker/kork/BootstrapComponents.java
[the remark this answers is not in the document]
jvz: Based on how I've redone the bean config, these changes might not be relevant anymore, but I did put them in a separate commit.

kork-core/src/main/java/com/netflix/spinnaker/kork/dynamicconfig/OpenFeatureDynamicConfigService.java
```diff
@@ -0,0 +1,118 @@
...
+      details = features.getBooleanDetails(configName, (Boolean) defaultValue);
+      value = (T) details.getValue();
+    } else if (configType == Integer.class) {
+      details = features.getIntegerDetails(configName, (Integer) defaultValue);
+      value = (T) details.getValue();
+    } else if (configType == Long.class) {
+      // TODO(jvz): https://github.com/open-feature/java-sdk/issues/501
+      var intDetails = features.getIntegerDetails(configName, ((Long) defaultValue).intValue());
+      details = intDetails;
+      value = (T) Long.valueOf(intDetails.getValue().longValue());
```
jvz, line 57: Note: I've found more issues with this part that I'm investigating. I'm considering adding a facade API for feature flags since OpenFeature is missing support for `long` (and any other `Number` type besides `int` and `double`), and I'm not trying to fork the library. While I'd normally criticize people for doing that, this seems to be a scenario where a facade is useful.

kork-core/src/main/java/com/netflix/spinnaker/kork/dynamicconfig/OpenFeatureImportBeanDefinitionRegistrar.java
```diff
@@ -0,0 +1,84 @@
...
+      return;
+    }
+    var features = beanFactory.getBeanProvider(Features.class).getIfUnique();
+    if (features == null) {
+      return;
+    }
+
+    var dynamicConfigServiceFallbackBean = new GenericBeanDefinition();
+    var enableSpringFlags =
+        environment.getProperty("flags.providers.spring.enabled", Boolean.class, true);
```
jvz, line 60: @dbyron-sf here's the feature flag to disable this dynamic property support from Spring when performance issues are a concern. It would be more efficient for dynamic flags to be handled by something like `flagd` which indexes all the feature flags by name in a hash map (something Spring doesn't do for whatever reason).

spinnaker-dependencies/spinnaker-dependencies.gradle
jvz: Seems like upgrading JGit to the latest version is easy to support in Clouddriver. It just requires a couple import statements to be updated as some classes have moved around.

jasonmcintosh: NOMINALLY since using it through-out, add to kork spinnaker-dependencies constraint ;)

kork-flags-git/src/main/kotlin/com/netflix/spinnaker/kork/flags/flagd/git/GitRepositoryPoller.kt
```diff
@@ -0,0 +1,81 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.kork.flags.flagd.git
+
+import com.netflix.spinnaker.kork.flags.flagd.SyncFlagsEvent
+import org.apache.logging.log4j.kotlin.Logging
+import org.eclipse.jgit.api.Git
```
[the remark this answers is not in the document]
jasonmcintosh: Ideally call out in docs "Hey don't use a huge monorepo for feature flags" or such.  Good docs on this be helpful!  Seems ok :)

kork-core/src/main/java/com/netflix/spinnaker/kork/dynamicconfig/OpenFeatureConfigConfiguration.java
jasonmcintosh: Awesome thanks!

jvz:
Since attempting to use OpenFeature, I was never able to get it to work nicely with Spinnaker and have had to rely on regular Spring for "feature flags". Closing for now.

PR #1220, opened 2025-01-08
feat(retrofit2): Improvise OkHttpClientProvider to accept interceptors

kirangodishala:
As an alternate to using `OkHttp3ClientConfiguration` for creating an OkHttpClient, `OkHttpClientProvider` was created through
https://github.com/spinnaker/kork/pull/639. But there are many existing retrofit clients that are passing Interceptor(s) which is not supported by `OkHttpClientProvider`.  This PR allows `OkHttpClientProvider` to accept `okhttp3.Interceptor` list. 

As we are slowly moving to retrofit2(rosco completed, [echo](https://github.com/spinnaker/echo/pull/1466) under review), the real consumer of OkHttpClientProvider will be `Retrofit2ServiceFactory`(as opposed to `RetrofitServiceFactory`) but since it's an implementation of `ServiceClientFactory`,  I had to add a method that takes `List<Interceptor>` which is specific to `Retrofit2ServiceFactory` and hence the method has no significance wrt `RetrofitServiceFactory`.

The idea is to use `OkHttpClientProvider` instead of `OkHttp3ClientConfiguration` while upgrading the http clients from retrofit1 to retrofit2.

### DIFF
--- a/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/RetrofitServiceFactory.java
+++ b/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/RetrofitServiceFactory.java
@@ -27,6 +27,8 @@ import com.netflix.spinnaker.kork.annotations.NonnullByDefault;
 import com.netflix.spinnaker.kork.client.ServiceClientFactory;
 import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerRetrofitErrorHandler;
 import com.netflix.spinnaker.retrofit.Slf4jRetrofitLogger;
+import java.util.List;
+import okhttp3.Interceptor;
 import retrofit.Endpoint;
 import retrofit.RequestInterceptor;
 import retrofit.RestAdapter;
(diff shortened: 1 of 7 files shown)


kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/RetrofitServiceFactory.java
```diff
@@ -62,4 +64,15 @@ public <T> T create(Class<T> type, ServiceEndpoint serviceEndpoint, ObjectMapper
...
   }
+
+  @Override
+  public <T> T create(
+      Class<T> type,
+      ServiceEndpoint serviceEndpoint,
+      ObjectMapper objectMapper,
+      List<Interceptor> interceptors) {
+    // retrofit1 doesn't support okhttp3.Interceptor. This method override is just to fulfill the
+    // contract.
```
dbyron-sf, line 75: Safer to throw an exception here if interceptors isn't empty?

kork-web/src/main/java/com/netflix/spinnaker/kork/client/ServiceClientFactory.java
```diff
@@ -29,11 +31,27 @@ public interface ServiceClientFactory {
...
+   * @param objectMapper mapper
+   * @param interceptors list of interceptors
+   * @param <T> type of client , usually an interface with all the remote method definitions.
+   * @return an implementation of the type of client given.
+   */
+  public <T> T create(
+      Class<T> type,
+      ServiceEndpoint serviceEndpoint,
+      ObjectMapper objectMapper,
+      List<Interceptor> interceptors); // This is supported only by Retrofit2ServiceFactory
```
dbyron-sf, line 53: I suspect this comment will get stale over time.

dbyron-sf:
We could use some test coverage for the new methods.

dbyron-sf (approved):

PR #1222, opened 2025-01-27
feat(retrofit2): add a method that returns Response object when retrofit2 client invokes an api

kirangodishala:
Existing method `Retrofit2SyncCall.execute()` returns response body when an api is invoked by a retrofit2 client and hence there is no way to capture headers when needed. So added a new method that returns `retrofit2.Response` object through which headers can be accessed.

### DIFF
--- a/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/Retrofit2SyncCall.java
+++ b/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/Retrofit2SyncCall.java
@@ -19,6 +19,7 @@ package com.netflix.spinnaker.kork.retrofit;
 import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerNetworkException;
 import java.io.IOException;
 import retrofit2.Call;
+import retrofit2.Response;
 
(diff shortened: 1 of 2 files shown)


kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/Retrofit2SyncCall.java
```diff
@@ -36,4 +37,20 @@ public static <T> T execute(Call<T> call) {
...
+
+  /**
+   * Handle IOExceptions from {@link Call}.execute method centrally, instead of all places that make
+   * retrofit2 API calls.
+   *
+   * @throws SpinnakerNetworkException if IOException occurs.
+   * @param call call to be executed
+   * @return Response<T> response after execution
+   */
+  public static <T> Response<T> executeCall(Call<T> call) {
```
dbyron-sf, line 49: What do you think of calling this from `execute` above to reduce some duplication?

kirangodishala: Done.

dbyron-sf (approved):

PR #1226, opened 2025-02-12
refactor(retrofit2): add more details to the SpinnakerConversionException's message

kirangodishala:
At present the SpinnakerConversionException.getMessage() gives only the static text : `Failed to process response body`. With this PR, the message will have more details on the actual conversion error.

Ex: `Failed to process response body: Unrecognized token 'this': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')`

### DIFF
--- a/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
+++ b/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
@@ -153,7 +153,7 @@ public class ErrorHandlingExecutorCallAdapterFactory extends CallAdapter.Factory
         }
       } catch (JsonProcessingException jpe) {
         throw new SpinnakerConversionException(
-            "Failed to process response body", jpe, delegate.request());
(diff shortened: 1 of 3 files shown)


kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/ErrorHandlingExecutorCallAdapterFactory.java
```diff
@@ -153,7 +153,9 @@ public Response<T> execute() {
         }
       } catch (JsonProcessingException jpe) {
         throw new SpinnakerConversionException(
-            "Failed to process response body", jpe, delegate.request());
+            "Failed to process response body: " + jpe.getOriginalMessage(),
```
dbyron-sf, line 156: What do you think of:
```suggestion
            "Failed to process response body: " + jpe.getMessage(),
```
With getOriginalMessage, the resulting message is e.g.:
```
Failed to process response body: Unexpected end-of-input: was expecting closing quote for a string value
```
With getMessage, the resulting message:
```
Failed to process response body: Unexpected end-of-input: was expecting closing quote for a string value
 at [Source: (okhttp3.ResponseBody$BomAwareReader); line: 1, column: 29]
```
I think the extra detail is helpful.

dbyron-sf (approved):

PR #1227, opened 2025-02-14
Add Observability Plugin as part of kork-actuator

danielGz:
This work is about donating the armory observability plugin: https://blog.spinnaker.io/announcing-the-new-spinnaker-observability-plugin-d7fbb17e1e07 as part of kork-actuator

This is a sample configuration
```
observability:
  enabled: true
  config:
    metrics:
      prometheus:
        enabled: true
management:
  metrics:
    export:
      datadog:
        enabled: false
  endpoints:
    web:
      exposure:
        include: health,info,aop-prometheus
```


kork-actuator/src/main/java/com/netflix/spinnaker/kork/actuator/observability/config/ObservabilityConfiguration.java
```diff
@@ -0,0 +1,135 @@
+/*
+ * Copyright 2025 Netflix, Inc.
```
christosarvanitis, line 2: Harness :)

spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -189,9 +189,9 @@ dependencies {
     // java.lang.NoClassDefFoundError at LogbackLoggingSystem.java:293
     // Hence pinning slf4j-api at 1.7.36 which spring boot 2.7.18 brings in.
     api("org.slf4j:slf4j-api"){
-      version {
-        strictly("1.7.36")
-      }
+//      version {
```
christosarvanitis, line 192: Any particular reason for this change? I think it is wise to pin this version as unpinning creates problems in the services.

kork-actuator/src/main/java/com/netflix/spinnaker/kork/actuator/observability/model/MetricsConfig.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @Bean
+  @ConditionalOnProperty(value = "observability.config.metrics.newrelic", matchIfMissing = true)
+  public MetricsNewRelicConfig newrelic() {
+    return this.newrelic;
+  }
+
+  @ConditionalOnProperty(value = "observability.config.metrics.datadog", matchIfMissing = true)
+  @Bean
+  MetricsDatadogConfig datadog() {
```
christosarvanitis, line 46: Datadog config is declared here but we also need the management.datadog config based on the description? Maybe we would consider overridiing the default micrometer configuration with the our implementation as we did for the Obervability plugin.

christosarvanitis (commented):
Thanks for the PR @danielGz. Some concerns around the Datadog implementation/configuration as i know that one is tricky to override

dbyron-sf:
Thanks so much for getting started on this work.  I'd like to wait until we release 1.37.0 though so we have plenty of time for all the git juggling, and general time having it soak before imposing it on everyone.

spinnaker-dependencies/spinnaker-dependencies.gradle
danielGz: that was not my change, it actually appears to be made by you https://github.com/armory-io/kork/commit/5f87a1f841197440ca8793d6dfac8993e2eaa190#diff-4f290478509c316a6dadfbd91cbf74598bd0969cf02a9bd7eb796624fe043ec6L193

christosarvanitis: I see! Yeah i did that because to run a composite build with kork. Its not needed as part of this PR

christosarvanitis:
@dbyron-sf  do you think we can restart this work?

dbyron-sf:
> @dbyron-sf do you think we can restart this work?

Sure, if it happens in https://github.com/spinnaker/spinnaker.  There are still a few other disruptive changes coming soon (the rest of the retrofit2 changes, new gate oauth, spring boot 3.  Let's see which ones are ready to go first.

